PR #4, opened 2016-04-03
Setup script clones at "root path"

dixpac:
* Scripts creates dirs one level up from dradis-ce
  so there is no need for user to manually copy them from
  community dir
* Also small typo change in model code documentation

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -91,7 +91,7 @@ class Node < ActiveRecord::Base
     end
   end
 
-  # Security scanner output files uploaded via the Upload Manager use thsi node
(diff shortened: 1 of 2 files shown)


etdsoft:
Hi @dixpac thanks for the PR. The intent is good, but the refactored code seems a lot less clear to read than the original and it will also make it tricky to spot bugs / omissions in the future.

Feel free to re-submit leaving the full `git clone` lines as they were (but dealing with the incorrect path).

dixpac:
@etdsoft sure, I will play with this in following days to make it more clear! 
thanks for the review

PR #11, opened 2016-05-16
add new create evidence form to issues#show

gregmolnar:
The textarea list creates a node for each line(as per the description of the tak), but I propose to allow to add existing nodes to that list too, so we would use `find_or_create`, because let's say I ran an nmap script to find ms10-054 vulnerable hosts on a bunch of IPs and 15 of them is vulnerable. I already have the grep-able output of nmap, so it is easy to create a host list and just paste it into a textarea regardless of which one of them is already in the system.


georgemillo:
Closed in favour of #19, as I can't push new commits to your fork ;)

PR #14, opened 2016-05-23
Issue columns

etdsoft:
* Detect the list of fields that the different Issues have and render all of them as columns in the Issues#index table.
* Allow the user to show/hide the different columns.
* Store the selection of columns in a per-user, per-project manner.

### DIFF
--- a/app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
+++ b/app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
@@ -22,7 +22,7 @@ class IssueTable
     @$column_menu.find('a').on 'click', @onColumnPickerClick
 
   loadColumnState: =>
-    if typeof(Storage) != "undefined"
+    if Storage?
       @selectedColumns = JSON.parse(localStorage.getItem(@storageKey()))
     else
       console.log "The browser doesn't support local storage of settings."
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
```diff
@@ -21,8 +22,13 @@ class IssueTable
     @$column_menu.find('a').on 'click', @onColumnPickerClick
 
   loadColumnState: =>
-    # TODO: persist this in browser local storage or a cookie
-    @selectedColumns = ['title', 'tags', 'affected']
+    if typeof(Storage) != "undefined"
```
gregmolnar: You could just use coffeescript's existential operator here:
```
if Storage?
```

app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
```diff
@@ -109,6 +116,13 @@ class IssueTable
           $($row.find('td')[2]).replaceWith("<td class='text-error'>Please try again</td>")
       }
 
+  saveColumnState: ->
+    if typeof(Storage) != "undefined"
```
gregmolnar: Same as above `Storage?` would be better.

gregmolnar:
What about using cookie storage as a backup if there local storage is not supported by the browser?

gregmolnar:
I just checked the browser support of localstorage and I think it wouldn't worth to implement a cookie based version as it is pretty widely supported: http://caniuse.com/#search=localstorage.
So ignore my previous comment.

PR #15, opened 2016-05-27
Fix #144

kulisu:
### copy to clipboard

> attachment url likes `!/pro/nodes/123/attachments/High-SQLi-Search-2.png!` could be copied to clipboard by clicking it's size info.

### preview uploaded image

> using [bootstrap tooltip](http://v4-alpha.getbootstrap.com/components/tooltips/) to preview image attachments while hovering name info.

### quick pasting image

> images would upload automatically from clipboard pasting while its checkbox is checked.

### auto increasing identity

> increase attachment identity likes `High-SQLi-Search-3.png`, `High-SQLi-Search-4.png` automatically while its checkbox is checked.

## REF:

[Better picture naming from Ctrl+C](https://github.com/securityroots/dradispro-tracker/issues/144)

> we're suggesting that you eliminate the .png/.jpg/.etc picture extensions,
> which is added by the code automatically and let the user focus on the picture name (rather than extension).

implemented, but seems wont work while pasting from clipboard.

> It would also be great if the name of the last saved image name would be remembered and the number after it would be automatically increased by 1:
> instead of always presenting a default name, which needs to be deleted and renamed every single time.

implemented, and it could be switch of/off by checkbox.

> We suggest you save the name of the picture in a cookie,
> which would make it browser based and each user/browser can have it's own remembered name.

implemented, but saves it in HTML5 [sessionstorage](http://www.w3schools.com/html/html5_webstorage.asp).

> this requires some work, but not much,
> but if you don't have time to actually implement this,
> you should at least remember the last picture name and present the test1.png to the user (rather than screenshot-XX.png).

implemented, as above.

### DIFF
--- a/Gemfile.plugins
+++ b/Gemfile.plugins
@@ -1,0 +1,2 @@
+gem 'clipboard-rails'
+
--- a/app/assets/javascripts/snowcrash.js
+++ b/app/assets/javascripts/snowcrash.js
@@ -11,6 +11,8 @@
 //= require d3
 //= require local_time
 
+//= require clipboard
+
 //= require snowcrash/plugins/jquery.breadcrums
 //= require snowcrash/plugins/jquery.treemodal
 //= require snowcrash/plugins/jquery.treenav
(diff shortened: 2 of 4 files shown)


etdsoft:
Closing this for now, copy to clipboard was implemented in https://github.com/dradis/dradis-ce/pull/17

The other ideas are good, but there is too much going in a single PR. Maybe we can work on individual PRs? Thanks for sharing though.

PR #17, opened 2016-06-02
Attachment URLs can be copied to clipboard

georgemillo:


etdsoft:
@georgemillo  the only comment I have here is: should we vendorize clipboard.js or rely on a gem like [clipboard-rails](https://github.com/sadiqmmm/clipboard-rails). We're doing both at the moment (jQuery UI is a gem dependency `d3.js` is vendorized.

I think that for something simple like this we may be better of vendorizing, but have you given it any thought?

georgemillo:
> should we vendorize clipboard.js or rely on a gem like clipboard-rails

I don't really see any strong argument either way. I'd vote to keep it vendorized just because that's how it currently is and it's working fine.

etdsoft:
@georgemillo I moved the JS trigger code to the right section of the file and included a `js-` prefix in the CSS class to signify we're only using this for a JS trigger.

I've noticed this though, some times the click works (text goes to clipboard) but the tooltip doesn't show:

attachments

Can you reproduce?

georgemillo:
Fixed; see the commit message for the latest commit.

Btw, what did you use to create that screencast GIF?

etdsoft:
Something seems to still be broken...

tooltip_madness

thoughts?

GIFs - http://www.cockos.com/licecap/

PR #19, opened 2016-06-04
As a user I can associate an Issue with multiple nodes at a time.

georgemillo:


georgemillo:
I've added the ability to specify a parent node for the newly-created nodes. But I think the current UI is very confusing.

When I open the 'add evidence' form, it's not at all clear what the difference is between checking nodes on the left or typing in their names into the textarea on the right. And regarding that textarea, it's also not clear that if you type in the name of a node that doesn't exist, it will be created, but if you type in the name of an existing node, it'll use that existing one. (I had to figure this out the hard way by playing around with the textarea and looking to see what nodes were or weren't created in the rails console.)

I don't think the problem is the with form inputs *per se* (I think the checkboxes/textarea work fine once you know how they're supposed to be used), just with the labelling - there need to be clearer explanations of what the different parts of the form actually do.

As another minor thing, the "instance" tabs are numbered from 0, not from 1:

screen shot 2016-06-03 at 15 19 47

I realise this is the computer-sciencey way of counting things, but will Dradis's users be thinking like this?

etdsoft:
> there need to be clearer explanations of what the different parts of the form actually do.

This is been in limbo for a few days, do you have any suggestions in terms of making the interface better / clearer?

app/assets/javascripts/snowcrash/modules/issues/evidence.js.coffee
```diff
@@ -4,4 +4,9 @@ jQuery ->
     $('#evidence-tabs .tab-pane').first().addClass('active')
 
     $('#evidence-tabs a[data-toggle="tab"]').on 'shown', (ev)->
-      $('#evidence-tabs #evidence-host-list').css('height', $('#evidence-tabs .tab-content').css('height'))
+      tabContentHeight = $('#evidence-tabs .tab-content').css('height')
+      $tabs            = $('#evidence-tabs #evidence-host-list')
+      if $tabs.height() < tabContentHeight
```
etdsoft: this isn't working as tabContentHeight comes with `px` (e.g. `80px`) and `.height()` is just a number.

But even fixing that it doesn't work because L12 doesn't "reset" the value, it seems to act just as a no-op leaving it with whatever CSS height it had before the call.

etdsoft:
@georgemillo can you take a look at the failing spec? It's driving me nuts!

PR #22, opened 2016-06-13
Optimistic locking

georgemillo:


app/controllers/authenticated_controller.rb
```diff
@@ -19,12 +20,6 @@ class AuthenticatedController < ApplicationController
     raise ActiveRecord::RecordNotFound.new
   end
```
etdsoft: Is this the right call? Those familiar with PaperTrail would expect to see this in a controller or controller concern and not pushed down to a Helper? Do we really need access to this method in the views?

app/controllers/notes_controller.rb
```diff
@@ -28,15 +28,20 @@ def create
   # Retrieve a Note given its :id
   def show
     @activities = @note.activities.latest
+    load_conflicting_versions(@note)
   end
 
   def edit
+    # TODO what if there are no versions? E.g. legacy data
```
etdsoft: Wouldn't PaperTrail make sure `#versions` always returns a `[]` even for legacy instances?

app/controllers/project_scoped_controller.rb
```diff
@@ -24,4 +24,28 @@ def initialize_nodes_sidebar
...
+      #
+      # Only primitive types (String, Array, Hash) can be saved in the flash;
+      # we can't use it to pass a Time objec - so pass the time as a string.
+      flash[:update_conflicts_since] = Time.at(params[name][:original_updated_at].to_i + 1).utc.to_s(:db)
+    end
+  end
+
+  def load_conflicting_versions(record)
+    if flash[:update_conflicts_since]
+      @conflicting_versions = record.versions\
```
etdsoft: Why the trailing `\` this is something we haven't done elsewhere in the code?

app/helpers/snowcrash_helper.rb
```diff
@@ -13,18 +13,18 @@ def css_class_for_sub_nodes(node)
...
-      # In general controllers use :error, but :alert is used with redirect_to
-      #   http://guides.rubyonrails.org/action_controller_overview.html#the-flash
-      flash_css << {
-        'alert'   => ' alert-error',
-        'error'   => ' alert-error',
-        'info'    => ' alert-info',
-        'notice'  => ' alert-success',
-        'warning' => ''
-      }[name]
+    flash.select { |key, _| alert_types.keys.include?(key) }.collect do |name, msg|
```
etdsoft: Whilst I always appreciate a good one-liner, I'm curious if this approach wins us anything over the original?

app/controllers/authenticated_controller.rb
georgemillo: The reason I moved this into a helper was so that I can call it from within `link_to_conflicting_revision`[https://github.com/dradis/dradis-ce/pull/22/files#diff-e86fe144a2807be4de67fe4095611223R37], which needs to be available in the views. 

I can't see a way of making that method aware of whether or not a revision is from the current user or not without duplicating the logic in `user_for_paper_trail`... and if we're going to duplicate the logic, we might as well just call the method.

But I suppose I don't have to *move* the method like this; I could keep in the controller but make it available to the views using `helper_method`, [like this](http://stackoverflow.com/a/8908087/1603071).

What do you think?

app/controllers/notes_controller.rb
georgemillo: You're right; I don't think I meant to commit this TODO. Will remove.

app/controllers/project_scoped_controller.rb
georgemillo: It was to split everything into multiple lines (to keep the line length down), but it's not important.

app/helpers/snowcrash_helper.rb
georgemillo: The point is to allow keys in the flash that aren't about Bootstrap alerts. Elsewhere in this PR I'm using `:update_conflicts_since` as a flash key and in the existing `SnowcrashHelper` using anything except 'alert', 'error' etc as a flash key will raise an error.

etdsoft: Could we amend L27 with a .fetch(name, '') instead?

app/controllers/authenticated_controller.rb
etdsoft: I think making it a `helper_method` from the controller is  more conventional with PaperTrail usage.

PR #12, opened 2016-05-16
Smart(er) Issues#index table

etdsoft:
1. Move away from the HelpScout "magic" toolbar that appears in the center of the screen on selecting a table row to a Gmail toolbar that contains some static button groups, and some more are shown when there are rows selected.

2. Allow for showing / hiding of columns (although the selection is not persisted for now.

3. Filter the table contents from input in the upper-right

PR #25, opened 2016-06-17
Project search

etdsoft:
The original version of this code was contributed by @dixpac


georgemillo:
One bit of UI feedback: as it is, when you view your search results it doesn't tell you *what* you search for. You can see the search query in the URL, and if you click the magnifying glass again, but it's not obvious or prominent.

Putting the search query in big letters at the top would be better imo, that's how Twitter does it:

screen shot 2016-06-20 at 19 36 34

app/assets/javascripts/snowcrash/behaviors.js.coffee
```diff
@@ -105,3 +105,22 @@ jQuery ->
...
+  $('.search-query').on 'blur', ->
+    # Without this, the form will be hidden before the user has a chance to
+    # click on the submit button with the mouse.
+    setTimeout ->
+      $('.form-search').toggleClass('hide')
+      $('#js-search-form-toggle').show()
+    , 10
+
+  $('.btn-search').on 'click', ->
+    console.log('click')
```
georgemillo: A stray `log` statement here.

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,18 @@
+class SearchController < ProjectScopedController
+	include SearchHelper
```
georgemillo: bad indentation

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    @search = Search.new(query: params[:q], scope: @scope, page: params[:page])
+  end
+
+  private
+  def set_scope
+    @scope = if params[:scope].blank? ||
+                  !%{all evidence issues nodes notes}.include?(params[:scope])
+               :all
+             else
+               params[:scope].to_sym
```
georgemillo: Creating symbols from user input - is this still a no-no in the version of Ruby we're using?

app/helpers/search_helper.rb
```diff
@@ -0,0 +1,40 @@
+module SearchHelper
+	def search_filter_path(options={})
```
georgemillo: more bad indentation; looks like this is coming from mixing tabs and spaces

app/helpers/search_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+  	exist_opts = {
+    	q: params[:q],
+      scope: params[:scope]
+    }
+
+  	options = exist_opts.merge(options)
+    search_path(options)
+	end
```
georgemillo: my eyes!

spec/features/user_searches_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+    end
+  end
+end
+
+def setup_test_data
+  create(:issue, text: "Issue search")
+  node = create(:node, label: "Node search")
+  create(:note, text: "Note search", node: node)
+  create(:evidence, content: "Evidence search")
+end
```
georgemillo: This is defined in global scope and will potentially pollute/conflict with other specs. Better to define it within the file's outermost `describe` block.

spec/helpers/search_helper_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+end
+
+def options(term:, scope:)
+  Hash.new.tap do
+    params[:q] = term
+    params[:scope] = scope
+  end
+end
```
georgemillo: same concern as above regarding global scope.

spec/models/search_spec.rb
```diff
@@ -0,0 +1,272 @@
...
+      expect(search.total_count).to eq 4
+      expect(search.notes_count).to eq 1
+      expect(search.issues_count).to eq 1
+      expect(search.nodes_count).to eq 1
+      expect(search.evidence_count).to eq 1
+    end
+  end
+
+  describe "#total_count" do
+    it "returns 0 if noting founded" do
```
georgemillo: typo: "noting" for "nothing"; the same mistake is repeated a few other times in this file.

spec/support/sign_in_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def fill_login_form(password:)
+    fill_in "login", with: "test@example.com"
+    fill_in "password", with: password
+    click_on "Let me in!"
+  end
+end
+
+RSpec.configure do |c|
+  c.include SignInHelper, type: :feature
```
georgemillo: I don't like the idea of splitting RSpec config into multiple files throughout `spec/support`, as opposed to just keeping it all in `spec_helper`.

Most other "spec modules" are being `include`d into RSpec in the same place, i.e. [in spec_helper]( https://github.com/dradis/dradis-ce/blob/master/spec/spec_helper.rb#L32). I think it would be better to include SignInHelper there, for consistency, rather than here.

georgemillo:
Regarding the tests that are failing on Codeship, this is c&ped from Slack:

georgemillo [7:28 PM]  
@etd those tests are passing for me locally

[7:28]  
but I imagine the failure is because...

[7:28]  
 ```    it "returns list of matches order by updated_at desc" do
      first = create(:issue, text: "First issue")
      second = create(:issue, text: "Second issue")

      results = Search.new(query: 'issue', scope: :issues).results

      expect(results.map(&:text)).to eq [second.text, first.text]
    end
```

[7:29]  
`first` and `second` are getting created less than one, er, second apart

[7:30]  
if Codeship’s system only records times to the nearest second precision rather than to the nearest millisecond, it will think they were both created at the exact same time

[7:30]  
so we can’t guarantee that they’ll be returned in the order the test thinks they will

[7:30]  
setting the `updated_at` timestamps manually might fix it?

```      first = create(:issue, text: "First issue", updated_at: 10.seconds.ago)
      second = create(:issue, text: "Second issue", updated_at: 5.seconds.ago)

georgemillo:
I've left some comments about the code, but when it comes to the UI/UX and the functionality, I have no comments except the thing about displaying the search query more prominently on the results page. It looks good and seems to work.

app/controllers/search_controller.rb
etdsoft: It isn't, but this should not be an issue anyway because in line #12 we're making sure that the value is in the whitelist.

spec/models/search_spec.rb
etdsoft: Good catch! Surprised you didn't pick up on "founded"

spec/support/sign_in_helper.rb
etdsoft: Agreed. And besides [ControllerMacros](https://github.com/dradis/dradis-ce/blob/dix-global-search/spec/support/controller_macros.rb) already provides this functionality.

spec/models/search_spec.rb
georgemillo: Haha yes, whoops.

PR #29, opened 2016-06-28
Search - case insensitive SQL

etdsoft:
This is a case-insensitive SQL query implementation that has the benefit
of working both in SQLite3 and MySQL.

    LOWER(label) LIKE LOWER(:q)

The MySQL-only solution could have been:

    label LIKE :q COLLATE UTF8_GENERAL_CI

However, a benchmark yielded very similar results:

```
$ bundle exec rails runner sql-lower-vs-collate.rb
       user     system      total        real
collate:  0.870000   0.010000   0.880000 (  0.875807)
  lower:  0.840000   0.010000   0.850000 (  0.853030)
```

From:

```
queries = [
  "label LIKE :q COLLATE UTF8_GENERAL_CI",
  "LOWER(label) LIKE LOWER(:q)"
]

n = 10_000

Benchmark.bm do |x|
  x.report('collate:') do
    n.times do
      Node.user_nodes.where(queries[0], q: "%host%").order(updated_at: :desc)
    end
  end

  x.report('  lower:') do
    n.times do
      Node.user_nodes.where(queries[1], q: "%host%").order(updated_at: :desc)
    end
  end
end
```

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -52,7 +52,7 @@ class Search
 
   def issues
     Issue.where(
-      "node_id = :node AND text LIKE :q",
+      "node_id = :node AND LOWER(text) LIKE LOWER(:q)",
       node: Node.issue_library,
       q: "%#{query}%"
     ).order(updated_at: :desc)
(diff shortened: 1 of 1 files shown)


etdsoft:
@georgemillo thoughts?

PR #30, opened 2016-06-29
Don't show load button in nodes tree when node has no children

georgemillo:


app/assets/javascripts/snowcrash/plugins/jquery.treenav.js.coffee
```diff
@@ -29,9 +29,6 @@ do ($ = jQuery, window, document) ->
       if !target.siblings('.children').first().has('li.node').length
         $menu = target.siblings('.children')
         $.get(target.attr 'href')
-        .done ->
-          $menu.find('li.loading').hide()
-          $menu.find('li.error').hide()
```
georgemillo: If you're wondering why I removed this, it's because when they're present, the 'Loading' text gets hidden before the tree has even started expanding, so you never see it (for nodes below the top level at least):

hideloading

With the lines removed you can see the "loading" text and the slide looks less jarring:

withouthideloading

PR #37, opened 2016-07-30
Combine issues

dormi:


georgemillo:
this is probably scope creep, but we need to think about what happens with the Trash when you combine issues. Here's what happens right now (and this is just one example, there are probably other cases to think about):

1. create two issues
2. combine them
3. view trash, and the two issues you just combined are in there (because as far as PaperTrail is concerned, those issues just got deleted the regular way.)

I haven't tested it, but I imagine another potential problem could be if you deleted an evidence, merged that evidence's issue, then tried to restore the evidence: the evidence's `issue_id` would point to an evidence that no longer exists.

georgemillo:
the 'toggle all issues' button doesn't always show/hide the 'combine' button when it's supposed to:

toggle-all-combine

(Although as you can see from the gif, I had to click around a bit before I got the error, and I'm not sure exactly which steps cause it.)

georgemillo:
the evidence-merging functionality seems to work fine, but it's not clear at all from the "Combine Issues" modal what the different options mean:

screen shot 2017-01-13 at 13 05 55

Questions the user might have: what does it mean to 'merge' one issue into another? What happens to the evidence in each issue and to the text?

One example: I can see how the user might not realise that when you combine all the issues into a new one, *the text of the original issues is lost* unless you copy it into the new textarea. And even then, there's no way to easily get the text of the old issues without opening those issues in a new tab.

georgemillo:
Also, when you first open the 'combine issues' modal, no radio button is selected by default. Looks like that if you submit the form with no radio selected, it defaults to the 'combine into new issue' option, but that should be clearer.

app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
```diff
@@ -119,6 +125,38 @@ class IssueTable
...
+      issues_to_combine.push({id: id, name: name})
+
+    for issue in issues_to_combine
+      html =  "<div class='radio'><label><input type='radio' name='id' value='#{issue.id}' />#{issue.name}</label></div>"
+      html += "<input type='hidden' name='sources[]' value='#{issue.id}' />"
+      $('#modal_combine_issues #issues_to_combine').append(html)
+
+    $("#modal_combine_issues").modal()
+
+    # TODO: $($row.find('td')[2]).replaceWith("<td class=\"loading\">Combining...</td>")
```
georgemillo: did you mean to leave this as a TODO?

spec/features/issues_spec.rb
```diff
@@ -49,6 +49,88 @@
...
+
+            # check that the form inside the modal has the required fields
+            expect(page).to have_selector("input[name='id']", count: 2)
+            expect(page).to have_selector("input[name='new_issue']", 1)
+            expect(page).to have_selector("input[name='sources[]']", count: 2)
+          end
+
+          it "combines issues into an existing one" do
+            # click > 1 issue checkboxes
+            page.all('input.js-multicheck').each do |checkbox|
```
georgemillo: could be compressed into one line: `page.all('input.js-multicheck').each(&:click)`

spec/features/issues_spec.rb
```diff
@@ -49,6 +49,88 @@
...
+            first("input[name='id']").click
+
+            click_button "Combine"
+
+            expect(page).to have_content("1 issues combined into ")
+          end
+
+          it "combines issues into a new one" do
+            # click > 1 issue checkboxes
+            page.all('input.js-multicheck').each do |checkbox|
```
georgemillo: same here

app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
dormi: removing, because we are not combining issues by ajax, we are using a full http request with page reloading

spec/features/issues_spec.rb
dormi: 👍

spec/features/issues_spec.rb
dormi: 👍

dormi:
Yet another problem: when combining into a new one, author is nor being saved ?

etdsoft:
> NOTE: If we want to add add the whole issue text into the modal, it may be better to switch to a dedicated "combining" view insetad of using a modal ?

I think you have a point, we need to promote `combine` to a 1st class view.

That way we can also show a preview of what will happen (e.g. a "Result" section where the content of the issue changes depending on the radio option that is selected. Once we have a basic design we can even think about ways of making the original text for the different issues available so users can mix & match and choose the best paragraphs of each of them.

We will also have more room to explain the operation and provide micro-copy to clarify.

etdsoft:
> this is probably scope creep, but we need to think about what happens with the Trash when you combine issues.

When merging I'm ok with having the original issues be deleted from PaperTrail / Trash perspective. That way if you realised you made a mistake in the merge, at least you can recover the originals.

It is worth writing a bunch of test cases to ensure this works (e.g. the delete evidence scenarios @georgemillo described).

At the same time, I'm not sure what to do with the Activity Feed on the combine operation. Probably we can leave it as it is until the functionality is implemented, and think about it later.

> This seems like a quick fix in L91:

    @object.touch if @object.persisted?

The `#touch` is failing not because the object isn't persisted but because the object failed to save in the previous line due to some validation error.

etdsoft:
Something else that has been brought up that it think makes sense: we need labels in the toolbar buttons. It's probably worth adding a text description to the right of each icon in every button except the "select all" checkbox.

dormi:
> I think you have a point, we need to promote `combine` to a 1st class view.

This is what we have now, comments welcome:
captura de pantalla 2017-01-24 a las 7 26 17

---

> Something else that has been brought up that it think makes sense: we need labels in the toolbar buttons. It's probably worth adding a text description to the right of each icon in every button except the "select all" checkbox.
captura de pantalla 2017-01-24 a las 7 25 49

---

> When merging I'm ok with having the original issues be deleted from PaperTrail / Trash perspective. That way if you realised you made a mistake in the merge, at least you can recover the originals.
It is worth writing a bunch of test cases to ensure this works (e.g. the delete evidence scenarios @georgemillo described).
Instead of using `delete_all` we are now using `destroy_all`, callbacks are triggered and PaperTrail works. Looks like that case where we are recovering an evidence whose issue had been previously deleted was already there: https://github.com/dradis/dradis-ce/blob/master/app/models/recoverable_revision.rb#L75

etdsoft:
@georgemillo can you take a look at the latest?

spec/features/issue_pages/table_toolbar_spec.rb
```diff
@@ -29,5 +29,42 @@
         expect(find('.js-issue-actions')).to be_visible
       end
     end
+
+    context "when clicking issues" do
+      before do
```
georgemillo: empty block

app/views/issues/merging.html.erb
```diff
@@ -0,0 +1,56 @@
...
+          <p class="merge_option" style="display: none">Merge issues into a new issue.
+        </div>
+        <div id="change_merge_option" class="span6 pull-right">
+          <a class="merge_option" href="#">Or click here to merge into a new issue</a>
+          <a class="merge_option" style="display: none" href="#">Or click here to merge into a existing issue</a>
+        </div>
+      </div>
+
+      <div class="merge_option">
+        <% checked = true %>
```
georgemillo: This `checked` variable is just to make the first radio checked, right? I don't generally like assigning local variables in views if they can be avoided, I think this can be accomplished more succinctly and with clearer intent using `each_with_index`:

```ruby
issues.each_with_index do |issue, i|
  radio_button_tag 'id', issue.id, i == 0
end
```

app/views/issues/merging.html.erb
```diff
@@ -0,0 +1,56 @@
...
+      <p class="text-warning">WARNING: Only the target issue will prevail, the others will be deleted.</mark>
+      <p>All issues evidences will be moved to the selected issue.
+      <div class="row-fluid">
+        <div class="span6">
+          <p class="merge_option">Select existing issue to merge the others into.
+          <p class="merge_option" style="display: none">Merge issues into a new issue.
+        </div>
+        <div id="change_merge_option" class="span6 pull-right">
+          <a class="merge_option" href="#">Or click here to merge into a new issue</a>
+          <a class="merge_option" style="display: none" href="#">Or click here to merge into a existing issue</a>
```
georgemillo: This text still isn't really clear enough, especially the WARNING in the first sentence (it sounds weird using the word "prevail" to talk about data).

I would rephrase the whole thing and give more detail, and change the order slightly:

> ## Merging X Issues
> 
> Select a target issue to merge the other issues into.
> 
> All issues except the target issue will be deleted, and their evidence will be moved to the target issue.
>
> WARNING: The text of all issues except the target will be lost. If you want to preserve the text, [click here to merge all issues into a new issue with custom text](#).

Then, when they click the link, change to:

> ## Merging X Issues
> 
> All issues will be deleted, and their evidence will be moved into a brand new issue.
>
> WARNING: Only the new text that you enter here will be saved; existing text in the old issues will be deleted.  [Click here to use one issue as a target issue that does not get deleted](#).

This wording might not be perfect, but I think it's worth being as explicit as possible about what happens, to avoid confusion. I also think it helps to describe exactly what happens in terms of how issues get deleted, because then users should be less confused when their "merged" issues appear in the Trash.

app/views/issues/merging.html.erb
dormi: Thank you for this one

dormi:
captura de pantalla 2017-01-30 a las 9 51 28

app/controllers/issues_controller.rb
```diff
@@ -94,6 +94,47 @@ def import
...
+      # create new issue if existing issue not given
+      if @issue.new_record?
+        @issue.author ||= current_user.email
+        if @issue.save && @issue.update_attributes(issue_params)
+          track_created(@issue)
+          tag_issue_from_field_content(@issue)
+        end
+      end
+
+      count = 0
```
aapomm: it seems like you instantiated `count` twice

config/routes.rb
```diff
@@ -21,7 +21,11 @@
   resources :configurations, only: [:index, :update]
 
   resources :issues do
-    collection { post :import }
+    collection do
+      post :import
+      post :merging
```
aapomm: nothing happens yet when at the `:merging` correct? I think it's better to use `GET` at that step.

app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
```diff
@@ -108,17 +114,36 @@ class IssueTable
...
-          if $('input[type=checkbox]:checked').length == 0
-            $('.js-issue-actions').css('display', 'none')
+          if $('input[type=checkbox]:checked.js-multicheck').length == 0
+            that.resetToolbar()
 
         error: (foo,bar,foobar) ->
           $($row.find('td')[that.tagColumnIndex]).replaceWith("<td class='text-error'>Please try again</td>")
       }
 
+  onMergeSelected: (event) ->
```
aapomm: If my understanding is correct, in this part, you're creating a form with the issue ids and then you submit the form to the "merging" step. I think it's unnecessary for the merging action to be `POST` since there is no change in state yet. Instead, you can use `GET` and construct the merging url as such: `issues/merging?issue_id[]=1&issue_id[]=2&issue_id[]=3`

app/controllers/issues_controller.rb
```diff
@@ -94,6 +94,47 @@ def import
...
+        source_ids = params[:sources].map(&:to_i) - [@issue.id]
+        count = @issue.merge source_ids
+      end
+
+    end
+
+    respond_to do |format|
+      format.html {
+        if count > 0
+          redirect_to issues_url, notice: "#{count} issues merged into #{@issue.title}."
```
aapomm: I think it would be better if, after merging, the user gets redirected to the new issue.

app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
dormi: You are right. I spent some time on this.
I decided to use a POST to avoid super long urls when we are merging many issues. I thought about being able to merge a few dozens of issues...
@etdsoft what's your opinion on this?

app/controllers/issues_controller.rb
dormi: thank you!

app/controllers/issues_controller.rb
dormi: Good idea. I change it.

app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
etdsoft: I think that the current /merging /merge URLs are not ideal in a RESTful / CRUD sense, maybe something like

GET /issues/merge/new
POST /issues/merge

Would be more resource-oriented?

Another idea, we don't need a discreet form field for each issue :id to be merged, we can pass them in a list `issues=1,32,124,532,1124` and split it server-side. That'd clean up the URL. In any case, I tend to agree that POST should be reserved to status-changing ops.

etdsoft:
Added some styling changes and updated the copy:

pr37-05

@rachkor ?

PR #38, opened 2016-08-02
bug fix: issue 'select all' btn not working on Firefox

georgemillo:
[Trello](https://trello.com/c/Uv79D16V/199-bug-fix-when-using-firefox-the-select-all-button-does-not-work)

The JS 'click' handler doesn't get called in Firefox, but changing the
wrapping '\<button\>' to a '\<span\>' fixes it.

"\<input\>" tags shouldn't be nested in a "\<button\>" anyway as this as a
violation of the HTML spec. Try it on validator.w3.org and you'll get
an error:

screen shot 2016-08-02 at 14 06 42


etdsoft:
This doesn't seem to be working in FF 48 for Mac?

empty_string

georgemillo:
fixed

PR #39, opened 2016-08-02
Change issue tagging from 1 tag per issue to multiple tags

bosunolanrewaju:
- The dropdown list for preloaded tags was changed to a textarea field to take multiple tags
- Tags display in D3 bar chart changed to list to cater for more than 5 tags in the project
- Icon on Issue#index changed from "bug" to "tag"

screen shot 2016-08-02 at 9 18 22 pm

screen shot 2016-08-02 at 9 21 06 pm

screen shot 2016-08-02 at 9 22 10 pm

screen shot 2016-08-01 at 11 38 23 am

screen shot 2016-08-01 at 11 37 52 am


bosunolanrewaju:
Up for discussion is the color that comes with the preloaded tags. I think we can switch from saving the color HEX with the tag in the database to using CSS with the tag name as the CSS class name. We can separate the design from the data saved like this:

```
.high {
  color: #d62728;
}

.critical {
   color: #9467bd;
}
```

And we can set them on tags in the `class` attribute.

That is if we still want to retain the color implementation on the tags.

PR #40, opened 2016-08-05
Users can add multiple nodes at the same time

georgemillo:


georgemillo:
CI is failing for reasons that don't seem to have anything to do with this PR *per se*. I've fixed it in #41 so merge that first.

app/controllers/nodes_controller.rb
```diff
@@ -18,18 +18,34 @@ def edit
...
-      track_created(@node)
-      flash[:notice] = 'Successfully created node.'
-      redirect_to @node
-    else
-      parent = @node.parent
-      if parent
-        redirect_to parent, alert: @node.errors.full_messages.join('; ')
-      else
-        redirect_to summary_path, alert: @node.errors.full_messages.join('; ')
+    @node.save!
```
etdsoft: Isn't this loosing the error displaying and conditional redirect logic of the original `#create`?

app/controllers/nodes_controller.rb
```diff
@@ -18,18 +18,34 @@ def edit
...
+    track_created(@node)
+    flash[:notice] = 'Successfully created node.'
+    redirect_to @node
+  end
+
+  def multiple
```
etdsoft: Naming convention? Maybe `#create_multiple` is more descriptive? I don't mind which one we agree on, but I'd rather it be consistent across controllers.

See:
https://github.com/dradis/dradis-ce/blob/add-multiple-nodes/app/controllers/evidence_controller.rb#L39

app/controllers/nodes_controller.rb
```diff
@@ -18,18 +18,34 @@ def edit
...
+    if params[:nodes][:parent_id].present?
+      @parent = Node.find(params[:nodes][:parent_id])
+    end
+
+    list = params[:nodes][:list].split(/\n/).map(&:strip).select(&:present?)
```
etdsoft: Maybe `#lines` gives us something here (instead of split):

https://github.com/dradis/dradis-ce/blob/add-multiple-nodes/app/controllers/evidence_controller.rb#L57

I've updated `Evidence#create_multiple` to use `:strip` instead of `:chomp`

app/views/nodes/modals/_add_node.html.erb
```diff
@@ -0,0 +1,97 @@
...
+          :multiple,
+          false,
+          class: "add_node_radio") %>
+        Add multiple
+      </label>
+    </div>
+
+    <hr>
+
+    <%= simple_form_for(
```
etdsoft: Is there a way to avoid having two forms duplicating the type, parent_id, etc.? I guess that one is using `node` and the other `nodes` to scope the parameters. Not sure if a solution can be found with a single `<form>` and the JS toggling the form's `:action` attribute. Did you look into it?

app/controllers/nodes_controller.rb
georgemillo: Yes, but the conditional logic is redundant because the node will never be invalid (i.e. the `else` block never gets called.) There's only one validation on `Node`, a presence validation on `label`. But that will never fail in this case because nodes are given a default label on [line 20](https://github.com/dradis/dradis-ce/pull/40/files#diff-f6042ba19178936a392b7082cf00c9e5L20).

In fact, I'll add a comment to this method to make this clearer.

etdsoft: True today, what if we add a new validation later? Isn't a standard `if .save else` quite a standard practice for controllers?

app/views/nodes/modals/_add_node.html.erb
georgemillo: I had a look at doing it with a single form and JS but this solution seemed much easier. I see the problem with the repetition of the `type` and `parent_id` inputs (could they be extracted to a helper method) but other than that I think having two separate forms can still make sense because "create a node" and "create multiple nodes" feel different enough from each other conceptually to justify having wholly separate forms (after all, they also have separate controller actions.)

PR #45, opened 2016-08-23
Trash feature

etdsoft:


app/controllers/revisions_controller.rb
```diff
@@ -40,6 +40,44 @@ def show
             )
   end
 
+  def trash
+    # Get all versions whose event is destroy.
+    @revisions = PaperTrail::Version.where(event: 'destroy').order(created_at: :desc)
```
etdsoft: Careful, do we need to project-scope PaperTrail::Version in Pro?

app/controllers/revisions_controller.rb
```diff
@@ -40,6 +40,44 @@ def show
...
+  end
+
+  def recover
+    revision = PaperTrail::Version.find params[:id]
+    object = revision.reify
+    # If object's node was destroyed, assign it no a new node.
+    if !Node.exists?(object.node_id)
+      recovered_node = Node.create(label: 'Recovered', type_id: Node::Types::DEFAULT)
```
etdsoft: `find_or_create_by`? to avoid multiple "Recovered" nodes

app/controllers/revisions_controller.rb
```diff
@@ -40,6 +40,44 @@ def show
...
+        issue_object.node_id = Node.issue_library.id
+        issue_object.save
+        object.issue_id = issue_object.id
+        issue_revision.destroy
+      end
+    end
+
+    if object.save
+      # Destroy revision so item is not listed in trash any more.
+      revision.destroy
```
etdsoft: Is this the best option, or should we keep the "destroy" revision and append a new "recovered" revision.

Also, we should probably track this in Activity Feed.

app/helpers/revisions_helper.rb
```diff
@@ -46,4 +46,67 @@ def link_to_conflicting_revision(record, revision)
     link_to text, record_revision_path(record, revision)
   end
 
+  def render_revision_object_icon(revision_object,object_type=nil)
```
etdsoft: Not sure if this may work better in a presenter?

https://github.com/dradis/dradis-ce/tree/master/app/presenters

config/routes.rb
```diff
@@ -55,6 +55,8 @@
 
   get 'search' => 'search#index'
   post 'create_multiple_evidences' => 'evidence#create_multiple'
+  get 'trash' => 'revisions#trash'
```
etdsoft: should we promote Revisions to a resource?

georgemillo:
I don't think the UX is clear enough when you recover something that belongs to "a Node which has since been deleted". When I click "recover" I just see the trash page again with an alert saying "evidence/note/etc recovered" and it's not clear where I can now see the newly-recovered Evidence (or whatever).

Idea: either redirect to the new Evidence after recovering it (instead of showing the trash page again), or update the alert text so it provides a bit of explanation: "As the node which this evidence belongs to has also been deleted, the evidence has been restored a new node called 'Recovered'" (or "to the node called Recovered") if it wasn't newly created.

Or we could do both of those things.

What do you think?

app/controllers/revisions_controller.rb
georgemillo: Could be extracted to a class method called `Node.recovered`, like we do with `.issue_library`, `.plugin_uploads_node` etc?

app/controllers/revisions_controller.rb
georgemillo: I think it's worth adding a new type of activity called "recovered".

app/helpers/revisions_helper.rb
```diff
@@ -46,4 +46,67 @@ def link_to_conflicting_revision(record, revision)
...
+  end
+
+  def revision_object_location(revision_object,object_type)
+    object_type = revision_object_type(revision,revision_object) if object_type.nil?
+    # Get node if object is a Note or an Evidence.
+    if ['Note','Evidence'].include?(object_type)
+      if revision_object.node
+        "at " + link_to(revision_object.node.label,revision_object.node)
+      else
+        'at a Node which has since been deleted'
```
georgemillo: should Node be capitalized here?

In any case, I feel like it's too wordy:

screen shot 2016-08-24 at 18 25 10

And when the note/evidence has a title it looks a bit jarring:

screen shot 2016-08-24 at 18 25 41

How about something like:

for e.g. a note with no title and no node:

`evidence with no title - unknown node`

and with both:

`My Title - note - Node 'my node'`

app/controllers/revisions_controller.rb
etdsoft: Agreed

app/helpers/revisions_helper.rb
etdsoft: I like the last part, but we should try to be consistent re `which has since been deleted`:

https://github.com/dradis/dradis-ce/blob/master/app/views/activities/_node.html.erb#L7

Unknown -> is strange (how come the app doesn't know?), and not 100% accurate.

georgemillo:
Possible scope creep, but do we also want to add an "empty trash" feature? Conceptually, users will expect the trash feature to work like it does everywhere else - as a temporary holding ground before 'permadeleting' a note/evidence/etc.

config/routes.rb
georgemillo: We could do, but it would mean changing the URL `trash` to `revisions/trash` if we wrap this line a `resources` block.

The line directly below this one could be replaced with the functionally-identical:

    resources :revisions, only: [] do
      member { post :recover }
    end

... if you think that's better (it *is* more "Railsy")

app/controllers/revisions_controller.rb
georgemillo: Done:

screen shot 2016-09-02 at 15 18 48

But the revision history link doesn't work - I'll post another comment about it

georgemillo: I just found [this article](https://leonid.shevtsov.me/post/how-to-use-papertrail-for-soft-deletetion/) which has a (very complicated) ActiveRecord query that we can supposedly use to get all deleted, non-recovered revisions to list here *without* having to destroy the "delete" revisions.

What do you think about this (assuming that the code there works; I haven't tested it)? I think that this is a better idea in principle for the sake of completeness - it feels icky deleting parts of the revision history.

georgemillo:
A new problem: when you've deleted and recovered an issue (or whatever), you'll have more than one `create` revisions in the revision history. (You won't have any `delete` revisions because they get, er, deleted). This causes the revision history page to crash because `revision.reify` and `revision.object` return `nil` when revision type is "create". I can't be arsed to give the complicated technical explanation, but this means that things break on our current revision hist page (since it assumes that the only time the event will be 'create' is for the oldest revision).

The least bad solution I can think of is to filter out these extra "create" actions at the controller level, but my first attempt at doing so is really crappy - replacing:


    class RevisionsController < ProjectScopedController
      # ...

      def index
        redirect_to action: :show, id: @record.versions.last.try(:id) || 0
      end

      def show
        # Use `reorder`, not `order`, to override Paper Trail's default scope
        @revisions = @record.versions.includes(:item).reorder("created_at DESC")
        @revision  = @revisions.find(params[:id])
        # ...
      end
    end

... with ... 

    class RevisionsController < ProjectScopedController
      # ...

      def index
        revisions  = @record.versions.includes(:item).reorder("created_at DESC")
        @revisions = revisions[0..-2].select { |v| v.event == "update" } << revisions[-1]
        redirect_to action: :show, id: @revisions.first.try(:id) || 0
      end

      def show
        # Use `reorder`, not `order`, to override Paper Trail's default scope
        revisions  = @record.versions.includes(:item).reorder("created_at DESC")
        @revisions = revisions[0..-2].select { |v| v.event == "update" } << revisions[-1]
        @revision  = @revisions.find { |r| r.id.to_s == params[:id] }
        # ...
      end
    end

... which is, of course, really shitty (and probably slow if they have a lot of revisions since I'm doing the filtering at the Ruby level instead of the SQL.

I'm sure this can be improved, but before I spend more time figuring it out, do you have any comments on the general approach of filtering out extra 'create' actions?

app/controllers/revisions_controller.rb
etdsoft: I think that heading to the Trash is going to be a relatively uncommon operation, so it may be OK to do a slightly convoluted SQL there.

config/routes.rb
etdsoft: At the end of the day we're acting on the Revision resource, so I think that works, maybe even use "/revisions/deleted" which is more meaningful, although I could be persuaded to keep trash as it's familiar enough.

app/controllers/revisions_controller.rb
georgemillo: Looks like it. Shouldn't be too hard to add the scope, but do you want to do this by monkey-patching `PaperTrail::Version` with a method called `set_project_scope`, or do it some other way? Monkey-patching is generally frowned upon but it might be the least bad option here.

(Note that we can't just call `PaperTrail::Version.default_scope` directly from the controller since `default_scope` is a protected method)

And do we need to modify anything in this PR for this, or can it wait until we merge into Pro?

app/controllers/revisions_controller.rb
```diff
@@ -40,6 +40,54 @@ def show
...
+      issue_revision = PaperTrail::Version.where(event: 'destroy', item_type: 'Note', item_id: object.issue_id).limit(1).first
+      # A destroy revision should always be present, but just in case.
+      if issue_revision
+        issue_object = issue_revision.reify
+        issue_object.node_id = Node.issue_library.id
+        issue_object.save
+        object.issue_id = issue_object.id
+        issue_revision.destroy
+      end
+    end
```
georgemillo: Proposal: create a class called `RecoverableVersion` or something similar which wraps an instance of `Version`, then move all this spaghetti code into the new class and keep the controller simple with something like:

```
def recover
  version = RecoverableVersion.find(params[:id])
  if version.recover
    track_recovered(version.object)
    flash[:info] = 'Item recovered'
  else
    flash[:error] = "Can't recover item."
  end
  redirect_to trash_path
end
```

@etdsoft thoughts?

app/controllers/revisions_controller.rb
georgemillo: Here's the best I can come up with:

    ids = [Evidence, Note].flat_map do |model|
      table_name = model.table_name
      PaperTrail::Version.where(event: 'destroy', item_type: model.to_s).
        joins("LEFT JOIN #{table_name} ON item_id=#{table_name}.id").
        where("#{table_name}.id IS NULL"). # avoid showing deleted objects
        select("distinct item_id, versions.id").
        pluck("versions.id")
    end
    PaperTrail::Version.where(id: ids)

But this isn't ideal because it makes multiple SQL queries. Do you think this is okay?

For posterity's sake, here's something I tried that didn't work (although I can't figure out why):

    PaperTrail::Version.where(event: 'destroy').
      joins("LEFT JOIN #{Note.table_name} ON item_id=#{Note.table_name}.id").
      joins("LEFT JOIN #{Evidence.table_name} ON item_id=#{Evidence.table_name}.id").
      where("#{Note.table_name}.id IS NULL").
      where("#{Evidence.table_name}.id IS NULL").
      select("distinct item_id, versions.*")

app/helpers/revisions_helper.rb
```diff
@@ -46,4 +46,67 @@ def link_to_conflicting_revision(record, revision)
...
+  def render_revision_object_info(revision)
+    revision_object = revision.reify
+    object_type = revision_object_type(revision,revision_object)
+    
+    [
+      render_revision_object_icon(revision_object,object_type),
+      render_revision_object_title(revision_object),
+      object_type.downcase,
+      revision_object_location(revision_object,object_type),
+    ].join(" ").html_safe
```
georgemillo: note to self: XSS issue on `.title`?

app/presenters/trashed_revision_presenter.rb
```diff
@@ -0,0 +1,73 @@
...
+        'at a Node which has since been deleted'
+      end
+    else
+      ''
+    end
+  end
+
+  def title
+    # Get title or content first characters.
+    if trashed_object.fields.empty?
```
etdsoft: here we're assuming that the trashed object has fields, in which case, using `HasFields#title` may just be a more consistent option.

app/controllers/revisions_controller.rb
etdsoft: It can wait until the merge to Pro, but we shouldn't forget!

app/controllers/revisions_controller.rb
etdsoft: Sounds good, please add some comments at the top of the class explaining the intent of `RecoverableVersion`.

etdsoft: Also, it seems that `Item recovered` isn't really providing much info about what just happened, so I tried to change that to at least use the class name (maybe we should even a link to the newly recovered model?)

Problem is that tests are coupled with this `Item recovered` string. See: https://codeship.com/projects/128584/builds/18089838

etdsoft:
To summarize:

- [ ] RESTful routes
- [ ] RecoverableVersion
- [ ] Using `HasFields#title`
- [ ] `Item recovered` flash message / specs coupling

config/routes.rb
georgemillo: Resolved by da199f5

app/presenters/trashed_revision_presenter.rb
georgemillo: > we're assuming that the trashed object has fields

A safe assumption for now, since the only items that will appear in the trash are `Issue`, `Note`, and `Evidence`, which all have fields.

But the problem with using `HasFields#title` here is that if there isn't a title we'll get the ugly string "This (class) doesn't provide a #[Title]# field", which ends up looking like this:

screen shot 2016-09-07 at 16 36 20

I think the current approach looks better. What do you think?

spec/models/recoverable_version_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    before do
+      @evidence = create(:evidence)
+      @note     = create(:note)
+      @issue    = create(:issue)
+
+      @deleted_evidence = create(:evidence)
+      @deleted_note     = create(:note)
+      @deleted_issue    = create(:issue)
+
+      # More issue/note hackery :(
```
georgemillo: Another proposal: whenever we have to do something like this, mark it with a unique comment (e.g. `FIXME - ISSUE/NOTE INHERITANCE`) - and add the same comment to any existing places with Issue/Note hackery, as we remember them. That way when we get around to sorting out the Issue/Note hierarchy issue, we can easily find all the places like this that will need updating.

app/controllers/revisions_controller.rb
```diff
@@ -46,13 +46,21 @@ def trash
   end
 
   def recover
-    version = RecoverableVersion.find(params[:id])
-    class_name = object.class.name.humanize
+    version    = RecoverableVersion.find(params[:id])
+    class_name = if version.object.is_a?(Note) && version.object.node_id == Node.issue_library.id
```
etdsoft: Would this be better hidden inside `RecoverableVersion`?

georgemillo: Good idea: done. But pleeeease can we fix this whole Issue/Note thing soon. It's giving me nightmares 😿

app/presenters/trashed_revision_presenter.rb
georgemillo: Nudge

app/controllers/revisions_controller.rb
etdsoft: I hear you, on the one hand, it is painful to know it's there. On the other, it's difficult to justify prioritising this over other stuff when it isn't really working, and some of the other fish we need to fry are a lot bigger.

georgemillo:
@etdsoft I think this is (finally) done now, apart from the HasField#title thing - see my comment [here](https://github.com/dradis/dradis-ce/pull/45#r77834661).

Before the trash feature was implemented, every item's revision history would consist **only** of "update" revisions, apart from the first revision, which would be "create". This meant that the revision history logic could be pretty simple - there's no revision history to show when there's only one revision (nothing to compare), and when there's more than one revision then we knew that the current revision's type would always be "update". (If the item got destroyed, a "destroy" revision would be created, but the user wouldn't have been able to see the revision history anyway... because the item was destroyed.)

Now that you can recover destroyed items (remember that internally PaperTrail saves "recover" revisions as `event=="create"`), we have to handle the case where the revision history contains items that aren't the very first revision, and aren't "update" events:

screen shot 2016-09-20 at 14 33 56

The way I've handled this is to **only** show links to revisions where the type is 'update', since these are the only revisions for which a diff makes sense. (We could show the revision page with no diff, and just the list of revisions that's in the screenshot above, but this wouldn't achieve very much - it's just a duplicate of the activity feed.)

What do you think?

app/presenters/trashed_revision_presenter.rb
etdsoft: Maybe we're better of changing the implementation of `HasFields#title` providing something more user friendly than just duplicating the code / logic in two different places (and potentially other views further down the line)?

georgemillo: Well I've always thought the `This note does not provide a #[Title]# field` is ugly and overly wordy . Is there any reason we can't change it to something simpler like `Note with no title` or `(No title)` or something like that?

etdsoft: Agreed, Maybe `(No #[Title]#) Lorem ipsum dolor...`. Think hinting at the `#[Title]#` field can be important to try to explain what is going on, and maybe providing a snippet of the content can also help locate the right one.

app/presenters/recoverable_revision_presenter.rb
```diff
@@ -0,0 +1,75 @@
...
+        "at " + h.link_to(trashed_object.node.label, trashed_object.node)
+      else
+        'at a Node which has since been deleted'
+      end
+    else
+      ''
+    end
+  end
+
+  def title
```
georgemillo: screen shot 2016-09-21 at 14 40 39

PR #46, opened 2016-08-24
Always render :services property as a table

etdsoft:
We want the 'services' property to always render as a table, so users
can rely on it to create their reports.

Unfortunately when a host has a single port associated with it, the
standard node properties mechanism would store :services as a simple
value instead of an array. This commit patches the rendering code to
make sure :services are always displayed in a table.


georgemillo:
looks good to me :+1:

PR #47, opened 2016-09-09
Move notes to other nodes

levionessa:
<span class="emoji-outer emoji-sizer"><span class="emoji-inner" style="background: url(chrome-extension://immhpnclomdloikkpcefncmfgjbkojmh/emoji-data/sheet_apple_64.png);background-position:35% 92.5%;background-size:4100%" title="wave"></span></span>  @etdsoft, @georgemillo:

Tried to use existing patterns as much as possible. 

See commit bodies for details on some choices (like creating a new plugin for the note moving modal).

Let me know if there's some other cases I should cover. Still not super familiar with how folks use Dradis day-to-day and haven't used it myself yet.

### DIFF
--- a/app/assets/javascripts/snowcrash.js
+++ b/app/assets/javascripts/snowcrash.js
@@ -14,6 +14,7 @@
 
 //= require snowcrash/plugins/jquery.breadcrums
 //= require snowcrash/plugins/jquery.treemodal
+//= require snowcrash/plugins/jquery.treemodalForNote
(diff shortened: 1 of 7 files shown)


levionessa:
PS when I first pulled down there were two failing tests in the shared examples for classes that behave like `ActiveModel`. I started to fix them until I saw there was a conditional to not run them in CI so thinking y'all now about those failures. Let me know if I can help clean that up along the way.

PR #48, opened 2016-09-17
Quick pasting feature

kulisu:
split from #15, users can add attachments quickly to focus on evidence writing.
### quick pasting image

> images would upload automatically from clipboard pasting while its checkbox is checked.
### auto increasing identity

> increase attachment identity likes `High-SQLi-Search-3.png`, `High-SQLi-Search-4.png` automatically while its checkbox is checked.
## REF:

[Better picture naming from Ctrl+C](https://github.com/securityroots/dradispro-tracker/issues/144)

> we're suggesting that you eliminate the .png/.jpg/.etc picture extensions,
> which is added by the code automatically and let the user focus on the picture name (rather than extension).

implemented, but seems wont work while pasting from clipboard.

> It would also be great if the name of the last saved image name would be remembered and the number after it would be automatically increased by 1:
> instead of always presenting a default name, which needs to be deleted and renamed every single time.

implemented, and it could be switch of/off by checkbox.

> We suggest you save the name of the picture in a cookie,
> which would make it browser based and each user/browser can have it's own remembered name.

implemented, but saves it in HTML5 [sessionstorage](http://www.w3schools.com/html/html5_webstorage.asp).

> this requires some work, but not much,
> but if you don't have time to actually implement this,
> you should at least remember the last picture name and present the test1.png to the user (rather than screenshot-XX.png).

implemented, as above.

### DIFF
--- a/app/assets/javascripts/snowcrash/behaviors.js.coffee
+++ b/app/assets/javascripts/snowcrash/behaviors.js.coffee
@@ -16,11 +16,46 @@ jQuery ->
       if confirm('Are you sure?')
         $.blueimp.fileupload.prototype.options.destroy.call(this, e, data);
 
+    # Start: a patch to implement #144
+    # https://github.com/securityroots/dradispro-tracker/issues/144
     paste: (e, data)->
-      $.each data.files, (index, file) ->
-        if (!file.name?)
-          file.name = prompt('Please provide a filename for the pasted image', 'screenshot-XX.png') || 'unnamed.png'
+      # High-SQLi-account_login-3.png
+      record = sessionStorage.getItem('attachments.last-upload')
 
+      $.each data.files, (index, file) ->
+        if (record == null)
+          # read file ext from jQuery.fileUpload
+          type = '.' + file.type.split('/').pop()
+        else
+          last = record.split(/[-\.]+/)
+          type = '.' + last.pop()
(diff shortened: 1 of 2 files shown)


etdsoft:
Hi @kulisu thanks for PR. However, I'm confused re: the paste from clipboard functionality.

This is offered by the browser and it is already available in Dradis out of the box, why do we need the new code?

This is from the current code:

paste_from_clipboard

kulisu:
Hello @etdsoft,

current pasting process is:

1. copy your image
2. paste image to browser
3. rename file name
4. **increase its index (optional)**
5. **press upload button**
6. and repeat above if there are more images

this PR add two switches and aim to let users pasting their image quickly.
when you enable `Quick Pasting`, a new process is:

1. copy your image
2. paste image to browser
3. rename file name
4. increase its index (optional)
5. ~~press upload button~~
6. and repeat above if there are more images

and `Auto Increasing` switch will increase filename's index automatically.
while an evidence has multi images to be uploaded, it can reduce lot of routines.

1. copy your image
2. paste image to browser
3. rename file name (optional for first time)
4. ~~increase its index (optional)~~
5. ~~press upload button~~
6. and repeat above if there are more images

untitled

MattBudz:
Closing due to inactivity.

PR #49, opened 2016-10-01
Move note

aapomm:
@etdsoft, this is ready for review.
- Used a modal copied from the `Node` moving modal.
- Decoupled the `:parent_id` field from the `Node` treemodal js.
- Added `Note` moving test.

Thanks!


georgemillo:
There seems to be a lot of wasted space above the main textarea:

screen shot 2016-12-08 at 20 41 21

I think this would look better if the four buttons in the top right were all on the same row, below the breadcrumbs. This would also reduce the risk of the layout getting messed up when you have a lot of nested nodes and the breadcrumbs reach far enough right to interfere with the buttons as they're currently placed.

georgemillo:
Aaaand I've just realised that the 'move' button in my screenshot is to move the node, not the note, so my comment is irrelevant to this PR. Nevermind!

etdsoft:
This has to do with screen width, on a bigger screen all the "action links" are on a single row

georgemillo:
There's a bug: I can't move a note to a subnode of the note's current node.

1. create nodes "A" and "B", where "A" is the parent of "B"
2. create a note under "A".
3. go to move the note, and expand "A" in the tree. The link for "B" is disabled and I can't click it.

This only happens for subnodes of the current node, not for other non-root nodes.

spec/features/note_moving_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    it "should redirect to note show path" do
+      expect(current_path).to eq(node_note_path(@node_1, current_note))
+    end
+  end
+
+
+
+  def create_node(attrs={})
+    create(:node, attrs)
```
georgemillo: I don't really see what advantage this method provides over just calling `create(:node, ...)` directly?

aapomm:
@georgemillo updated code and ready for another review.

georgemillo:
Looks good now :+1:

app/views/nodes/modals/_move.html.erb
```diff
@@ -1,5 +1,5 @@
 <div id="modal_move_node" class="modal hide" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true">
-  <%= simple_form_for @node, html: {class: 'form-horizontal modal-node-selection-form', data:{ node_id: @node.id }} do |f| %>
+  <%= simple_form_for @node, html: {class: 'form-horizontal modal-node-selection-form', data: {node_id: @node.id, hidden_field: '#node_parent_id'}} do |f| %>
```
etdsoft: I'd have changed this myself, but this PR is from your own repo still, please review the styling of the `{`:

     data: { node_id: @node.id, hidden_field: '#node_parent_id' } }

aapomm:
@etdsoft finished requested changes.

PR #51, opened 2016-10-17
Capybara is acting on CI - try screen size

etdsoft:
Capybara is detecting the modal as being displayed on top of the rest of
the interface and then throwing

`
Poltergeist detected another element with CSS selector
`

Not completely sure this is the solution, but worth giving it a try.

See:
  https://makandracards.com/foxsoft/38755-poltergeist-detected-another-element-with-css-selector


etdsoft:
Closing as this was fixed by @rachkor #52

PR #61, opened 2016-12-07
Use rails to filter the versions instead

aapomm:
Problem:
- mysql performs wonky prioritization when using `distinct` or `group`

Solution:
- Delegate some filtering logic to rails instead to make the code a lot more database agnostic.


georgemillo:
Looks good to me :+1:

PR #65, opened 2016-12-08
Add issue trash recover spec

aapomm:
- Added a test to check recoverable versions of edited issues in trash.


georgemillo:
looks good to me :+1:

PR #66, opened 2016-12-09
Issue refresh

aapomm:
- Added refresh button in the issues toolbar.

### DIFF
--- a/app/views/issues/_toolbar.html.erb
+++ b/app/views/issues/_toolbar.html.erb
@@ -37,6 +37,10 @@
     </ul>
   </div>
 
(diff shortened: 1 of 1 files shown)


georgemillo:
The refresh button works, but I found an unrelated bug in the toolbar: the delete button isn't shown/hidden when you click on the area around the 'select all' checkbox (as opposed to the checkbox itself):

deletebtn

georgemillo:
Other than that everything seems fine: :+1:

etdsoft:
Good catch, do you mind adding the select-all issue to the Backlog?

PR #67, opened 2016-12-19
Select all issues bug

aapomm:
[Trello card](https://trello.com/c/CGrKsrPV/235-when-i-select-all-issues-i-want-to-show-hide-the-trash-button-when-i-click-anywhere-on-the-button-not-just-on-the-checkbox-so-i-)

- Fixed unresponsive select all button click
- Added issue table toolbar tests


georgemillo:
no complaints 👍

georgemillo:
Wait, I didn't notice it earlier, but CI is failing. However, the problem also seems to be occurring on `master`: https://app.codeship.com/projects/128584/builds/21068243

spec/features/issue_trash_recover_spec.rb
```diff
@@ -15,6 +15,8 @@
     visit "/trash"
     click_link 'Recover'
 
+    sleep(1)
```
etdsoft: What's the source cause of this behavior? Is there a way around introducing a 1 second delay in the process? 
/cc @georgemillo @aapomm @dormi

aapomm: @etdsoft I haven't confirmed this yet, but I think the SQL query in `RecoverableRevisions` fails to order the issues if they are made within the same second. One solution is to refactor that query. I'm not sure why this test fails on this unrelated branch, though.

etdsoft: I think we should get to the bottom of this, adding a spurious 1sec delay is only making our life worse in the long run.

georgemillo: @aapomm was it passing on your local machine but failing on CI? I struggled with a similar issue recently, I can't remember where exactly, but I think the cause was what you describe: an `order` query messing up if two records were created within the same second (or maybe microsecond?) as each other. I fixed it in my tests by artificially setting `created_at` to a couple of seconds ago when the record(s) are created.

aapomm: @etdsoft I agree. Though, I would suggest we fix this in a separate and dedicated branch to keep things atomic.

@georgemillo on my machine, sometimes it passed and sometimes it didn't (which made debugging difficult). Also, is this line (https://github.com/dradis/dradis-ce/blob/master/spec/models/recoverable_revision_spec.rb#L95) the one you are referring to?

aapomm:
@etdsoft I updated with your requested changes.

app/models/recoverable_revision.rb
```diff
@@ -28,7 +28,7 @@ def self.all
       versions = PaperTrail::Version.where(event: 'destroy', item_type: model.to_s).
         joins("LEFT JOIN #{table_name} ON item_id=#{table_name}.id").
         where("#{table_name}.id IS NULL"). # avoid showing deleted objects
-        order(created_at: :desc)
+        order(id: :desc)
```
georgemillo: I understand why this change is being made, but I think it's worth adding a comment to explain why.

georgemillo:
other than my point about adding a code comment: 👍

PR #68, opened 2016-12-21
Upgrade to Rails 5

georgemillo:


app/assets/javascripts/cable.js
```diff
@@ -0,0 +1,13 @@
+// Action Cable provides the framework to deal with WebSockets in Rails.
```
georgemillo: Is there any point adding these files if we're not actually using ActionCable?

app/jobs/application_job.rb
```diff
@@ -0,0 +1,2 @@
+class ApplicationJob < ActiveJob::Base
```
georgemillo: Since we have `ApplicationJob` now, maybe it would make sense to merge `BaseWorker` into this class. While we're at it we could also replace the weird `process -> process_delegate` setup with the more conventional `ActiveJob::Base#perform` and an AJ callback

app/assets/javascripts/cable.js
etdsoft: There is, because they're added by default by the Rails new app generator. It will make comparing / diffing for the next upgrades easier.

app/jobs/application_job.rb
etdsoft: It is the right time, but probably scope creep. Let's merge this PR and then open a new flow for standardising the ActiveJob stuff. Can you please add the card to Trello?

etdsoft:
It seems that the only thing missing before merge is clarifying the `create_users` situation... @georgemillo ?

georgemillo:
re: the create_users migration, I think the real issue was that the Rails 5 upgrade broke our custom method `User.column`, but I've fixed that (hackily, but then the original method was hacky 😉 ) in 4510c55bda

The question is do we want to keep this migration in CE, and by extension keep the `users` table around in the CE database, even though CE doesn't actually use the `users` table? It doesn't seem to make a difference whether or not the table is present in the DB, but if it's not in CE then we might accidentally remove it from Pro every time we merge... which has happened [before](15342ca178). I'd suggest keeping the migration in.

PR #70, opened 2017-01-18
Generalize configurations styling with table-ajax-inputs class

aapomm:
Requirement for Pro PR162.

### DIFF
--- a/app/assets/stylesheets/snowcrash/modules.scss
+++ b/app/assets/stylesheets/snowcrash/modules.scss
@@ -81,6 +81,7 @@
     }
   }
 }
+
 .content {
   padding: 1em;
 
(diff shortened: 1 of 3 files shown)


aapomm:
@dormi yes. Styles shared by sheets within `modules/` should be placed within `modules.scss`

etdsoft:
> Should these styles be placed in modules.scss ? What styles are supposed to go in there? The ones that are shared by many views ?

I'm not a fan of the current structure of the Snowcrash assets, as you mention "module" seems to be something that is reusable in a couple of places (like the `table-ajax-input` stuff). But the reality today is that we have page-specific classes under `./modules/<page>`. I think that a better structure would be:
```
./stylesheets
|
 \modules/
 \pages/
```

Where we have re-usable classes in `./modules/` and page-specific classes in `./pages`.

If we had such structure, then it would make sense to move these PR's contents to something like `./modules/table_ajax_input.scss`. But until we do, dropping it in `modules.scss` is a good option.

PR #71, opened 2017-01-18
[Rails 5] A few things fell through the cracks the 1st time around

etdsoft:
See https://github.com/dradis/dradis-ce/pull/68


georgemillo:
;)

PR #73, opened 2017-01-20
Fix CSS styling of sidebar collection items

etdsoft:
I just realised that a piece of code dealing with the CSS classes that list items of the sidebar were supposed to get wasn't actually working as expected.

Independently, there is a recurring request for being able to edit from the sidebar list instead of having to click on the item, then click on edit. Thoughts on this approach?

list-hover

* Adding the :hover effect
* Adding the edit link

And along the same lines... how do we allow direct edit from the Issues table? Any of these two options any good? Is there a better way?

Edit link in its own column:

table-hover-01

Edit link in Title column:

table-hover-02

### DIFF
--- a/app/helpers/snowcrash_helper.rb
+++ b/app/helpers/snowcrash_helper.rb
@@ -40,4 +40,29 @@ module SnowcrashHelper
     yield presenter if block_given?
     presenter
   end
+
+
+  def icon_for_model(model, icon_class, extra_class = nil)
+
+    css =  ['fa']
+    css << icon_class
+    css << extra_class if extra_class
+
+    options = { class: css.join(' ') }
+    tag     = nil
+
+    case model
+    when Evidence
+      tag = model.issue.tags.first
+    when Issue
+      tag = model.tags.first
+    end
+
+    if tag
+      options[:style] = "color: #{tag.color}"
+    end
+
+    content_tag :i, nil, options
+  end
+
 end
--- a/app/helpers/snowcrash_helper.rb
+++ b/app/helpers/snowcrash_helper.rb
@@ -42,7 +42,7 @@ module SnowcrashHelper
   end
 
 
-  def icon_for_model(model, icon_class, extra_class = nil)
+  def colored_icon_for_model(model, icon_class, extra_class = nil)
 
     css =  ['fa']
     css << icon_class
--- a/app/views/evidence/_evidence.html.erb
+++ b/app/views/evidence/_evidence.html.erb
@@ -1,4 +1,9 @@
-<%= content_tag :div, id: "#{dom_id(evidence)}_link", class: ('list-item' << ' active' if (request.format.html? and @evidence == evidence)) do %>
+<%
+  item_css = ['list-item']
+  item_css << ' active' if (request.format.html? and @evidence == evidence)
+%>
+
+<%= content_tag :div, id: "#{dom_id(evidence)}_link", class: item_css.join do %>
   <div class="expansion-container">
     <%= link_to node_evidence_path(@node, evidence) do %>
       <% if tag = evidence.issue.tags.first %>
--- a/app/views/evidence/_evidence.html.erb
+++ b/app/views/evidence/_evidence.html.erb
@@ -9,13 +9,7 @@
       <i class="fa fa-pencil"></i> Edit
     <% end %>
     <%= link_to node_evidence_path(@node, evidence) do %>
-      <%
-        options = { class: 'fa fa-flag list-item-icon' }
-        if tag = evidence.issue.tags.first
-          options[:style] = "color: #{tag.color}"
-        end
-      %>
-      <%= content_tag :i, nil, options%>
+      <%= icon_for_model(evidence, 'fa-flag', 'list-item-icon') %>
       <%= evidence.issue.title %>
     <% end %>
   </div>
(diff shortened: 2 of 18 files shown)


aapomm:
For the sidebar, I'd like the edit button to be closer to the issue title. Probably an edit pen icon before the title (this makes it consistent for issues that span 1+ lines).

For the issues table, what if we put dedicated columns at the end of the row for edit and delete?

dormi:
For the issues table, I would prefer to also add the "edit" text with the pen icon.

rachkor:
**Sidebar**
I'm not sure how to add it to the left of the title in the sidebar (since we already have the "bug" icon). I like how it looks now with the smaller title but would like to make sure it looks ok with a long title that stretches into 2 lines in the sidebar. 

**Issues Table**
I like the second option (edit link in title column) better. I don't think we need the "edit" text there as it's a pretty universal icon these days (check out the pen on the comments here on GH!).

aapomm:
For better visualization, here's what I imagined:
**Sidebar**
sidebar
**Issues Table**
issuestable

My reasoning behind this is that, this is where my mind _expects_ the buttons to be (even when using different apps). Let me know what you think! 😄

etdsoft:
**Re: IssuesTable**, I think it makes sense, specially considering the Trash table already has actions on the right-hand side, and so do all the tables in Mintcreek.

**Re: Sidebar**, I think that pen + bug icon on the left are a bit over the top and starts feeling crowded. Maybe we could get fancy and replace the bug with a pen!


sidebar-03
sidebar-04

etdsoft:
table-hover-03

etdsoft:
sidebar-05

rachkor:
@etdsoft re: Sidebar, I love the latest one where the bug is replaced by the pen. 👍 

For the Issues Table, the icons on the right make sense in the screenshot when there is a lot of whitespace to the right, but what about when the table view is full? (example from the OSCP compliance pack project below)

screen shot 2017-01-23 at 8 31 03 am

etdsoft:
I believe they should work, the action links are on a fixed-width, right-aligned column at the end:

issues-table

I've pushed the changes so you can git pull and take that screenie again.

app/views/evidence/_evidence.html.erb
```diff
@@ -1,11 +1,21 @@
...
+    <% end %>
     <%= link_to node_evidence_path(@node, evidence) do %>
-      <% if tag = evidence.issue.tags.first %>
-        <i class="fa fa-flag" style="color: <%= tag.color %>"></i>
-      <% else %>
-        <i class="fa fa-flag"></i>
-      <% end %>
+      <%
+        options = { class: 'fa fa-flag list-item-icon' }
+        if tag = evidence.issue.tags.first
```
aapomm: I'm not a fan of assignments in conditions because of the readability. I know that this can be found all over the codebase, but I think we should reduce the usage of this. What do you think?

app/views/issues/_issue.html.erb
```diff
@@ -1,8 +1,22 @@
...
+<%= content_tag :div, id: dom_id(issue), class: item_css.join do %>
   <div class="expansion-container">
+    <%= link_to edit_issue_path(issue), class: 'list-item-edit' do %>
+      <i class="fa fa-pencil"></i> Edit
+    <% end %>
     <%= link_to issue, class: 'view-toggle-off', data: {target: "##{dom_id(issue)}_content"} do %>
-      <%= tag_icon_for(issue) %>
+      <%= tag_icon_for(issue, 'list-item-icon') %>
       <%= issue.title %>
+      <!--
```
aapomm: what's this code for?

app/views/evidence/_evidence.html.erb
etdsoft: Agreed, too much logic in this view, I've extracted it to a helper method and also avoided the assignment in conditional.

etdsoft:
@aapomm all requested changes have been made

etdsoft:
@dormi can I please get a sanity check review on this before it's merged? Thanks!

app/helpers/snowcrash_helper.rb
```diff
@@ -40,4 +40,29 @@ def present(object, klass = nil)
     yield presenter if block_given?
     presenter
   end
+
+
+  def icon_for_model(model, icon_class, extra_class = nil)
```
dormi: as a newbie, I think that a name that related this method to adding color to the icon would have helped me. Mey be something like `color_icon`,  `paint_icon`, `paint_icon_from_tag`, `colored_icon`...

dormi:
Also, in the issue list, I think that it may be not enough clear that on hover, we have two links with two different destinations: edit issue vs show issue
I've tried adding some border to the edit link to give it the feeling that it's a button:
edit_link

app/helpers/snowcrash_helper.rb
etdsoft: I like `colored_icon_for_model`

etdsoft:
@dormi  can you upload a quick GIF of the same, but addting a `pull-right` class? I'm starting to think we should combine the concepts and a) remove the bug Icon, and have the links on the right-hand side. 

Also, can you upload a sample Nessus file to get some real-life titles and hopefully some that span across two lines?

dormi:
Moving the edit link to the right, but keeping the icons:
edit_to_the_right_and_keep_icon

dormi:
edit_to_the_right_and_keep_icon

Moving the link to the right just with `pull right` means that we move the issue title text when the link is displayed. I don't really like that.

dormi:
edit_to_the_right_do_not_keep_icon

PR #77, opened 2017-01-30
Attachments api

dormi:
Adding endpoints to the API to manage attachments.

[Trello Card](https://trello.com/c/F0CaIWJQ/267-1-when-using-the-api-i-want-to-have-endpoints-that-allow-handling-attachments-so-i-can-perform-basic-crud-operations-with-attach)

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -43,7 +43,7 @@ class AttachmentsController < ProjectScopedController
     attachment  = Attachment.find(filename, conditions: { node_id: @node.id } )
     attachment.close
     new_name    = CGI::unescape(params[:rename])
-    destination = Attachment.pwd.join(@node.id, new_name).to_s
+    destination = Attachment.pwd.join(@node.id.to_s, new_name).to_s
 
     if !File.exist?(destination) && !destination.match(/^#{Attachment.pwd}/).nil?
       File.rename attachment.fullpath, destination
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -61,7 +61,7 @@ class AttachmentsController < ProjectScopedController
     @attachment  = Attachment.find(filename, conditions: { node_id: @node.id } )
     send_options = { filename: @attachment.filename }
 
-    # Figure out the best way of displaying the file (by default send the it as
+    # Figure out the best way of displaying the file (by default send it as
     # an attachment).
     extname = File.extname(filename)
     send_options[:disposition] = 'attachment'
@@ -90,10 +90,10 @@ class AttachmentsController < ProjectScopedController
   # the :node_id parameter.
   def destroy
     filename = params[:id]
-    
+
     @attachment = Attachment.find(filename, conditions: { node_id: @node.id} )
     @attachment.delete
-    
+
     render json: { success: true }
   end
 
@@ -102,7 +102,7 @@ class AttachmentsController < ProjectScopedController
   # we are working with. This filter sets the @node instance variable if the
   # give :node_id is valid.
   def find_or_initialize_node
-    begin 
+    begin
       @node = Node.find(params[:node_id])
     rescue
       redirect_to root_path, alert: 'Node not found'
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -39,7 +39,7 @@ class AttachmentsController < ProjectScopedController
   # It is possible to rename attachments and this function provides that
   # functionality.
   def update
-    filename    = params[:id]
+    filename    = params[:filename]
     attachment  = Attachment.find(filename, conditions: { node_id: @node.id } )
     attachment.close
     new_name    = CGI::unescape(attachment_params[:filename])
@@ -56,7 +56,7 @@ class AttachmentsController < ProjectScopedController
   # displayed inline. By default the <tt>Content-disposition</tt> will be set to
   # +attachment+.
   def show
-    filename = params[:id]
+    filename = params[:filename]
 
     @attachment  = Attachment.find(filename, conditions: { node_id: @node.id } )
     send_options = { filename: @attachment.filename }
@@ -89,7 +89,7 @@ class AttachmentsController < ProjectScopedController
   # attachment's file name in the :id parameter and the corresponding node in
   # the :node_id parameter.
   def destroy
-    filename = params[:id]
+    filename = params[:filename]
 
     @attachment = Attachment.find(filename, conditions: { node_id: @node.id} )
     @attachment.delete
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -108,23 +108,4 @@ class AttachmentsController < ProjectScopedController
       redirect_to root_path, alert: 'Node not found'
     end
   end
-
-  # Obtain a suitable attachment name for the recently uploaded file. If the
-  # original file name is still available, use it, otherwise, provide count-based
-  # an alternative.
-  def get_name(args={})
-    original = args.fetch(:original)
-
-    if @node.attachments.map(&:filename).include?(original)
-      attachments_pwd = Attachment.pwd.join(@node.id.to_s)
-
-      # The original name is taken, so we'll add the "_copy-XX." suffix
-      extension = File.extname(original)
-      basename  = File.basename(original, extension)
-      sequence  = Dir.glob(attachments_pwd.join("#{basename}_copy-*#{extension}")).collect { |a| a.match(/_copy-([0-9]+)#{extension}\z/)[1].to_i }.max || 0
-      "%s_copy-%02i%s" % [basename, sequence + 1, extension]
-    else
-      original
-    end
-  end
 end
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -43,7 +43,7 @@ class AttachmentsController < ProjectScopedController
     attachment  = Attachment.find(filename, conditions: { node_id: @node.id } )
     attachment.close
     new_name    = CGI::unescape(params[:rename])
-    destination = Attachment.pwd.join(params[:node_id], new_name).to_s
+    destination = Attachment.pwd.join(@node.id, new_name).to_s
 
     if !File.exist?(destination) && !destination.match(/^#{Attachment.pwd}/).nil?
       File.rename attachment.fullpath, destination
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -14,7 +14,7 @@ class AttachmentsController < ProjectScopedController
   def create
     uploaded_file = params.fetch('attachment_file', params.fetch('files', []).first)
 
-    attachment_name = get_name(original: uploaded_file.original_filename)
+    attachment_name = Attachment.available_name(@node, original: uploaded_file.original_filename)
 
     @attachment = Attachment.new(attachment_name, node_id: @node.id)
     @attachment << uploaded_file.read
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -42,7 +42,7 @@ class AttachmentsController < ProjectScopedController
     filename    = params[:id]
     attachment  = Attachment.find(filename, conditions: { node_id: @node.id } )
     attachment.close
-    new_name    = CGI::unescape(params[:rename])
+    new_name    = CGI::unescape(attachment_params[:filename])
     destination = Attachment.pwd.join(@node.id.to_s, new_name).to_s
 
     if !File.exist?(destination) && !destination.match(/^#{Attachment.pwd}/).nil?
@@ -107,5 +107,9 @@ class AttachmentsController < ProjectScopedController
     rescue
       redirect_to root_path, alert: 'Node not found'
     end
+  end
+
+  def attachment_params
+    params.require(:attachment).permit(:filename)
   end
 end
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -12,7 +12,7 @@ class AttachmentsController < ProjectScopedController
   # Create a new attachment for a give :node_id using a file that has been
   # submitted using an HTML form POST request.
   def create
-    uploaded_file = params.fetch('attachment_file', params.fetch('files', []).first)
+    uploaded_file = params.fetch(:attachment_file, params.fetch(:files, []).first)
 
     attachment_name = Attachment.available_name(@node, original: uploaded_file.original_filename)
 
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -39,7 +39,7 @@
 #   attachment = Attachment.new("images/my_image.gif", :node_id => 1)
 #
 # This will create an instance of an attachment that belongs to node with ID = 0
-# Nothing has bee saved yet
+# Nothing has been saved yet
 #
 #   attachment.save
 #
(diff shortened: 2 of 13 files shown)


engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+      def index
+        @attachments = @node.attachments.each(&:close)
+      end
+
+      def show
+        filename = params[:id]
+        begin
+          @attachment = Attachment.find(filename, conditions: { node_id: @node.id } )
+        rescue
+          raise ActiveRecord::RecordNotFound, "Couldn't find attachment with id '#{params[:id]}'"
```
aapomm: why not just render 404?

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+        end
+      end
+
+      def destroy
+        filename = params[:id]
+
+        @attachment = Attachment.find(filename, conditions: { node_id: @node.id} )
+        @attachment.delete
+        
+        render_successful_destroy_message
```
aapomm: I think it's better to also return the filename of the deleted attachment.

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+      def update
+        filename    = params[:id]
+        attachment  = Attachment.find(filename, conditions: { node_id: @node.id } )
+        attachment.close
+
+        begin
+          new_name    = CGI::unescape(params[:attachment][:filename])
+          destination = Attachment.pwd.join(params[:node_id], new_name).to_s
+
+          if !File.exist?(destination) && !destination.match(/^#{Attachment.pwd}/).nil?
```
aapomm: What happens if this condition is unmet? Wouldn't `@attachment` be unset?

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+      end
+
+      def attachment_params
+        params.require(:files).permit()
+      end
+
+      # Obtain a suitable attachment name for the recently uploaded file. If the
+      # original file name is still available, use it, otherwise, provide count-based
+      # an alternative.
+      def get_name(args={})
```
aapomm: I feel that this should have a more meaningful name since, for the most part, you're not actually "getting" a name. Also, what's the reasoning behind using a hash argument?

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+      before_action :set_node
+
+      skip_before_action :json_required, :only => [:create]
+
+      def index
+        @attachments = @node.attachments.each(&:close)
+      end
+
+      def show
+        filename = params[:id]
```
aapomm: You can replace the `:id` param name with `:filename` instead. `resources :attachments, param: :filename`

dormi: 👍

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
dormi: I wanted the error to be handled here:
https://github.com/dradis/dradis-ce/blob/master/engines/dradis-api/app/controllers/dradis/ce/api/api_controller.rb#L14
and render all errors from a single method.

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
dormi: In the rest of API controllers, we are returning this error message
(https://github.com/dradis/dradis-ce/blob/master/engines/dradis-api/app/controllers/dradis/ce/api/api_controller.rb#L83)
I think that we keep it like this, or change it everywhere. I would keep it.

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
dormi: You are right! I am missing a case here: when the file with the new name already exists. We should fail with a 422

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
dormi: This method was copies from https://github.com/dradis/dradis-ce/blob/master/app/controllers/attachments_controller.rb#L115

Since I was duplicating code, I've moved it to the `Attachment` model

Also, I have renamed it to `avaialble_name`

The point of the hash argument I guess is to keep it easy extendable.

dormi:
@aapomm I've added comments/changes to your review.

Also, something we may discuss:

In `POST /api/nodes/X/attachments/:filename` (the `create` endpoint), now we are expecting a "normal" http post (like a form).
I think this way it's easier to upload files with the API.
In all the other API  controllers, in the `create` method we are passing info inside a JSON payload.
I skipped passing file info inside a JSON for this endpoint, I think it makes things simpler.
What do you think? @etdsoft comments on this?

engines/dradis-api/spec/requests/dradis/ce/api/v1/attachments_spec.rb
```diff
@@ -0,0 +1,308 @@
...
+        expect(retrieved_attachment["link"]).to eq "/nodes/#{node.id}/attachments/image.png"
+      end
+    end
+
+    describe "POST /api/nodes/:node_id/attachments" do
+      let(:post_attachment) {
+        file = fixture_file_upload(Rails.root.join('spec/fixtures/files/rails.png'))
+
+        params = {files: []}
+        params[:files] << file
```
aapomm: you can merge these two lines to `params = {files: [file]}`

engines/dradis-api/spec/requests/dradis/ce/api/v1/attachments_spec.rb
```diff
@@ -0,0 +1,308 @@
...
+
+        expect(Dir["#{node_attachments}/*"].count).to eq(2)
+      end
+
+      it "returns JSON information about the attachments" do
+        file1 = fixture_file_upload(Rails.root.join('spec/fixtures/files/rails.png'))
+        file2 = fixture_file_upload(Rails.root.join('spec/fixtures/files/rails.png'))
+
+        params = {files: []}
+        params[:files] << file1 << file2
```
aapomm: you can instantiate this in one line

app/models/attachment.rb
```diff
@@ -148,6 +148,25 @@ def self.pwd
...
+  def self.available_name(node, args={})
+    original = args.fetch(:original)
+
+    if node.attachments.map(&:filename).include?(original)
+      attachments_pwd = Attachment.pwd.join(node.id.to_s)
+
+      # The original name is taken, so we'll add the "_copy-XX." suffix
+      extension = File.extname(original)
+      basename  = File.basename(original, extension)
+      sequence  = Dir.glob(attachments_pwd.join("#{basename}_copy-*#{extension}")).collect { |a| a.match(/_copy-([0-9]+)#{extension}\z/)[1].to_i }.max || 0
```
aapomm: You're using `collect` and `map`(L157) here even though they're the same method. I feel that you should pick one and be consistent.

engines/dradis-api/spec/requests/dradis/ce/api/v1/attachments_spec.rb
```diff
@@ -0,0 +1,308 @@
...
+        retrieved_filenames = retrieved_attachments.map{ |json| json['filename'] }
+        expect(retrieved_filenames).to match_array(@attachments)
+      end
+
+      it "returns JSON information about attachments" do
+        attachment_0 = retrieved_attachments.detect { |n| n["filename"] == "image0.png" }
+        attachment_1 = retrieved_attachments.detect { |n| n["filename"] == "image1.png" }
+        attachment_2 = retrieved_attachments.detect { |n| n["filename"] == "image2.png" }
+
+        expect(attachment_0.keys).to match_array %w[filename link]
```
aapomm: I think it's enough to expect these hashes to be `eq` the expected hash (something like `expect(attachment_n).to eq(expected_hash)`), instead of asserting their keys and then asserting the values of each key. What do you think?

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      def index
+        @attachments = @node.attachments.each(&:close)
+      end
+
+      def show
+        filename = params[:filename]
+        begin
+          @attachment = Attachment.find(filename, conditions: { node_id: @node.id } )
+        rescue
+          raise ActiveRecord::RecordNotFound, "Couldn't find attachment with filename '#{params[:filename]}'"
```
aapomm: you're using `params[:filename]` here even though you already assigned it to the `filename` variable.

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+        if @attachments.any? && @attachments.count == uploaded_files.count
+          render status: 201
+        else
+          render status: 422
+        end
+      end
+
+      def update
+        filename    = params[:filename]
+        attachment  = Attachment.find(filename, conditions: { node_id: @node.id } )
```
aapomm: I don't think you need to assign the `filename` variable here. You can go straight to `attachment  = Attachment.find(params[:filename], conditions: { node_id: @node.id } )`

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+        rescue
+          @attachment = attachment
+          render status: 422
+        end
+      end
+
+      def destroy
+        filename = params[:filename]
+
+        @attachment = Attachment.find(filename, conditions: { node_id: @node.id} )
```
aapomm: I think you can go straight into using `params[:filename]` here instead of assigning it.

engines/dradis-api/spec/requests/dradis/ce/api/v1/attachments_spec.rb
dormi: 👍

engines/dradis-api/spec/requests/dradis/ce/api/v1/attachments_spec.rb
dormi: ✅

engines/dradis-api/spec/requests/dradis/ce/api/v1/attachments_spec.rb
dormi: ✅

app/models/attachment.rb
dormi: ✅

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
dormi: ✅

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
dormi: ✅

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
dormi: ✅

dormi:
@aapomm added those changes

@aapomm @etdsoft Please remember to give your opinion about not using a json payload on the `create` action. (my opinion obviously is :+1:)

aapomm:
@dormi I think the approach to the `create` action is 👍

etdsoft:
The code looks good, re `Attachments#create` it seems that part of the issue is that we're accepting multi-file uploads via the API to a single-resource endpoint, in that case using multi-part seems like the way to go, but maybe we should implement a one-file-per-request policy? In which case maybe a binary blob is done?

https://philsturgeon.uk/api/2016/01/04/http-rest-api-file-uploads/

However, there is some value in being able to conform to the standard and for example use `curl` to exercise the `#create` endpoint.

So I think JSON doesn't make sense, and what we need to decide is multi-upload vs single-upload.

app/controllers/attachments_controller.rb
```diff
@@ -14,7 +14,7 @@ def index
   def create
     uploaded_file = params.fetch('attachment_file', params.fetch('files', []).first)
```
etdsoft: Should we take this PR and clean up both controllers? E.g. why `'files'` instead of `:files`?

app/models/attachment.rb
```diff
@@ -148,6 +148,25 @@ def self.pwd
     AttachmentPwd
   end
 
+  # Obtain a suitable attachment name for a recently uploaded file. If the
+  # original file name is still available, use it, otherwise, provide count-based
+  # an alternative.
+  def self.available_name(node, args={})
```
etdsoft: It is a good extraction but the application's `AttachmentsController` has a copy of this method? I think the idea was to remove it from there too?

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+        render_successful_destroy_message
+      end
+
+      private
+
+      def set_node
+        @node = Node.find(params[:node_id])
+      end
+
+      def attachment_params
```
etdsoft: I like that we're embracing strong parameters here, but we should go all-in, some places are currently using `params` and some are using `attachment_params` I think let's get rid of `params` and use the sanitised version throughout.

I realise the app's controller isn't using Strong Params yet, but maybe after this new controller is cleaned up we can make the app's controller use it too?

spec/factories/attachments.rb
```diff
@@ -0,0 +1,15 @@
+FactoryGirl.define do
+  factory :attachment do
+    skip_create
+    
+    sequence(:filename) { |n| "image#{n}.png" }
+    node { Node.issue_library }
```
etdsoft: I think that just adding `node` will let FactoryGirl figure out the rest, we don't need it to be the `.issue_library`

app/models/attachment.rb
dormi: ✅

engines/dradis-api/app/controllers/dradis/ce/api/v1/attachments_controller.rb
dormi: You are right, I really wasn't using them.
Strong params prevent mass assignment when using ActiveModel.
Our `attachment` model does not inherit from ActiveModel, so I am not sure if this applies.
Even that I tried to enforce it in the `update` action, just to have it there for the future.
In the create action we are dealing with uploaded files, so I think it was not the right place to use them.
✅

spec/factories/attachments.rb
dormi: ✅

engines/dradis-api/config/routes.rb
```diff
@@ -6,6 +6,9 @@
         resources :nodes do
           resources :evidence
           resources :notes
+          constraints(:filename => /.*/) do
+            resources :attachments, param: :filename
```
dormi: ✅

app/controllers/attachments_controller.rb
dormi: ✅

PR #79, opened 2017-01-31
Use the new visual style for the attachment list

etdsoft:
Doing some unrelated testing I realised that now the Attachments are the only list in the sidebar without the :hover styling.

This screenie shows the new behaviour vs the existing one (where links are always shown):

attachments-01

Now this change is not as straightforward because we're using `jquery-fileupload` for the upload box, and that plugin needs / assumes that uploaded files will be rendered as a table (maybe not 100% true, but all of the code we've got at the moment assumes so).

This isn't too bad if we've got some attachments in the node and decide to drag'n'drop a few more for upload like this:

attachments-02

However it isn't as pretty after we upload the attachments because `jquery-fileupload` takes care of re-rendering this already-uploaded attachments as rows of the table in the old style, so we'd have situation where both styles are mixed.

The video bellow shows the initial state of some new-style attachments and some recently uploaded, old-style ones coexisting. After a browser refresh, everything goes back to new-style:

attachments-03

And the final piece of the puzzle is the `Nothing yet` placeholder that we use for Notes and Evidence. Because we're relying completely on `jquery-fileupload` to handle the UI refresh, we can end up with a view that shows both the `Nothing yet` message AND some recently uploaded attachments.

attachments-04

So a few questions:

* Does it make sense to use the new :hover style here?
* Can we live with the quirks listed above or should we dig deeper into `jquery-fileupload` to figure everything out and make the UI smoother?

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -94,7 +94,7 @@ class AttachmentsController < ProjectScopedController
     @attachment = Attachment.find(filename, conditions: { node_id: @node.id} )
     @attachment.delete
     
-    render json: { success: true }
+    redirect_to node_path(@node), notice: 'Attachment deleted'
   end
 
   private
--- a/app/views/attachments/_attachment_box.html.erb
+++ b/app/views/attachments/_attachment_box.html.erb
@@ -7,44 +7,48 @@
     </div>
   </div>
 
+  <div id="attachments" class="note-list">
+    <% if node.attachments.any? %>
+
+      <% node.attachments.each do |attachment| %>
+        <div class="list-item">
+          <div class="expansion-container">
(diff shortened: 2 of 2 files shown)


rachkor:
@etdsoft I don't mind the existing style where the links are always shown. I think of the attachments as a different type of list since we're not adding any functionality (e.g. removing an extra click to get to `edit`) by adding the `:hover`. I'd rather stick with the existing style than introduce the quirks like "(nothing yet)" when there is already a populated list.

leesoh:
As an aside, I'd love it if the rest of the links had `:hover` removed. Being able to quickly ctrl + click and open a bunch of issues is more useful to me than the minimal aesthetic. As it is now, I have to mouseover and then move to wherever the edit button has popped up.

brianp:

PR #80, opened 2017-01-31
Enable code highlight via the %%{{}}%% syntax in code blocks

etdsoft:
The feature is implemented by splitting our old `#markup` helper in an [HTML Pipeline](https://github.com/jch/html-pipeline/) with the following filters:

1. First we convert the field syntax (e.g. `#[Title]#`) into valid Textile (i.e. 'h4.')
2. Then we process Textile
3. Then we add code-highlights to textile code blocks
4. We do auto-linking of URLs

The results:

code-highlight-01

code-highlight-02

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,12 +1,27 @@
 module ApplicationHelper # :nodoc:
   def markup(text)
-    return unless text.present?
+    # return unless text.present?
+    #
+    # output = text.dup
+    # Hash[ *text.scan(/#\[(.+?)\]#[\r|\n](.*?)(?=#\[|\z)/m).flatten.collect{ |str| str.strip } ].keys.each do |field|
+    #   output.gsub!(/#\[#{Regexp.escape(field)}\]#[\r|\n]/, "h4. #{field}\n\n")
+    # end
+    #
+    # auto_link(RedCloth.new(output, [:filter_html, :no_span_caps]).to_html, :sanitize => false ).html_safe
+    new_markup(text)
+  end
 
-    output = text.dup
-    Hash[ *text.scan(/#\[(.+?)\]#[\r|\n](.*?)(?=#\[|\z)/m).flatten.collect{ |str| str.strip } ].keys.each do |field|
-      output.gsub!(/#\[#{Regexp.escape(field)}\]#[\r|\n]/, "h4. #{field}\n\n")
-    end
+  def new_markup(text)
+    context = { }
 
-    auto_link(RedCloth.new(output, [:filter_html, :no_span_caps]).to_html, :sanitize => false ).html_safe
+    textile_pipeline = HTML::Pipeline.new [
+      HTML::Pipeline::DradisFieldableFilter,
+      HTML::Pipeline::DradisTextileFilter,
+      HTML::Pipeline::DradisCodeHighlightFilter,
+      HTML::Pipeline::AutolinkFilter
+    ], context
+
+    result = textile_pipeline.call(text)
+    result[:output].to_s.html_safe
   end
 end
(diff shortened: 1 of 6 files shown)


etdsoft:
The meat and potatoes of the PR are here:

https://github.com/dradis/dradis-ce/pull/80/files#diff-127a850b7d7b20173ad3fba88f8a25e3

We're we've split our original multi-purpose `#markup` helper into a structured pipeline with different filters.

The interesting part of this approach is that we can add more filters to the pipeline for things like @mentions, emojis, etc. (possibly for comments, and not for Issue/Note/Evidence).

rachkor:
@etdsoft I found an issue with the way that the highlighting works if the lines aren't double-spaced (which they won't be in most cases!)

The following content in Dradis: 
````
#[Fix]#
bc.. def foo

  this is a normal code line

  %%{{the first half}}%% of this line is highlighted

  the second half %%{of this line is highlighted}%%

end
````
Renders as expected: 
screen shot 2017-01-31 at 3 38 25 pm

But if you remove the new blank lines: 
````
bc.. def foo
  this is a normal code line
  %%{{the first half}}%% of this line is highlighted
  the second half %%{{of this line is highlighted}}%%
end
````
It breaks the highlighting (even if there is another un-highlighted line between them): 
screen shot 2017-01-31 at 3 42 22 pm

**Comment by etdsoft** [comment]
@rachkor I think I've addressed the issue, let me know if you can find any other quirks please

**Comment by rachkor** [comment]
@etdsoft Tested and couldn't find any other quirks 👍

File: app/helpers/application_helper.rb [hunk 1]
```diff
@@ -1,12 +1,27 @@
 module ApplicationHelper # :nodoc:
   def markup(text)
-    return unless text.present?
+    # return unless text.present?
```
dormi: why we don't have to check this anymore?

app/helpers/application_helper.rb
```diff
@@ -1,12 +1,27 @@
...
+    #
+    # output = text.dup
+    # Hash[ *text.scan(/#\[(.+?)\]#[\r|\n](.*?)(?=#\[|\z)/m).flatten.collect{ |str| str.strip } ].keys.each do |field|
+    #   output.gsub!(/#\[#{Regexp.escape(field)}\]#[\r|\n]/, "h4. #{field}\n\n")
+    # end
+    #
+    # auto_link(RedCloth.new(output, [:filter_html, :no_span_caps]).to_html, :sanitize => false ).html_safe
+    new_markup(text)
```
dormi: why cannot the changes be in the `markup` method itself?

lib/html/pipeline/dradis_code_highlight_filter.rb
```diff
@@ -0,0 +1,23 @@
+%%{{}}%%
```
dormi: what's this? should be a comment?

lib/html/pipeline/dradis_fieldable_filter.rb
```diff
@@ -0,0 +1,22 @@
...
+    # Context options:
+    #   n/a
+    #
+    # This filter does not write any additional information to the context hash.
+    class DradisFieldableFilter < TextFilter
+      # Convert the #[Field]# syntax into h4. headers
+      def call
+        output = @text.dup
+
+        Hash[ *@text.scan(/#\[(.+?)\]#[\r|\n](.*?)(?=#\[|\z)/m).flatten.collect{ |str| str.strip } ].keys.each do |field|
```
dormi: 😪

lib/html/pipeline/dradis_code_highlight_filter.rb
etdsoft: Not sure why it was there, and not sure why it didn't crash the interpreter!

PR #90, opened 2017-02-27
Fixing Gems

th3xer0:
Gemfile.plugins.template had a typo.


th3xer0:
Fixed another Gemfile issue, updating dradis-calculator_cvss from 3.0 to 3.0.1 so it works.

etdsoft:
@th3xer0 thanks for this, the typo in  	7ddd190 is spot on, however the other changes I'm not so sure about, from [the docs](http://bundler.io/v1.5/gemfile.html):

>  The specifier ~> has a special meaning, best shown by example. ~> 2.0.3 is identical to >= 2.0.3 and < 2.1. ~> 2.1 is identical to >= 2.1 and < 3.0. ~> 2.2.beta will match prerelease versions like 2.2.beta.12

If we pin the gems to `'~> 3.0.1', then we'll have to modify the Gemfile again when `3.1` is released. I'd rather leave them as `'~> 3.0'` and then `bundle update XXX` the gem to get the latest version than constraint versions too much.

Makes sense?

etdsoft:
I tried to cherry-pick this commit and do a new PR but it didn't work (it will lose your attribution). Can you please create a new PR for the Qualys change while we talk about the other changes?

This is what I did in case it helps:

    # start fresh
    $ git clone git@github.com:th3xer0/dradis-ce th3xer0-dradis
    # move all your changes to a new branch
    $ git checkout -b fix-qualys-gem
    # roll back to the first change (7ddd190)
    $ git reset --hard HEAD~5
    # push to your fork
    $ git push origin fix-qualys-gem

And open a PR from that branch. I'll merge it immediately. Thanks!

th3xer0:
Thanks, started over under th3xer0/dradis-ce gemfile-qualys-typo:
https://github.com/dradis/dradis-ce/pull/94.

PR #96, opened 2017-03-08
Use client side parsing of Differ HTML

aapomm:


app/assets/javascripts/snowcrash.js
```diff
@@ -25,6 +25,7 @@
 //= require snowcrash/modules/issues
 //= require snowcrash/modules/methodologies
 //= require snowcrash/modules/nodes
+//= require snowcrash/modules/revisions
```
etdsoft: Can you move this to `snowcrash/pages/revisions` please?

app/views/revisions/show.html.erb
```diff
@@ -36,7 +36,7 @@
     </div>
   </div>
 
-  <pre><%= raw @diffed_revision.diff.format_as(:html) %></pre>
+  <pre class='diff-body'><%= @diffed_revision.diff.format_as(:color) %></pre>
```
etdsoft: Since this is a JS-related class (i.e. no styling), can you prefix it accordingly? `js-diff-body`

PR #97, opened 2017-03-09
issues spec possible fix: clicking button with js instead of capybara

dormi:
My understanding is that when we choose the "merge into new issue" option, we alter the page in some way that prevents capybara from finding the form button (sometimes).
The thing was that `click_button` didn't raise an error, but the form was never submitted (sometimes).

Clicking the button with `excecute_script` I've not seen any error yet.


etdsoft:
not sure if adding more JS to the tests is the best way forward...

PR #101, opened 2017-03-15
Fix multiple submit

aapomm:


dormi:
I would have preferred a no js solution as @aapomm mentioned in the trello card (may be duplicating the button on each form, maybe extracting the button to a partial, not sure...)

But maybe this is not our focus now, so I give this a 👌 . What do you think?

PR #104, opened 2017-03-17
Issue table selection

aapomm:
- Added `:visible` selector when finding selected checkboxes in the issue table
- Added spec for this case and `WaitForAjax` helper.


spec/features/issue_pages/table_toolbar_spec.rb
```diff
@@ -62,5 +62,16 @@
         expect(Issue.exists?(@issue1.id)).to be false
       end
     end
+
+    context "when filtering issues" do
+      it "should not delete filtered issues" do
+        find('.js-table-filter').set('1')
+        find('#select-all').click
+        find('#delete-selected').click
+        wait_for_ajax
```
dormi: I think that we should try not to use "wait_for_ajax", while debugging other tests I read that capybara waits for the configured default time while searching for a specified change. This should work too:
`expect(page).not_to have_selector "#issue_#{@issue1.id}"`
instead of `wait_for_ajax`

aapomm: didn't know about that! ✅

PR #117, opened 2017-04-28
Fix random spec error

dormi:
We have seen lately some weird behaviour on our CI service.
After some digging I think this is due to some of the macros not being included in some type of tests (`request` and  `job` type).
In this PR I pretend:
- include those macros (also renamed them to `SessionMacros` since it makes more sense to me)
- removed `login_to_project_as_user` from CE, since we have no projects in CE
- remove references to `@project`, there are no projects in CE

### DIFF
--- a/spec/features/attachments_spec.rb
+++ b/spec/features/attachments_spec.rb
@@ -12,7 +12,7 @@ describe "Describe attachments" do
 
   describe "as authenticated user" do
     before do
-      login_to_project_as_user
+      login_as_user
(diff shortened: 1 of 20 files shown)


etdsoft:
The idea of having `login_to_project_as_user` in CE was to make sure that the specs were the same both in CE and Pro. By removing this method does it mean that we'll be using `login_as_user` in CE but then sync to Pro and change everything back to `login_to_project_as_user`?

PR #122, opened 2017-05-10
changes needed to package with traveling-ruby

dormi:
I don't want to merge all these changes.

For example I would like to skip this change: https://github.com/dradis/dradis-ce/compare/traveling-ruby-3.6.0?expand=1#diff-1d98f0039bbc6df92adee882ed99d993

(but I needed that in the package, so user is not redirected to `https`)


etdsoft:
we've shamelessly used `sed` to modify `Gemfile` and other files in the past upon packaging, we could do the same for the SSL setting in `production.rb` ?

brianp:

PR #139, opened 2017-06-30
do not pass options to other thor tasks if not required

dormi:
I noticed that when using an option in `dradis:reset`, like:
```
thor dradis:reset --no_backup
```
I saw an error message like:
```
ERROR: "thor configure" was called with arguments ["--no_backup"]
Usage: "thor dradis:setup:configure"
```
Here `dradis:configure` is one of the tasks called using `invoke` by `dradis:reset`. Looks like we cannot pass undesired arguments to a task when calling it with invoke?


etdsoft:
Do we really need to pass an empty array everywhere?

Can we just keep line 25?

dormi:
I know, looks very ugly.
But if I don't specify when invoking `dradis:setup:configure` that I don't want options, I see:
```
ERROR: "thor configure" was called with arguments ["--no-backup"]
Usage: "thor dradis:setup:configure"
```

Docs are here: http://www.rubydoc.info/github/wycats/thor/Thor%2FInvocation:invoke
They say:
> You can also supply the arguments, options and configuration values for the command to be invoked, if none is given, the same values used to initialize the invoker are used to initialize the invoked.
And there is a link to the source at the bottom.

Thoughts?

PR #148, opened 2017-08-04
Require node parent

aapomm:
### Spec
You can currently create "invisible" Nodes (that don't show up in the sidebar of the project but can be access if you browse directly to them) if you use the REST API `POST` command and set the "parent_id" to an ID not found in the project.

Steps to replicate:

* Use the POST command and make sure that the value you pass in `parent_id` is not one of the Nodes in your project.
* Confirm that the Node is created and can be accessed by navigating to `/api/nodes/#ID#` in the browser. And, confirm that the Node does not appear in the sidebar or with the other Nodes.

**Proposed solution**
Make sure that the REST API doesn't allow you to create a Node with an invalid `parent_id`. If the `parent_id` value is not a Node in the project, the Node should not be created.

### How to test
1. Use the POST command and make sure that the value you pass in `parent_id` is not one of the Nodes in your project.
2. Confirm that the Node is not created and we get a validation error message.
3. Use the POST command and make sure that the value you pass in `parent_id` is your `issue_library` node.
4. Confirm that the Node is not created and we get a validation error message.

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -51,6 +51,7 @@ class Node < ApplicationRecord
 
   # -- Validations ----------------------------------------------------------
   validates_presence_of :label
+  validates_presence_of :parent, if: Proc.new { |node| node.parent_id }
 
   # -- Scopes ---------------------------------------------------------------
   scope :in_tree, -> {
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -12,6 +12,7 @@ class Node < ApplicationRecord
     HOST = 1
     METHODOLOGY = 2
     ISSUELIB = 3
+    USER_TYPES = [DEFAULT, HOST]
   end
 
   acts_as_tree counter_cache: true, order: :label
@@ -59,7 +60,7 @@ class Node < ApplicationRecord
   }
 
   scope :user_nodes, -> {
-    where("type_id IN (?)", user_node_types)
+    where("type_id IN (?)", Types::USER_TYPES)
   }
 
 
@@ -94,10 +95,6 @@ class Node < ApplicationRecord
     find_or_create_by(label: 'Recovered', type_id: Node::Types::DEFAULT)
   end
 
-  def self.user_node_types
-    [Node::Types::DEFAULT, Node::Types::HOST]
-  end
-
   # -- Instance Methods -----------------------------------------------------
   def ancestor_of?(node)
     node && node.ancestors.include?(self)
@@ -109,7 +106,7 @@ class Node < ApplicationRecord
   end
 
   def user_node?
-    Node.user_node_types.include?(self.type_id)
+    Types::USER_TYPES.include?(self.type_id)
   end
 
   private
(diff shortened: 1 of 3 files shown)


dormi:
When creating a node with an invalid `parent_id`, now an error is returned:
```
{
  "message": "Validation error",
  "description": "Some validation errors were found and the node couldn't be saved",
  "errors": {
    "parent": [
      "can't be blank"
    ]
  }
}
``` 
I like it. But from the PR description looks like it should create the node as a root node.
Is this intended?

dormi:
Almost by accident I noticed that we can still create nodes with `parent_id` being a node of type `METHODOLOGY` or `ISSUELIB`. I think we may also need to validate that parent node has type `DEFAULT` or `HOST`.

dormi:
Ideally we could add a spec for all this.

aapomm:
@dormi I think it's better to prevent the creation of a node than creating it as a root node when an invalid `parent_id` is inputted. So that in the case that they made a mistake, no change was committed. What do you think?

spec/models/node_spec.rb
```diff
@@ -58,6 +58,27 @@
     end
   end
 
+  describe '#parent' do
+    it 'does not create with an invalid parent_id' do
+      lib_node = create(:node, type_id: 2)
```
dormi: can we use `Node::Types::METHODOLOGY` here?

spec/models/node_spec.rb
```diff
@@ -58,6 +58,27 @@
...
+      expect(node.parent).to eq(nil)
+    end
+
+    it 'creates the node as a child of the parent node' do
+      parent_node = create(:node)
+      node.parent_id = parent_node.id
+      node.save!
+      expect(node.parent).to eq(parent_node)
+    end
+  end
```
dormi: I would add a case for a non existing `parent_id`

app/models/node.rb
```diff
@@ -110,4 +111,13 @@ def destroy_attachments
     attachments_dir = Attachment.pwd.join(self.id.to_s)
     FileUtils.rm_rf attachments_dir if File.exists?(attachments_dir)
   end
+
+  def parent_node
+    if self.parent.nil? ||
+      !(self.parent.type_id == Node::Types::DEFAULT ||
+        self.parent.type_id == Node::Types::HOST)
+
+      errors.add(:base, 'Invalid parent.')
```
etdsoft: why add to :base and not to :parent_id ?

etdsoft: Since we're checking both "invalid parent" (nil) and parent assigned to a user node, should we present a different error message? I'm not sure if it's worth it.

app/models/node.rb
```diff
@@ -110,4 +111,13 @@ def destroy_attachments
     attachments_dir = Attachment.pwd.join(self.id.to_s)
     FileUtils.rm_rf attachments_dir if File.exists?(attachments_dir)
   end
+
+  def parent_node
+    if self.parent.nil? ||
+      !(self.parent.type_id == Node::Types::DEFAULT ||
```
etdsoft: This approach means we need to tweak this line if in the future we add a new Type.

The same list of types is being used by the `#user_nodes` scope. Should we extract into a helper method and reuse in both places?

PR #169, opened 2017-10-18
Multi delete issues

aapomm:
### Spec

Currently, when deleting multiple issues, multiple calls for deletion are issued at a small amount of time. This causes the mysql server to issue a deadlock error and fail to delete most of the issues, requiring a refresh from the user.

**Proposed solution**

One solution is to delete multiple issues in one delete call.

### How to test
1. Upload a file with multiple issues
2. Go to the All Issues page
3. Select all issues and delete all
4. Assert that there were no undeleted issues.

### DIFF
--- a/app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
+++ b/app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
@@ -83,13 +83,18 @@ class IssueTable
         method: 'POST'
         dataType: 'json'
         data: {ids: issue_ids}
-        success:  ->
+        success: ->
           for id in issue_ids
             $("input#issue_#{id}").closest('tr').remove()
             $("#issue_#{id}").remove()
 
           if $(@selectedIssuesSelector).length == 0
             that.resetToolbar()
+
+        error: ->
+          for id in issue_ids
+            $row = $("input#issue_#{id}").closest('tr')
+            $($row.find('td')[2]).replaceWith("<td class='text-error'>Please try again</td>")
       }
 
     # prevent Rails UJS from doing anything else.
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
```diff
@@ -73,26 +73,29 @@ class IssueTable
...
-          success: (data) ->
-            # Delete row from the table
-            $row.remove()
-            # TODO: show placeholder if no issues left
+        issue_ids.push($(this).val())
 
-            # Delete link from the sidebar
-            $("#issue_#{data.id}").remove()
+      $.ajax $('#issue-table').data('destroy-url'), {
+        method: 'POST'
```
dormi: Is it possible to use `DELETE` here?

app/controllers/issues_controller.rb
```diff
@@ -159,4 +171,16 @@ def tag_issue_from_field_content(issue)
       issue.save
     end
   end
+
+  def destroy_multiple_issues(issue_ids)
+    issue_ids.each do |id|
+      issue = Issue.find_by_id(id)
+      if issue
```
dormi: In the singular destroy action we are also checking if the issue was destroyed. May be:
```
if issue && issue.destroy
```

config/routes.rb
```diff
@@ -23,6 +23,7 @@
   resources :issues do
     collection do
       post :import
+      post :multiple_destroy
```
dormi: Same about using `DELETE`

dormi:
When deleted all issues, I see this page:
captura de pantalla 2017-10-18 a las 12 19 52

I would like to see the green box with `Issues deleted` message.

Also when deleting all available issues, I would expect to see the "empty issues page", with the `Issues deleted` green box.

Not sure if doing this without ajax could help.

dormi:
also, this should add a spec

PR #171, opened 2017-10-23
Notes tab in node view

dormi:
### Spec
Currently, there is no way of deleting multiple notes in a node in one go. 

**Proposed solution**
Adding an evidence tab in Nodes#show that has a table with checkboxes for each record should solve the problem

### How to test
1. Add a node.
2. Add multiple notes to the node (about 20)
Useful code: 
```
30.times {|i| Note.create!(text: "Note #{i}", node_id: 27, category: Category.default)}
```
(27 was my node_id, use yours)

3. Assert that the tab works as expected:
 - assert that when the mouse is over a note row, edit/delete links appear for that row. Check that those links work
  - assert that when a note is selected, the "Delete" button shows up on the toolbar above
  - assert that by clicking the select all checkbox, all notes can be selected unselected. Check also that this works clicking not only in the checkbox, but also on its surrounding grey box
  - check that notes can be filtered with the search box in the top right
  - delete a few notes (< 15) using the checkboxes and the delete button in the toolbar. The notes should be deleted inline, and a "Notes deleted" message should appera. Notes should be removed from the table and the sidebar
   - delete a few notes (> 15) usning the checkboxes and the delete button in the sidebar. A console should popup and notes should be delete in the background. When the process finished we should see a 'Worker process completed' log message, and a button to close the console must appear. When closing the button the page should reload (No deleted notes on the sidebar and on the table)


app/assets/stylesheets/snowcrash/modules/index_table.scss
```diff
@@ -0,0 +1,53 @@
+.btn-toolbar {
+  input[type='checkbox'] {
+    margin: 0;
+  }
+
+  .js-index-table-actions {
```
aapomm: let's use a non-js class for css

app/assets/stylesheets/snowcrash/modules/index_table.scss
```diff
@@ -0,0 +1,53 @@
...
+      display: none; // even when issue-actions becomes visible, keep this one hidden
+    }
+  }
+
+  #delete-selected {
+    color: #b94a48;
+  }
+}
+
+.js-items-table {
```
aapomm: same for here. let's use a non-js class

app/assets/javascripts/snowcrash/pages/nodes/notes/table.js.coffee
```diff
@@ -0,0 +1,58 @@
+document.addEventListener "turbolinks:load", ->
+  if $('body.nodes.show').length
+
+    indexTable = new IndexTable('note')
+    
+    # # Checkbox behavior: select all, show 'btn-group', etc.
```
aapomm: what's the purpose of this commented code?

spec/features/node_pages/notes_table_toolbar_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    describe 'when clicking notes' do
+      it 'displays action buttons (delete button) if 1 note is clicked' do
+        expect(find('.js-index-table-actions', visible: :all)).to_not be_visible
+        check "checkbox_note_#{@notes[0].id}"
+        expect(find('.js-index-table-actions')).to be_visible
+      end
+
+      # context 'deleting with background job' do
```
aapomm: what is this commented code for?

spec/features/node_pages/notes_table_toolbar_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      @node  = create(:node)
+      @notes = []
+      (Note::MAX_DELETED_INLINE + 1).times do |i|
+        @notes << create(:note, node: @node, text: "#[Title]#\r\ntest#{i}\r\n\r\n#[Description]#\r\nnote#{i}\r\n")
+      end
+
+      visit node_path(@node, tab: 'notes-tab')
+    end
+
+    context 'when Select All is clicked' do
```
aapomm: some of these are very similar to the specs in `spec/features/issue_pages/table_toolbar_spec.rb`. Maybe it would be better to add shared examples?

spec/features/node_pages/notes_table_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    before do
+      login_to_project_as_user
+
+      node = create(:node)
+      @note = create(:note, node: node, text: "#[Title]#\nNote1\n\n#[Description]#\nn/a\n#[Extra]#\nExtra field")
+      visit node_path(node, tab: 'notes-tab')
+    end
+
+    it 'displays column controls for Title, Created, Created by, Updated' do
```
aapomm: very similar to the `spec/features/issue_pages/table_spec.rb` specs. Maybe dry up with shared examples?

config/routes.rb
```diff
@@ -48,6 +49,14 @@
 
     resources :notes do
       resources :revisions, only: [:index, :show]
+
+      collection do
+        get 'multi-destroy-status'
```
aapomm: continue consistency here with underscore case

app/controllers/notes_controller.rb
```diff
@@ -55,10 +55,56 @@ def destroy
...
+      format.json
+    end
+  end
+
+  def multi_destroy_status
+    @logs = Log.where(
+      'uid = ? and id > ?',
+      params[:item_id].to_i, params[:after].to_i
+    )
+    @destroying = !(@logs.last.text == 'Worker process completed.') if @logs.any?
```
aapomm: `@logs.last.text != 'Worker process completed.'` is more concise

config/routes.rb
```diff
@@ -48,6 +49,14 @@
...
+      end
+
+      member do
```
aapomm: use `collection` instead? do you need the note id here?

config/routes.rb
```diff
@@ -48,6 +49,14 @@
...
+        delete 'multi-destroy'
```
aapomm: use underscore case for consistency

app/views/nodes/notes/_empty.html.erb
```diff
@@ -0,0 +1,34 @@
...
+          :title       => "Add note",
+        } %>
+    </li>
+
+    <li class="divider"></li>
+
+    <% if NoteTemplate.all.any? %>
+      <% NoteTemplate.all.each do |template| %>
+        <li>
+          <%= link_to template.name, "#{new_node_note_path(@node)}?template=#{template.id}", tabindex: '-1' %>
```
aapomm: can you use the syntax: `new_node_note_path(@node, template: template.id)` here?

app/views/nodes/notes/_empty.html.erb
```diff
@@ -0,0 +1,34 @@
...
+        :tabindex => '-1',
+        :data => {
+          :category_id => Category.default.id,
+          :title       => "Add note",
+        } %>
+    </li>
+
+    <li class="divider"></li>
+
+    <% if NoteTemplate.all.any? %>
```
aapomm: maybe query this from the controller

app/views/nodes/show.html.erb
```diff
@@ -34,7 +40,26 @@
...
-  <div class="tab-pane" id="activity-tab">
+
+  <div class="tab-pane notes <%= "active" if params[:tab] == "notes-tab" %>" id="notes-tab">
+    <div class="inner">
+      <h3>Notes</h3>
+      <div id="note-viewer">
+        <div class="inner note-text-inner">
+          <% if @sorted_notes.any? %>
+            <div id="node-notes-table" class="js-index-table" data-destroy-url="<%= multi_destroy_node_note_path(@node) %>">
+              <%= render 'nodes/notes/toolbar' %>
```
aapomm: will `render 'toolbar'` work here? (also applies to the `render`s below)

app/controllers/notes_controller.rb
dormi: 👍

app/assets/javascripts/snowcrash/pages/nodes/notes/table.js.coffee
dormi: removed

app/views/nodes/notes/_empty.html.erb
dormi: moved code to a partial, still querying like this

spec/features/node_pages/notes_table_toolbar_spec.rb
dormi: I have a problem with sqlite and the background process here, have to take a deeper look into it

app/views/nodes/show.html.erb
dormi: I can't make it work the other way...

app/assets/stylesheets/snowcrash/modules/index_table.scss
dormi: since we are already forcing ourselves to add that class in the html to make the table js work, I think that we can use the same class to make the table css work, instead of adding another `index-table-actions` class or similar 
Thoughts?

spec/features/node_pages/notes_table_spec.rb
dormi: 👍

app/assets/stylesheets/snowcrash/modules/index_table.scss
aapomm: sounds like a good idea 👍

spec/features/node_pages/notes_table_toolbar_spec.rb
dormi: 👍

config/routes.rb
```diff
@@ -48,6 +49,14 @@
 
     resources :notes do
       resources :revisions, only: [:index, :show]
+
+      collection do
+        get :multi_destroy_status
```
aapomm: you can fuse this in one `collection` block.

app/assets/stylesheets/snowcrash/modules/index_table.scss
aapomm: @dormi looking at this again, i feel like it would be better to use a class not prepended by `js-`. We should use `js-` classes for strictly js behavior only. What do you think of this convention?

app/assets/javascripts/snowcrash/pages/nodes/notes/table.js.coffee
```diff
@@ -0,0 +1,4 @@
+document.addEventListener "turbolinks:load", ->
+  if $('body.nodes.show').length
```
aapomm: @dormi I don't think this will work once we have multiple tabs in the nodes view because there will be multiple index tables rendered in the page. Do you think we should address this now or should we cross the bridge once we get there?

aapomm:
@dormi I feel like there should be some way to go to the Node/Notes index from the notes show view:

screen shot 2017-11-14 at 3 32 40 pm


Probably replace the `Nodes` breadcrumb with `Notes`? Or make the Notes header in the sidebar a link?

aapomm:
@dormi 
screen shot 2017-11-14 at 3 38 51 pm

In my opinion, the empty page looks better centered

aapomm:
@dormi maybe add a small description of what this console is for? This might look shocking and invasive for new users experiencing this.

screen shot 2017-11-14 at 3 53 52 pm

Also, how hard is it to add a cancel button? If the change is trivial, it might be good to add one.

app/assets/javascripts/snowcrash/pages/nodes/notes/table.js.coffee
dormi: let's work on this when we have the problem, probably when dealing with the evidence or attachments tab

config/routes.rb
dormi: 👍

etdsoft:
From first-time user review perspective:

- In the empty state, the Add note button displays a misaligned drop down menu
- The Delete button's text should match that of Issues#index
- Having a "Node properties" link at the top of the sidebar no longer makes sense, maybe a "Node summary" ?
- I'd rename the properties tab to just "Properties" and sort in alpha order.

app/assets/javascripts/snowcrash.js
```diff
@@ -23,6 +23,7 @@
 //= require snowcrash/modules/configurations
 //= require snowcrash/modules/console_updater
 //= require snowcrash/modules/export
+//= require snowcrash/modules/index_table
```
etdsoft: something more descriptive maybe? js table, table extender, "index" is confusing as it could mean the MVC view.

app/assets/javascripts/snowcrash.js
```diff
@@ -32,3 +33,4 @@
 //= require snowcrash/modules/uploads
 
 //= require snowcrash/pages/revisions
+//= require snowcrash/pages/nodes/notes/table
```
etdsoft: alpha order

app/assets/javascripts/snowcrash/modules/index_table.js.coffee
```diff
@@ -0,0 +1,188 @@
...
+    "project.#{@project}.#{@itemName}_columns"
+
+  refreshToolbar: =>
+    checked = $(@selectedItemsSelector).length
+    if checked
+      $('.js-index-table-actions').css('display', 'inline-block')
+    else
+      $('.js-index-table-actions').css('display', 'none')
+
+window.IndexTable = IndexTable
```
etdsoft: Isn't there a way in Coffee to achieve this without polluting `window.`?

app/assets/javascripts/snowcrash/pages/nodes/notes/table.js.coffee
```diff
@@ -0,0 +1,4 @@
...
+
+    new IndexTable('note')
```
etdsoft: Wouldn't this be problematic from a local storage point of view?

If I understand this correctly the storage key will be given by:

```
 +    "project.#{@project}.#{@itemName}_columns" 
```

and if itemName is just "note" the key will be the same for all the nodes in the project?

app/assets/stylesheets/snowcrash/modules.scss
```diff
@@ -2,11 +2,13 @@
 @import "snowcrash/modules/alerts";
 @import "snowcrash/modules/configurations";
 @import "snowcrash/modules/export";
+@import "snowcrash/modules/index_table";
```
etdsoft: same re: "index" in naming

app/assets/javascripts/snowcrash.js
dormi: I would go with `items_table`. Yes?

app/controllers/notes_controller.rb
```diff
@@ -2,7 +2,7 @@
 # resource.
 class NotesController < NestedNodeResourceController
 
-  before_action :find_or_initialize_note, except: [:index, :new]
+  before_action :find_or_initialize_note, except: [:index, :new, :multi_destroy]
```
etdsoft: different naming convention that in IssuesController?

app/controllers/notes_controller.rb
```diff
@@ -55,10 +55,56 @@ def destroy
...
   end
 
+  def multi_destroy
+    @notes = Note.where(
+      'node_id = :node AND id in (:ids)',
+      node: @node,
+      ids: params[:ids]
+    )
+
+    if @notes.any?
```
etdsoft: This triggers it's own SQL request.

Since further down (L72) you're using `#count`, why don't you capture the count outside of the `if/end` and then use `if count > 0` as a condition.

I believe it will save us one SELECT

app/controllers/notes_controller.rb
```diff
@@ -55,10 +55,56 @@ def destroy
...
+      @job_logger = Log.new(uid: (Log.maximum(:uid) || 0) + 1)
```
etdsoft: I've seen this pattern in multiple places around the code, can we encapsulate it in the initializer and make it the default behaviour if no `:uid` is passed?

app/controllers/notes_controller.rb
```diff
@@ -55,10 +55,56 @@ def destroy
...
+
+      if @notes.count > Note::MAX_DELETED_INLINE
```
etdsoft: I think this should be a Configuration value, so users can adjust it (some will have machines powerful enough to delete 50 at a time, some will stick to 15).

Also, we need to think if we want a new "MAX_DELETED_INLINE" setting for each model or a single one that we can use when deleting Issues, Notes, etc.

app/controllers/notes_controller.rb
```diff
@@ -55,10 +55,56 @@ def destroy
...
+        @job_logger.write 'Enqueueing destroy job to start in the background.'
```
etdsoft: At some point we reserved the use of "destroy" for code and methods, and agreed to use "delete" for UI / user-facing messages.

app/controllers/notes_controller.rb
```diff
@@ -55,10 +55,56 @@ def destroy
...
+        job = DestroyJob.perform_later(
```
etdsoft: Maybe adding "Multi" would help to the name.

Also, I had to scroll down and see if this was a general AR multi-destroyer, but it seems it's tied to Note. So maybe we should include that in the name of the job too?

app/controllers/notes_controller.rb
```diff
@@ -55,10 +55,56 @@ def destroy
...
+          items: @notes.to_a,
```
etdsoft: I believe that passing a class name (as string) and the list of ids instead of the full AR objects was the recommended way to handle data exchange between the app and background workers. Does it ring a bell?

app/controllers/notes_controller.rb
```diff
@@ -55,10 +55,56 @@ def destroy
...
+          author_email: current_user.email,
+          uid: @job_logger.uid
+        )
+        @job_logger.write "Job id is #{ job.job_id }."
+
+      elsif @notes.count > 0
+        @job_logger.write 'Performing destroy job inline.'
```
etdsoft: "delete"

app/jobs/destroy_job.rb
```diff
@@ -0,0 +1,24 @@
...
+  queue_as :dradis_project
+
+  def perform(items:, author_email:, uid:)
+    # FIXME: migrate logs#uid to uuid ?
+    logger = Log.new(uid: uid)
+    logger.write do
+      "Deleting #{items.count} #{items.first.class.to_s.pluralize}"
+    end
+
+    Note.transaction do
```
etdsoft: this is triggering 1 DESTROY per object, I thought we wanted to do 1 DESTROY in total?

app/assets/javascripts/snowcrash/modules/index_table.js.coffee
```diff
@@ -0,0 +1,188 @@
+class IndexTable
+  project: null
+  selectedColumns: []
+  selectedItemsSelector: ''
+  columnIndices: {}
+
+  constructor: (@itemName) ->
+    @$jsTable     = $('.js-index-table')
```
etdsoft: let's start moving away from `js-` classes and use data-behaviour for JS behaviour

https://signalvnoise.com/posts/3167-code-spelunking-in-the-all-new-basecamp

Maybe split this in a separate card?

app/views/nodes/notes/_console.html.erb
```diff
@@ -0,0 +1,21 @@
+<div id="modal-console" class="modal hide" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true">
```
etdsoft: should this be a partial that lives under `./jobs/` ?

app/views/nodes/notes/_table.html.erb
```diff
@@ -0,0 +1,47 @@
+<% cache ["node-notes-table", notes.map(&:id), notes.map(&:updated_at).map(&:to_i).sort.last] do %>
```
etdsoft: you need to account for the columns in the cache, otherwise in production changing the columns will break the UI.

This is what we have in Issues#index:

```
<% cache ['issues-table', @columns, issues.map(&:id), @issues.map(&:updated_at).map(&:to_i).sort.last] do %>
```

app/views/notes/multi_destroy.json.erb
```diff
@@ -0,0 +1,16 @@
+<% if @notes.size > Note::MAX_DELETED_INLINE %>
```
etdsoft: careful if we're changing the CONSTANT

app/views/notes/multi_destroy_status.js.erb
```diff
@@ -0,0 +1,16 @@
+<% if @logs.any? -%>
```
etdsoft: this should probably go to Jobs#status

config/routes.rb
```diff
@@ -23,6 +23,7 @@
   resources :issues do
     collection do
       post :import
+      delete :multiple_destroy
```
etdsoft: I know I've mentioned it above, but naming convention of :actions / :routes for Issues / Notes / X

app/controllers/notes_controller.rb
dormi: My intention was to this be a general job, will recheck

app/controllers/notes_controller.rb
```diff
@@ -55,10 +55,56 @@ def destroy
...
+    respond_to do |format|
+      format.html do
+        redirect_to node_path(@node, tab: 'notes-tab'),
+                    notice: 'Deleting notes...'
+      end
+      format.json
+    end
+  end
+
+  def multi_destroy_status
```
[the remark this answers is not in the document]
dormi: @etdsoft  `jobs` is already used by resque. I have used `console`, with  url `/console/status`.
Other ideas? `logs` ? or may be use `console/show` ?

app/jobs/destroy_job.rb
```diff
@@ -0,0 +1,24 @@
+class DestroyJob < ApplicationJob
+  include ActivityTracking
+
+  queue_as :dradis_project
+
+  def perform(items:, author_email:, uid:)
+    # FIXME: migrate logs#uid to uuid ?
```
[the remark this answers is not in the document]
dormi: I card this one

app/assets/javascripts/snowcrash/modules/index_table.js.coffee
dormi: Yes please

app/jobs/destroy_job.rb
dormi: I wanted to call `destroy` and not `delete` so possible callbacks are called. 
If I do: Note.destroy(array_of_ids)`, looks like it also loops through the list destroying each one one by one: http://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-delete
Then we are good as we are? (also we are already on a background job here)

app/assets/javascripts/snowcrash/modules/items_table.js.coffee
```diff
@@ -175,8 +175,10 @@ class @ItemsTable
     setTimeout(ConsoleUpdater.updateConsole, 200);
 
   storageKey: ->
-    @project ||= $('.brand').data('project') || 'ce'
-    "project.#{@project}.#{@itemName}_columns"
+    project = $('.brand').data('project') || 'ce'
+    id = $('.note-list').data('id') || ''
+    id = "#{id}." if id
+    "project.#{project}.#{id}#{@itemName}_columns"
```
etdsoft: why don't we let the HTML pass the storage-key as a data attribute instead of doing the gymnastics in the JS?

app/views/nodes/_sidebar.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for :sidebar do %>
 
-  <div id="<%= dom_id(@node) %>_link" class="note-list">
+  <div id="<%= dom_id(@node) %>_link" data-id="<%= dom_id(@node) %>" class="note-list">
```
etdsoft: `data-id` is also an unclear name, maybe `data-storage-key`?

app/models/log.rb
```diff
@@ -1,8 +1,13 @@
 class Log < ApplicationRecord
+  after_initialize :set_uid
 
-  def write(trace=nil, &block)
+  def set_uid
```
etdsoft: Instances of the old pattern:

app//controllers/export_controller.rb:36:    @log_uid = (Log.maximum(:uid) || 1) + 1
app//controllers/upload_controller.rb:20:    @last_job = Log.maximum(:uid) || 1

app/controllers/notes_controller.rb
```diff
@@ -107,7 +108,7 @@ def multiple_destroy_status
     @destroying = @logs.last.text != 'Worker process completed.' if @logs.any?
   end
 
-  private
+private
```
etdsoft: this indenting is not the one we use?

app/controllers/console_controller.rb
```diff
@@ -0,0 +1,11 @@
+# This controller is used by the console_updater js to retrieve logs
+# for a specific job
+class ConsoleController < ProjectScopedController
```
etdsoft: Can we card using Console#status for Upload and Export (report and validate) too?

Also, maybe `JobsController` would be a better name as it can allow us to grow in the future to a Jobs#index, etc.

Food for thought.

etdsoft:
@dormi  a few minor things added. Please merge when ready. 👍

app/controllers/console_controller.rb
dormi: @etdsoft  I wanted to use `jobs`, but we have a collision with `resque` then?

app/assets/javascripts/snowcrash/modules/items_table.js.coffee
dormi: Yes, good idea. Will implement that in the `evidence tab` PR, because in there the table is identified by its id, and that will help me find the unique storage key for each table

PR #187, opened 2017-11-06
Reuse table js: minor edits after using this code

dormi:
Some changes after using the table js code extracted from issues table in the new notes table: 
- renamed `issueIds` to `ids`
- I used `DELETE` instead of `POST`, I think this makes sense, but feel free to reject it
- some code used to setup a modal console when a multi-delete background task is successfully set


aapomm:
@dormi you should also update the routes with the `DELETE` method

PR #190, opened 2017-11-13
de-duplicated services table

dormi:
### Spec

Currently, the Services tables are de-duplicated based on the contents of the entire row. 

The attached screenshot is created when you upload the 3 attached files (Nmap, Qualys, Nessus). There are many duplicate entries (rows) in the table. For example, multiple entries exist for tcp/22 when there should only be one! 

Plus, the Services table exports extra columns that should not appear in the Services table by default. 

### Proposed solution
**De-duplicate**

De-duplicate the Service tables entries based on the Port + Protocol column contents. Instead of the 

**Only 7 Columns**
There should only be the following columns in both the display **and** in the exported Services table: 

* name
* port
* product
* protocol
* reason
* state
* version

Currently, these 7 columns display in Dradis but extra columns (e.g. Scripts) are added on export. 

**Extra data**
The extra columns (e.g. scripts, X Nessus, etc) should go into the services extra data section below the Services table. 

Check out the attached screenshot showing the current configuration of the NSE script output. This should become a Services Extra data section (instead of NSE script output). Each port should have a tab (as shown), then each type of data gets a new row in the table. Add a new row to that table for the source of the data like so: 
```
|_ source |_. id |_. output |
|Nmap NSE script output | ssh-hostkey | <foo> |
|Nessus | X Nessus | <bar>
```

### How to test
Upload the Nmap, Nessus, and Qualys files to a (blank) sample project

Check to confirm that the Services table is created as expected: 

* There are no duplicate entries (rows) in the Services table, just 1 row for each unique port/protocol combination
* There are only the correct 7 columns in the Services table (list above!) 
* All extra columns, e.g. `X Nessus` have been moved to the Services Extras data below the table. 

Also upload these files:
https://github.com/dradis/dradis-nmap/blob/master/spec/fixtures/files/nse-01.xml
https://github.com/dradis/dradis-metasploit/blob/master/spec/fixtures/files/msf5.xml
Check that what we have in `Services Extras` table makes sense.

### DIFF
--- a/app/models/concerns/node_properties.rb
+++ b/app/models/concerns/node_properties.rb
@@ -120,14 +120,9 @@ private
   def merge_service(new_row)
     table = self.properties[:services]
 
-    # work always with an array of hashes
-    table = [table] if table.is_a?(Hash)
-    table = [] if table.nil?
+    table = to_array_of_hashes(table)
 
-    # find new_row in table
-    position = table.find_index do |row|
-      row.values_at(:port, :protocol) == new_row.values_at(:port, :protocol)
-    end
+    position = find_position_in_table(table, new_row)
 
     if position.nil?
       # if the row was not in the table, add it
@@ -172,14 +167,9 @@ private
   def merge_service_extra(new_row)
     table = self.properties[:services_extra]
 
-    # work always with an array of hashes
-    table = [table] if table.is_a?(Hash)
-    table = [] if table.nil?
+    table = to_array_of_hashes(table)
 
-    # find new_row in table
-    position = table.find_index do |row|
-      row.values_at(:port, :protocol) == new_row.values_at(:port, :protocol)
-    end
+    position = find_position_in_table(table, new_row)
 
     if position.nil?
       # if the row was not in the table, add it
@@ -195,5 +185,17 @@ private
     end
 
     self.properties[:services_extra] = table
+  end
+
+  def to_array_of_hashes(table)
+    table = [table] if table.is_a?(Hash)
+    table = [] if table.nil?
+    table
+  end
+
+  def find_position_in_table(table, new_row)
+    table.find_index do |row|
+      row.values_at(:port, :protocol) == new_row.values_at(:port, :protocol)
+    end
   end
 end
--- a/app/models/concerns/node_properties.rb
+++ b/app/models/concerns/node_properties.rb
@@ -76,30 +76,30 @@ module NodeProperties
 
 private
 
-  # Private: Adding a :services key to the node properties is a special case:
-  #   * We check if a row exists in the services table by matching the port
-  #     and protocol columns
-  #   * We only allow a set of known columns in the :services table. The rest
-  #     of columns are moved to an extra data table called 'supplemental'. The
-  #     allowed columms are :name, :port, :product, :protocol, :reason, :state,
-  #     :version
-  # value - The new entry (Hash) we want to add to :services table
+  # Private: Adding a :services key to the node properties is a special case.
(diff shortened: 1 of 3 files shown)


spec/models/node_spec.rb
```diff
@@ -199,6 +199,84 @@
...
+
+      it 'merges \'services_extra\' by port and protocol' do
+        extra1 = [
+          { 'source' => 'nmap', 'id' => 'ssh-script', 'output' => 'extra info 1' },
+        ]
+        extra2 = [
+          { 'source' => 'nmap', 'id' => 'ssh-script', 'output' => 'extra info 2' },
+        ]
+
+        node.set_property(
```
aapomm: maybe loop this instead?

app/presenters/property_presenter.rb
```diff
@@ -40,69 +40,69 @@ def property_value
...
-        end.join.html_safe
-      end
-    end
+    output =
+      if property_key == 'services_extra'
+        render_tabs(sorted_entries)
+      else
+        thead = content_tag(:thead) do
+          content_tag(:tr) do
+            column_names.collect do |column_name|
```
aapomm: use `map` instead of `collect`

app/presenters/property_presenter.rb
```diff
@@ -40,69 +40,69 @@ def property_value
...
-    #   class: 'content-textile'
-    # )
     content_tag(:table, thead.concat(tbody), class: 'table table-condensed')
   end
 
   def render_table
     values         = property_value
     column_names   = values.map(&:keys).flatten.uniq.sort.map(&:to_sym)
-    sorted_entries = column_names.include?(:port) ? values.sort_by{ |h| h[:port] } : values
+    sorted_entries = column_names.include?(:port) \
```
aapomm: use `if else` instead of a multiline ternary operator

app/models/concerns/node_properties.rb
```diff
@@ -72,4 +73,127 @@ def raw_properties=(value)
...
+  #
+  # Returns the updated services table
+  def add_service(value)
+    # extract extra info from value
+    value_extra = value.slice(:port, :protocol, :extra)
+    value.slice!(:name, :port, :product, :protocol, :reason, :state, :version)
+
+    merge_service(value)
+
+    unless value_extra[:extra].nil?
```
aapomm: we're using double negative here. This is equivalent to `if value_extra[:extra]`, no?

app/models/concerns/node_properties.rb
```diff
@@ -72,4 +73,127 @@ def raw_properties=(value)
...
+  # is not present in the destination row, it is added. If the destination row
+  # doesn't exist at all it is created.
+  #
+  # table   - the existing table where we want to merge a new entry.
+  #         May be nil, a Hash or an Array of hashes
+  # new_row - the new entry (a Hash) we want to put in the existing table.
+  #
+  # Returns table with new_row merged
+  def merge_service_extra(new_row)
+    table = self.properties[:services_extra]
```
aapomm: a few lines are shared with `merge_service` method. I think it's possible to DRY it up

dormi: @aapomm tried to do so here: https://github.com/dradis/dradis-ce/pull/190/commits/1b028fcc49853720598f075c2e7d940fae35b797 
Suggestions?
I liked the idea to keep two separate methods, helps me understand what is going on.

aapomm: looks good 👍

georgemillo:
Closed in favour of #222

PR #192, opened 2017-11-21
Evidence tab

dormi:
### Spec
Currently, there is no way of deleting multiple evidences in a node in one go. 

**Proposed solution**
Adding an evidence tab in Nodes#show that has a table with checkboxes for each record should solve the problem

### How to test
1. Add a node.
2. Upload multiple evidences to the node.
3. Assert that the tab works as expected.

4. Delete multiple evidences in the node.
5. Assert that the evidences are deleted.


app/controllers/evidence_controller.rb
```diff
@@ -108,6 +108,41 @@ def destroy
     end
   end
 
+  def multiple_destroy
+    @evidence = Evidence.where(
```
aapomm: you can Rails-ify this query more: `@node.evidence.where(id: params[:ids])`

aapomm:
@dormi Maybe add another dedicated column for the error message
screen shot 2017-11-22 at 7 11 43 pm

app/views/nodes/show.html.erb
```diff
@@ -20,6 +22,26 @@
 </ul>
 
 <div class="tab-content">
+  <div class="tab-pane inner <%= "active" if params[:tab] == "evidence-tab" %>" id="evidence-tab">
```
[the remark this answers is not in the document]
dormi: Not sure, @etdsoft moved this to alpha order in `notes-tab` branch.
It's active by default.

dormi:
@aapomm how did you get that screenshot with columns pushed to the right when error.
I see:
captura de pantalla 2017-11-22 a las 11 45 55
(column replaced, not pushed to the right)

aapomm:
@dormi I selected 15> evidences and delete them with the `resque` service disabled.

etdsoft:
This doesn't seem to be working as expected:

nodes-show

app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
```diff
@@ -1,6 +1,6 @@
 class IssuesTable extends ItemsTable
   constructor: ->
-    super('issue')
+    super('#issues-table', 'issue')
```
etdsoft: instead of hard-coding '#issues-table' this could be a parameter to the constructor.

etdsoft: does this subclass need a special constructor or could we use some sort of "after_initialize" hook in the parent to add the tag/merge functionality?

app/controllers/evidence_controller.rb
```diff
@@ -108,6 +108,34 @@ def destroy
...
+        ids: @evidence.map(&:id),
+        klass: 'Evidence',
+        uid: @job_logger.uid
+      }
+
+      if @count > @max_deleted_inline
+        @job_logger.write 'Enqueueing multiple delete job to start in the background.'
+        job = MultiDestroyJob.perform_later(job_params)
+        @job_logger.write "Job id is #{ job.job_id }."
+
```
etdsoft: why the extra space?

app/views/console/_console.html.erb
```diff
@@ -6,16 +6,16 @@
                     style: 'display:none',
                     data: {
                       url: status_console_index_path,
-                      close_url: close_url,
+                      close_url: local_assigns.fetch(:close_url, ''),
```
etdsoft: this seems unnecessary elsewhere, looks like we're missing something in the caller to this partial.

app/views/nodes/show.html.erb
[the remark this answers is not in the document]
etdsoft: looks ok for now!

app/views/console/_console.html.erb
etdsoft: if it's because we have a single partial handling all the consoles, that's great, but then someone else (either the server via RJS or the client JS should tie each instance where the console can appear with a :close_url data attribute somehow?

app/controllers/evidence_controller.rb
etdsoft: lol, I meant the new line before `else`, but that was extra space too :-D

engines/dradis-api/spec/requests/dradis/ce/api/v1/evidence_spec.rb
```diff
@@ -82,13 +82,13 @@
         ev_1 = retrieved_evidence.find { |n| n["issue"]["id"] == @issues[1].id }
         ev_2 = retrieved_evidence.find { |n| n["issue"]["id"] == @issues[2].id }
 
-        expect(ev_0["fields"].keys).to match_array %w[Label a]
+        expect(ev_0["fields"].keys).to match_array %w[Label Title a]
```
etdsoft: since `#local_fields` is public, can we `match_array` against `ev_0.local_fields + a` to future-proof this spec?

PR #193, opened 2017-11-22
Project show turbolinks

aapomm:
### Spec

Browsing back and forth Project#show in a project with findings and methodologies would some times result in a "black" findings chart (same column count and sizes, but all filled in black, that disappears after a second.

I think it may have to do with how we draw the SVG canvas for the issue / methodology charts.

**Proposed solution**

Review the JS code involved in Projects#show both in CE and Pro to make sure we're hooking into the right Turbolinks events.

### How to test

1. Load a Welcome project and open it.
2. Navigate outside of the summary and back into it to see the "black chart" glitch.
3. Go to Methodologies add some boards / lists / cards.
4. Navigate back to Projects#show to see the weirdness of the methodology chart.


app/assets/javascripts/snowcrash/modules/issues/summary.js.erb
```diff
@@ -0,0 +1,106 @@
+document.addEventListener("turbolinks:load", function(){
```
dormi: do we need the `erb` extension ?

dormi:
@aapomm when I click the browser's back button to go back to `projects#show` the black graph appears, and it doesn't hide.
captura de pantalla 2017-11-22 a las 10 41 41

app/controllers/projects_controller.rb
```diff
@@ -26,5 +26,8 @@ def show
       @issues_by_tag[:unassigned] << issue unless assigned
     end
 
+    @count_by_tag = @issues_by_tag.map do |tag, issues|
```
etdsoft: Can't you initialize this like we do in L13 and then reuse the previous loop to update the count (at the same time you update the `@issues_by_tag` in L22 and L26?

app/assets/javascripts/snowcrash/modules/issues.js.coffee
```diff
@@ -1,8 +1,9 @@
 //= require snowcrash/modules/issues/evidence
 //= require snowcrash/modules/issues/importer
+//= require snowcrash/modules/issues/merge
 //= require snowcrash/modules/issues/table
 //= require snowcrash/modules/issues/tag-input
-//= require snowcrash/modules/issues/merge
+//= require snowcrash/modules/issues/summary
```
etdsoft: For the sake of consistency I'd make this a "page" and not a "module" and I'd put it under "pages/projects" as this is where the code is used (and not in IssuesController).

app/views/projects/_issue_summary.html.erb
```diff
@@ -40,106 +40,5 @@
...
-<% @tags.each_with_index do |tag, i| %>
-$($('.tick')[<%= i %>]).attr('fill', '<%=j tag.color %>');
-$($('.bar')[<%= i %>]).attr('fill', '<%=j tag.color %>');
-$($('.counter')[<%= i %>]).attr('fill', '<%=j tag.color %>');
-<% end %>
-$($('.tick')[<%= @tags.count %>]).attr('fill', '#ccc');
-$($('.bar')[<%= @tags.count %>]).attr('fill', '#ccc');
-$($('.counter')[<%= @tags.count %>]).attr('fill', '#ccc');
-</script>
+  <div id='issues-summary-data' data-tags='<%= @tags.map{ |t| [t.name, [t.display_name, t.color]] }.to_h.to_json %>' data-issues-count='<%= @count_by_tag.to_json %>'></div>
```
etdsoft: If we're extracting the stats calculation up to the Controller for `@count_by_tag` wouldn't it make sense to do the same for the `@tags` and extract it too?

app/assets/javascripts/snowcrash/pages/methodologies.js.coffee
```diff
@@ -1,3 +1,5 @@
+#= require snowcrash/pages/methodologies/chart
```
etdsoft: Don't we normally include all files directly from the theme manifest (snowcrash.js)?

etdsoft:
This doesn't seem to be working as expected?


projects-show

PR #205, opened 2018-01-11
Multi destroy concern

aapomm:
### Spec

We have a `multiple-destroy` action that looks alomost identical in 3 places:
- notes_controller
- evidence_controller
- issues_controller

### Proposed solution
Extract that shared code to a controller concern.

### How to test
Delete multiple notes, evidence and issues.

### DIFF
--- a/app/controllers/concerns/multiple_destroy.rb
+++ b/app/controllers/concerns/multiple_destroy.rb
@@ -11,7 +11,7 @@ module MultipleDestroy
       job_params = {
         author_email: current_user.email,
         ids: params[:ids],
-        klass: item_class_name,
+        klass: params[:controller].singularize.capitalize,
         uid: @job_logger.uid
       }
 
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/multiple_destroy.rb
```diff
@@ -0,0 +1,28 @@
...
+    # cache these values
+    @count = params[:ids].size
+    @max_deleted_inline = ::Configuration.max_deleted_inline
+
+    if @count > 0
+      @job_logger = Log.new
+      job_params = {
+        author_email: current_user.email,
+        ids: params[:ids],
+        klass: item_class_name,
```
dormi: here we could use something like: `params[:controller].singularize.capitalize`
and avoid these ìtem_class_name` methods

etdsoft:
@aapomm can you add a card to create a Routing concern as well? That way we don't have to repeat the same set of instructions in `routes.rb` for every controller that mixes in the Concern.

aapomm:
Added: https://trello.com/c/7oUEPuG9

PR #212, opened 2018-03-02
Update bundler binstub

dormi:
### Spec

We are using bundler `1.16` in our `2.9` release.
We want to have in our `bin/` folder a `bundle` binstub generated by bundler.
If not, looks like other binstubs generated by bundler are not going to start, because bundler `1.16` doesn't work with the `bundle` binstub generated by rails. 
But we want the other binstubs to be generated by rails, like when we run `rails app:update:bin`, so we don't see a warning when we start rails console or run a script.

**Proposed solution**
From what we mentioned in the specs and specially in the basecamp post, try to find a solution where we end up with rails generate binstubs so we don't see weird warning messages in rails console), except for the `bundle` binstub, where we want a bundler generated binstub.

Another solution could be to just upgrade bundler or rails, maybe this is fixed.

### How to test

- install a binstub generated with bundler:
`bundle binstub unicorn` -> creates `bin/unicorn`

- assert we can execute that binstub:
`bundle exec bin/unicorn`
and that we don't see an error like:

```
Your `bin/bundle` was not generated by Bundler, so this binstub cannot run.
Replace `bin/bundle` by running `bundle binstubs bundler --force`, then run this command again.
````

- assert that we can still execute binstubs generated by rails, with no warning/error:
`bundle exec rails c`
we should not see:

```
Looks like your app's ./bin/rails is a stub that was generated by Bundler.

In Rails 5, your app's bin/ directory contains executables that are versioned
like any other source code, rather than stubs that are generated on demand.

Here's how to upgrade:

  bundle config --delete bin    # Turn off Bundler's stub generator
  rails app:update:bin          # Use the new Rails 5 executables
  git add bin                   # Add bin/ to source control

You may need to remove bin/ from your .gitignore as well.

When you install a gem whose executable you want to use in your app,
generate it and add it to source control:

  bundle binstubs some-gem-name
  git add bin/new-executable

Loading development environment (Rails 5.1.3)
```

### CODE REVIEW

File: bin/setup [hunk 1]
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env ruby
-require 'fileutils'
```
aapomm: why was the alpha order changed?

bin/setup
```diff
@@ -10,33 +10,25 @@ def system!(*args)
   system(*args) || abort("\n== Command #{args} failed ==")
 end
 
-Dir.chdir APP_ROOT do
+chdir APP_ROOT do
   # This script is a starting point to setup your application.
   # Add necessary setup steps to this file.
 
-  puts "\n== Enabling default add-ons =="
```
aapomm: why are some of these removed from the setup binary?

bin/setup
dormi: In this PR `bin/bundler` is generated by bundler
But also the rest of binstubs that rails generates have been updated: `bin/rails app:update:bin`
Looks like `bin/setup` has changed a bit.
I's an autogenerated file

bin/setup
dormi: same response as above

bin/setup
aapomm: maybe add it back? don't we need those for setting up the development environment?

georgemillo:
Looks good to me 👍

bin/setup
dormi: @aapomm you are right!
Restored copying  `Gemfile.plugins`, `secrets.yml`, `database.yml`

PR #213, opened 2018-03-06
Headless chrome

dormi:
### Spec

It is recommended to move from phantomjs to headless chrome, now that chrome supports a headless version.

**Proposed solution**
Install headless chrome and switch gems + config, tune up specs so they keep green.

### How to test

- install headless chrome. On Macos:
`brew install chromedriver`

- install required gems:
`bundle install`

- asserts specs pass:
`bin/rspec spec engines/dradis-api/spec`


Gemfile.lock
```diff
@@ -1,3 +1,111 @@
+GIT
```
aapomm: @dormi I don't think we need these changes in the Gemfile

PR #222, opened 2018-03-22
better handling of node 'service' properties

georgemillo:


app/presenters/property_presenter.rb
```diff
@@ -40,39 +46,14 @@ def property_value
...
-    column_names   = values.map(&:keys).flatten.uniq.sort.map(&:to_sym)
-    sorted_entries = column_names.include?(:port) ? values.sort_by{ |h| h[:port] } : values
-
-    output = if column_names.delete(:scripts)
-               render_tabs(sorted_entries)
-             else
-               ''
-             end
+    column_names   = values.map(&:keys).flatten.uniq.sort.map(&:to_sym).
+                     delete_if{ |cn| !services_table_columns.include?(cn) }
```
georgemillo: this line is from @dormi's original commits. But isn't `render_table` going to be used for other tables, not just for the Services table? It's called whenever `property_value` is an array of hashes - but with this line (unless I'm misunderstanding something), the table is going to be empty whenever none of the columns are in `services_table_columns` (which is likely going to be any table other than services).

Do we need to handle the case where the property is an array of hashes, but isn't `:services` or `:services_extras`? If so, shouldn't this line be removed? If not, `render_table` can be removed entirely.

app/presenters/property_presenter.rb
```diff
@@ -40,39 +46,14 @@ def property_value
...
+    table          = table_info(
+      values: values,
+      sort_by_port: column_names.include?(:port)
```
georgemillo: Likewise, this line can be removed if we're not dealing with the 'services' table.

app/models/concerns/node_properties.rb
```diff
@@ -18,10 +18,17 @@ def self.included(base)
     end
   end
 
+  SERVICE_KEYS = %i[protocol port state product reason name version]
+
   # -------------------------------------------- Individual property management
   # Sets a property, storing value as Array when needed
   # and taking care of duplications
   def set_property(key, value)
+    if [:services, :services_extras].include?(key) # let's get defensive
```
etdsoft: nice touch!

app/models/concerns/node_properties.rb
```diff
@@ -49,6 +56,50 @@ def has_any_property?
...
+      self.properties[:services].push(core)
+    else
+      # the variable 'this_service' is a reference, so updating it like this
+      # will update the entry in original properties[:services] array
+      this_service.merge!(core)
+    end
+
+    if extra.any?
+      self.properties[:services_extras] ||= {}
+      self.properties[:services_extras]["#{protocol}/#{port}"] ||= []
```
etdsoft: is there any value in extracting `"#{protocol}/#{port}"` into a variable?

etdsoft:
@rachkor can you give this a go from a user's perspective?

app/presenters/property_presenter.rb
etdsoft: > But isn't render_table going to be used for other tables, not just for the Services table?

That's the idea, yes.

Right now, this rendering code is failing for a property like:

```
...
"foo": [
{ "port": "80", "state": "open" },
{ "port": "443", "state": "open" },
{ "port": "25", "state": "open" }
],
...
```

When a simple table should be shown.

app/presenters/property_presenter.rb
```diff
@@ -98,38 +79,45 @@ def render_table
...
+  # [ {"port" => 80, "name" => "foo"},
+  #   {"port" => 22, "name" => "bar"},
+  #   {"port" => 80, "name" => "baz"}, ... ]
+  # into
+  # [ {"port" => 22, "name" => "baz"},
+  #   {"port" => 80, "name" => "foo<br></br>bar"}, ... ]
+  # Id sort_by_port is true it merges duplicated rows by [port/protocol],
+  # adding duplicted values with <br/><br/>
+  # It always filters the keys to those available in `services_table_columns`
+  def table_info(values:, sort_by_port:)
```
etdsoft: I don't think it's this layer's Is it the view layer's responsibility to do this de-dup (e.g. to know what properties should be considered equivalent and hence merged?)

I'd argue is the `NodeProperties` concern responsibility, maybe via a special accessor.

Otherwise each view layer needs to replicate the de-dup code (I'm thinking Word rendering). I'm aware that each layer needs to have some code to go from properties to the table representation of the view format (HTML vs Word), but that's view-specific details. De-dupping is a higher level concept.

georgemillo:
I think this is good to go now.

georgemillo:
@etdsoft @rachkor  I just made the changes discussed on Slack, please have another look :)

PR #226, opened 2018-03-29
move items_table.js to shared

georgemillo:


etdsoft:
thanks @georgemillo. I know where this PR is coming from, but for CE we try to include a bit of a description of what the PR is about for git log history purposes

PR #229, opened 2018-04-03
Add migration to increase node properties db limit

aapomm:
### Spec

When uploading from a plugin (ex: Nessus), it's possible for the process to encounter the error: `ActiveRecord::ValueTooLong: Mysql2::Error: Data too long for column 'properties'`. This is usually caused by a node with a significant amount of open ports, which is then stored in the node's properties attribute.

**Proposed solution**

Add a migration to increase the db limit of the node's properties attribute to accommodate all the data.

### How to test

1. Upload a large nessus file through the command line.
2. Assert that there was no error with the upload.

### DIFF
--- a/db/migrate/20180403094945_increase_node_properties_limit.rb
+++ b/db/migrate/20180403094945_increase_node_properties_limit.rb
@@ -1,0 +1,5 @@
+class IncreaseNodePropertiesLimit < ActiveRecord::Migration[5.1]
+  def change
+    change_column :nodes, :properties, :text, limit: 4.gigabytes - 1
+  end
+end
--- a/db/migrate/20180403094945_increase_node_properties_limit.rb
+++ b/db/migrate/20180403094945_increase_node_properties_limit.rb
@@ -1,5 +1,9 @@
 class IncreaseNodePropertiesLimit < ActiveRecord::Migration[5.1]
-  def change
-    change_column :nodes, :properties, :text, limit: 4.gigabytes - 1
+  def up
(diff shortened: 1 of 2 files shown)


db/migrate/20180403094945_increase_node_properties_limit.rb
```diff
@@ -0,0 +1,5 @@
+class IncreaseNodePropertiesLimit < ActiveRecord::Migration[5.1]
+  def change
+    change_column :nodes, :properties, :text, limit: 4.gigabytes - 1
```
etdsoft: this number is absurd, wouldn't it be a waste of resources?

aapomm: I'm assuming it's just the limit and in an ideal scenario this amount is never reached. This limit is also the same for the `versions` table `object` column.

dormi: so we are just using that big number because this forces rails to use a `longtext` column in the database?

We are using a different number in the versions table?  https://github.com/dradis/dradis-ce/blob/master/db/migrate/20151119124120_create_versions.rb#L7

I have also seen someone doing it like this: https://gist.github.com/ozgun/7672578
Does this work?

aapomm: @dormi In using the mysql adapter, the value in the versions migration forces the column to be of `longtext` type (see the schema after migration).

Updated the PR with the `:longtext` option.

brianp:

PR #242, opened 2018-05-16
Make User a true AR model

etdsoft:
If we're going to start creating more complex relationships w/ the User model, we might as well have it become truly DB-backed.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,43 +1,9 @@
 class User < ApplicationRecord
-  # -- Table-less ActiveRecord hackery ----------------------------------------
-  def self.columns
-    @columns ||= [];
-  end
-
-  def self.column(name, default = nil, sql_type_metadata = nil, null = true)
-    # Warning: this is relying undocumented Rails behaviour that may change
-    # without notice. (Last tested on Rails 5.0.0.1)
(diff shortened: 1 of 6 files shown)


aapomm:
@etdsoft I'm seeing more `User.new` instances that needs to be updated:
- [app/jobs/multi_destroy_job.rb#L21](https://github.com/dradis/dradis-ce/blob/master/app/jobs/multi_destroy_job.rb#L21)
- [dradis-api engines/dradis-api/lib/dradis/ce/api/warden_strategy.rb#L24](https://github.com/dradis/dradis-ce/blob/master/engines/dradis-api/lib/dradis/ce/api/warden_strategy.rb#L24)
- [engines/dradis-api/spec/support/shared_context.rb#L17](https://github.com/dradis/dradis-ce/blob/master/engines/dradis-api/spec/support/shared_context.rb#L17)

app/models/user.rb
```diff
@@ -1,43 +1,9 @@
...
-  def id; 1; end
-  def new_record?; false; end
-  def persisted?; true; end
-
   # -- Relationships --------------------------------------------------------
   has_many :activities
 
   # -- Callbacks ------------------------------------------------------------
   # -- Validations ----------------------------------------------------------
-  validates :email,
```
dormi: why do we stop validating the `email` ?

Issue #228, opened 2018-04-02
Long node names are cut


aapomm:
### Steps to reproduce
Create a node with a very long label.

### Expected behavior
The node name breaks to a newline.

### Actual behavior
The node name does not break and gets cut.

![nodes](https://user-images.githubusercontent.com/1908272/38188248-b77a45f8-368d-11e8-8a15-f015bdc27383.png)

### System configuration
**Dradis version**: CE 3.9 / PRO 2.9

**Ruby version**: 2.4

**OS version**: Any

MattBudz:
Fixed in v3.16

PR #244, opened 2018-05-17
fix long nodenames are cut

haseebeqx:
### Summary

This is a  fix for issue #228 Long Names are Cut when using large node names.
after this fix long node names will break to newlines making it visible to end user.   

### Copyright assignment

### DIFF
--- a/app/assets/stylesheets/snowcrash/base.scss
+++ b/app/assets/stylesheets/snowcrash/base.scss
@@ -292,7 +292,6 @@ body {
           &.node {
             a {
               word-break: break-all;
-              margin-left: 30px;
             }
           }
         }
--- a/app/assets/stylesheets/snowcrash/base.scss
+++ b/app/assets/stylesheets/snowcrash/base.scss
@@ -171,6 +171,8 @@ body {
             display: inline-block;
             width: auto;
             color: lighten($primaryBgColor, 40%);
+            word-break: break-all;
+            line-height: 1.5em;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/snowcrash/base.scss
```diff
@@ -289,6 +289,11 @@ body {
             font-size: 10px;
             padding: 0 0 1em 3em;
           }
+          &.node {
```
dormi: Two comments:

- Maybe there is no need to apply this change to the `node` class only, it can be applied above, like here: https://github.com/dradis/dradis-ce/blob/master/app/assets/stylesheets/snowcrash/base.scss#L170

- we may use `break-word` instead of `break-all`, so if the name have a white space it will break there. 

Thank you for your contribution!

haseebeqx: hi @dormi in the original issue (#228) it is a word without space . so if we give `break-word` it will not serve that purpose. node names can actually be without space.

haseebeqx:
breaking a word creates a very big gap, so i added `line-height` also

Without `line-height`

screenshot-2018-5-24 project summary


with `line-hieght : 1.5em`
screenshot-2018-5-24 project summary 2

haseebeqx:
@dormi moved code to said area and changed line height

haseebeqx:
hello, can i close this?

PR #250, opened 2018-05-30
Ignore vagrant local files.

rimenes:
The files placed at the .vagrant directory are relative to the host OS
and should not be committed to the git repository.


etdsoft:
Thanks @rimenes

PR #253, opened 2018-06-05
Comments & Notifications scaffold

dormi:
### Spec
We have been discussing on how to add nomments and notifications

**Proposed solution**

Propose a base set of db model classes that allow us to start working on this.

### How to test
Specs should pass.


.rubocop.yml
```diff
@@ -83,7 +83,7 @@ Style/MethodDefParentheses:
   Enabled: true
 
 Style/FrozenStringLiteralComment:
-  Enabled: true
+  Enabled: false
```
etdsoft: why have you removed this?

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,10 @@
+class CommentsController < ProjectScopedController
```
etdsoft: Let's not keep digging this hole.

include a new `ProjectScoped` controller concern (which would be empty at this stage) instead.

app/controllers/subscriptions_controller.rb
```diff
@@ -0,0 +1,7 @@
+class SubscriptionsController < ProjectScopedController
```
etdsoft: ditto

etdsoft:
Can we please split this PR into different sub-components? Comments first, etc. Otherwise we're opening 15 different work streams here and closing none.

PR #257, opened 2018-06-12
Implement 'Send to...' menu for issues

etdsoft:
There are going to be a number of plugins that will let you push your Issues to other components like JIRA, VSTS, ServiceNow, etc. This PR creates a common infrastructure where this type of :sync plugins can hook into.


dormi:
@etdsoft what's the deal with `class="action-link"`? What is it for?

etdsoft:
It's just a CSS class to be applied to the action links next to the main title of the page. Before we applied some styles to all `a` children, but now with the `Send to...` menu, the items of the menu are also `a` elements, so we'd be braking the menu look & feel.

PR #258, opened 2018-06-12
Comments list

dormi:
### Spec

The first thing we need is a place to be able to read the comments related to the commentable item

When I see an item that allows comments, I want to see a comments tab, so I can check comments

**Proposed solution**
Add a tab to a the corresponding item.
The tab should be called "comments" and have a list of the related comments to that item.

The comments create, update, delete actions are not part of this task.
Adding a comments count is not part of this task.

We should use a partial to share the view among all items that may have comments.

### How to test
* Use the rails console to add a fe comments to an issue. Maybe like this:    
  `Comment.create(commentable: Issue.first, user: User.first, content: "hello")`
* Visit the item, assert the comments tab looks nice.

### DIFF
--- a/app/presenters/comment_presenter.rb
+++ b/app/presenters/comment_presenter.rb
@@ -1,0 +1,45 @@
+class CommentPresenter < BasePresenter
+  presents :comment
+
+  def avatar_with_link(size)
+    h.link_to(avatar_image(size), 'javascript:void(0)')
+  end
+
+  def created_at_ago
+    h.local_time_ago(comment.created_at)
+  end
+
+  # Interestingly enough we're not linking the email to anything yet as we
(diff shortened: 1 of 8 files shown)


dormi:
@aapomm about:

> CommentPresenter is very similar to ActivityPresenter that we can DRY the code with a parent class (we can just inherit from ActivityPresenter, too). The NotificationPresenter needs this too.

I wanted to do that. I was thinking of an `AvatarPresenter`. But I've noticed that `activity.user` (string) is not the same as `comment.user` (already a user object). This makes most of the shared methods be slightly different.
Maybe before doing this we should first work on a new PR to make activity use a user object.

dormi:
captura de pantalla 2018-06-13 a las 10 26 59

PR #260, opened 2018-06-13
Subscriptions table

dormi:
### Spec
We are going to have notifications for users. But before creating notifications, we must know if it must be created (i.e if the user is interested in that particular model).

**Proposed solution**

Propose a table structure + db model that allows us to store subscriptions in db.
The proposed table must be able to store a subscription with info that tells us:

- the model 
- the user

Anything else?

### How to test
Specs should pass.


etdsoft:
Looking good @dormi, resolve the merge conflicts and merge please.

PR #261, opened 2018-06-14
Comments list

dormi:
### Spec

The first thing we need is a place to be able to read the comments related to the commentable item

When I see an item that allows comments, I want to see a comments tab, so I can check comments

**Proposed solution**
Add comments to the bottom of the main tab.

The comments create, update, delete actions are not part of this task.
Adding a comments count is not part of this task.

We should use a partial to share the view among all items that may have comments.

concept art

### How to test
* Use the rails console to add a fe comments to an issue. Maybe like this:    
  `Comment.create(commentable: Issue.first, user: User.first, content: "hello")`
* Visit the item, assert the comments looks nice.


etdsoft:
We should probably eager load the Comments in Issues#show action in the controller via `#includes(:comments)`.

app/views/issues/_comments.html.erb
```diff
@@ -0,0 +1,2 @@
+<h3><a id="comments"></a>Comments <span class="badge"><%= @issue.comments.count %></span></h3>
```
etdsoft: you can attach the `:id` to the `<h3>` tag and skip the `<a>` altogether.

PR #262, opened 2018-06-14
General presenter

aapomm:
### Spec
Currently, we have an ActivityPresenter that we use when we display activities. With the comments and the notifications coming into play, we can use some of the methods from the ActivityPresenter to display these items such as the avatar, partial, etc. to avoid reinventing the wheel.

#### Proposed Solution
We need to create a general/higher Presenter that the other presenters can inherit from with the reusable methods.

Additionally, to make the general Presenter code be more general we need to associate Activities with users (now that we have the proper User DB) instead of having a user email attribute.

### How to test
This is a refactor PR, but we need to make sure that the Recent Activities view is unchanged.


app/presenters/activity_presenter.rb
```diff
@@ -1,30 +1,6 @@
...
-                when 'Node'
-                  'fa-folder-o'
-                when 'Note'
-                  'fa-file-text-o'
-                else
-                  ''
-                end
-    h.content_tag :span, nil, class: icon_css
-  end
+  collection :trackable
```
dormi: `collection` may sound like a `has many` relationship?
Is there a particular reason to choose that name ?

aapomm: I just thought that `collection` best defines the polymorphic collection type name for the model of the presenter. Ideas?

PR #264, opened 2018-06-20
Add comment charset migration and database config updates

aapomm:
### Spec
Currently, emojis cannot be saved in the database.

#### Proposed Solution
Update the character set for comments to allow emojis in the database.

### How to test
1. Run rails db:migrate
2. Create a comment with content with your favorite emoji.
3. Assert that it is saved.

### DIFF
--- a/db/migrate/20180620124108_use_utf8_mb4_for_comments.rb
+++ b/db/migrate/20180620124108_use_utf8_mb4_for_comments.rb
@@ -1,0 +1,11 @@
+class UseUtf8Mb4ForComments < ActiveRecord::Migration[5.1]
+  def up
+    execute "ALTER TABLE comments CHARSET utf8mb4 COLLATE utf8mb4_unicode_ci"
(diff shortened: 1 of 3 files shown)


db/migrate/20180620124108_use_utf8_mb4_for_comments.rb
```diff
@@ -0,0 +1,11 @@
+class UseUtf8Mb4ForComments < ActiveRecord::Migration[5.1]
```
dormi: By default we are using sqlite.
But this migration does not work with sqlite:
`SQLite3::SQLException: near "CHARSET": syntax error: ALTER TABLE comments CHARSET utf8mb4 COLLATE utf8mb4_unicode_ci`
Also, in sqlite emojis are aready working.
Does it make sense to make this a mysql only migration?
Maybe with https://stackoverflow.com/questions/1628608/how-do-i-check-the-database-type-in-a-rails-migration

dormi: What problems may we face if we try to make this change for all tables:
https://gist.github.com/amuntasim/f3b12f20a30e9a9f3fb0

PR #265, opened 2018-06-21
Create subscriptions

dormi:
### Spec
The next step after adding a comment, is knowing who must be notified about it.
To know that we look at the subscriptions table.
If we comment on a commentable, we should be included in that table, so next time somebody posts a comment on that commentable, we may get notified.

Also, when a subscribable that has an author (like an issue) is created, it makes sense to create a subscription of that author to the created subscribable (issue).

Since both subscription creation entries may share code, let's try to implement both cases in this same PR. 

**Proposed solution**
Using callbacks add a subscription for the comment author when a comment is created.
Uing callbacks, add a subscription for the subscribable author when the subscribable is created.
The subscription must be unique for a user and a subcribable.

### How to test
* Recreate the database, changes have been made in the migrations: `rails db:drop + rails db:create + rails db:migrate`
 * Create a user and an issue.
* Assert that a subscription has been created for that issue and that user.
* With the console, comment on that Issue with the user that created it: 
`Comment.create(commentable: Issue.last, user: User.last, content: "hello")`
* Assert a comment is created, but not a new subscription (use count, the insert into subscriptions query may be seen in the logs, but should be stopped by the db unique index)
* Create another user
* With the console, comment on that Issue with the user that did not create it (same command above)
* Assert that a comment and a subscription are created
* Comment again with the same params (command above again).
* Assert that a comment is created but no subscription (use count, the insert into subscriptions query may be seen in the logs, but should be stopped by the db unique index)

### DIFF
--- a/app/models/concerns/subscribable.rb
+++ b/app/models/concerns/subscribable.rb
@@ -1,0 +1,7 @@
+module Subscribable
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :subscriptions, as: :subscribable, dependent: :destroy
+  end
+end
--- a/app/models/concerns/subscribable.rb
+++ b/app/models/concerns/subscribable.rb
@@ -3,5 +3,13 @@ module Subscribable
 
   included do
     has_many :subscriptions, as: :subscribable, dependent: :destroy
+
+    after_create :create_subscription
+  end
+
+  def create_subscription
+    if respond_to?(:author) && user = User.find_by_email(author)
+      Subscription.subscribe(user: user, to: self)
+    end
   end
 end
--- a/db/migrate/20180613151829_create_subscriptions.rb
+++ b/db/migrate/20180613151829_create_subscriptions.rb
@@ -1,0 +1,8 @@
+class CreateSubscriptions < ActiveRecord::Migration[5.1]
+  def change
+    create_table :subscriptions do |t|
+      t.references :subscribable, polymorphic: true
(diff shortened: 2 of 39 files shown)


app/models/concerns/subscribable.rb
```diff
@@ -3,5 +3,13 @@ module Subscribable
 
   included do
     has_many :subscriptions, as: :subscribable, dependent: :destroy
+
+    after_create :create_subscription
```
etdsoft: Why aren't we using this in `Comment`?

db/migrate/20180613151829_create_subscriptions.rb
```diff
@@ -4,5 +4,10 @@ def change
       t.references :subscribable, polymorphic: true
       t.references :user, index: true
     end
+
+    add_index :subscriptions,
+              [:subscribable_id, :subscribable_type, :user_id],
+              unique: true,
+              name: 'uniqueness_index'
```
etdsoft: do we need to reference the table name in the index name?

app/models/concerns/subscribable.rb
dormi: Nor sure what you mean.
Yes we are? (https://github.com/dradis/dradis-ce/pull/265/files#diff-3b4cc13a815e6a32db590e41f5d00d49R7)
But `Comment` is not a `Subcribable`, so it does it itself, not with this concern.

PR #268, opened 2018-06-28
Create a project_path() route so we can nest other top-level resources

etdsoft:
Initial changes to allow nesting of resources inside projects to allow for one-project-per-tab URLs


etdsoft:
closed in favour of https://github.com/dradis/dradis-ce/pull/272

PR #270, opened 2018-06-29
Issues#show - always show Send to, make into a trigger link

etdsoft:
This PR changes the `Send to...` menu so it's always shown. A trigger link for the built-in IssueLibrary is added if the add-on isn't loaded yet.

### DIFF
--- a/app/views/issues/_send_to_menu.html.erb
+++ b/app/views/issues/_send_to_menu.html.erb
@@ -1,16 +1,22 @@
-<% sync_plugins = Dradis::Plugins::with_feature(:issue_sync) %>
-<% if sync_plugins.any? %>
 <span class="dropdown">
   <a class="dropdown-toggle action-link" data-name="issues" data-toggle="dropdown" href="javascript:void(0)">
     Send to...
   </a>
   <ul class="dropdown-menu" role="menu">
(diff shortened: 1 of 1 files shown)


app/views/issues/_send_to_menu.html.erb
```diff
@@ -1,16 +1,22 @@
...
+  <% sync_plugins = Dradis::Plugins::with_feature(:issue_sync) %>
+  <% if sync_plugins.any? %>
+    <% sync_plugins.each do |plugin| %>
+      <%=
+        plugin_path = ActiveSupport::Inflector.underscore(ActiveSupport::Inflector.deconstantize(plugin.name))
+        render partial: "#{plugin_path}/issues/send_to_menu"
+      %>
+    <% end %>
+  <% end %>
+  <% unless defined?(Dradis::Pro) && sync_plugins.include?(Dradis::Pro::Issuelib::Engine) %>
```
aapomm: @etdsoft This returns an error `NameError: uninitialized constant Dradis::Pro::Issuelib` if the IssueLibrary is not installed. It might be good to stringify the`sync_plugins` list and use the string `"Dradis::Pro::Issuelib::Engine"` to check for inclusion instead.

PR #271, opened 2018-06-29
add activity tracking to comments

dormi:
### Spec
Add activity tracking to comments, as we have for other resources

**Proposed solution**
Try to use different approach: use callbacks in the model that trigger a bg task to create the corresponding activity.

### How to test
Use the rails console to add a comment to an issue. Maybe like this:    
Comment.create(commentable: Issue.first, user: User.first, content: "hello")
The main page, assert we can see the activity.
Edit + delete the issue, check those activities too


app/jobs/activity_tracking_job.rb
```diff
@@ -0,0 +1,9 @@
+class ActivityTrackingJob < ApplicationJob
+  include ActivityTracking
```
etdsoft: is this the controller concern? It's not immediately clear here.

I'd duplicate the code here and eventually remove the controller concern.

Something we need to think about is how to "silence" tracking, otherwise we're going to end up with a bunch of activities when we don't mean to (rails console, specs, etc.).

dormi:
> Something we need to think about is how to "silence" tracking, otherwise we're going to end up with a bunch of activities when we don't mean to (rails console, specs, etc.).

One possible solution:

dormi:
Closing, we decided not to use callbacks here.

PR #274, opened 2018-06-29
Add comments to issues

dormi:
### Spec

We need is a place to be able to read the comments related to the commentable item, post those comments, edit them, and delete them.

When I see an item that allows comments, I want to see a comments list, so I can manage comments.

**Proposed solution**

The list must contain all related comments to that item, and links to edit and delete them.

Only the user that creates a comment must be able to edit and delete it.

### How to test

- Visit an issue, create some comments
- Assert that the comments can be updated and deleted.
- Assert that a user can only edit/delete her own comments
- Assert that when we create, edit, delete comments, an activity is created


app/controllers/comments_controller.rb
```diff
@@ -1,4 +1,42 @@
 class CommentsController < AuthenticatedController
+  include ActionView::RecordIdentifier
   include ProjectScoped
-end
 
+  before_action :find_or_initialize_comment
```
etdsoft: this isn't the standard `before_action` name we're now using. It also isn't the standard implementation for `set_comment` we're using elsewhere.

app/controllers/comments_controller.rb
```diff
@@ -1,4 +1,42 @@
...
+  before_action :check_comment_author, only: [:update, :destroy]
```
etdsoft: Instead of checking this and returning a blank screen (what `head` does in a browser), we've pre-empted the problem by assigning the model's user after the fact:

https://github.com/secroots/dradis-pro/blob/develop/app/controllers/notes_controller.rb#L18

Also, :user_id isn't included in `comment_params` which I believe means negates this problem?

I'd keep `:user_id` excluded, but also explicitly add a `comment.user = current_user` in #create

app/controllers/comments_controller.rb
```diff
@@ -1,4 +1,42 @@
...
+  private
+  def check_comment_author
+    head :forbidden unless @comment.user == current_user
+  end
+
+  def comment_params
+    params.require(:comment).permit(:content, :commentable_type, :commentable_id)
+  end
+
+  def find_or_initialize_comment
```
etdsoft: I'd migrate this to a `set_comment` and move the `#new` init to the action.

app/views/issues/_comments.html.erb
```diff
@@ -0,0 +1,2 @@
+<h3>Comments <span class="badge"><%= @issue.comments.count %></span></h3>
```
etdsoft: This whole partial seems unnecessary, I'd move the two lines to Issues#show or if you really want to extract it, I'd move it under `./app/views/comments` merging the <h3> into the `comments/feed` partial. That way you can re-use it for Evidence, Notes, etc.

config/routes.rb
```diff
@@ -35,6 +35,7 @@
     end
   end
 
+  resources :comments
```
etdsoft: duplicates L20

app/controllers/comments_controller.rb
dormi: @etdsoft this check is for the `:destroy` and `:update actions` only.
I wanted to prevent `current_user` from editing or deleting comments  if she/he is not the author of that comment.

etdsoft: then I'd rename to `:ensure_comment_author` and handle it properly instead of bailing with a `head`.

PR #269, opened 2018-06-29
using a dropdown menu in issue header

dormi:
Finding a way to make it look less cluttered, since we may need to add more actions here: send_to, unsubscribe, link_to comments...

menu

PR #280, opened 2018-07-04
Create subscriptions 2

dormi:
**NOTE:** rebuilding https://github.com/dradis/dradis-ce/pull/265, the base branch changed and it was adding unnecessary stuff.

---

### Spec
The next step after adding a comment, is knowing who must be notified about it.
To know that we look at the subscriptions table.
If we comment on a commentable, we should be included in that table, so next time somebody posts a comment on that commentable, we may get notified.

Also, when a subscribable that has an author (like an issue) is created, it makes sense to create a subscription of that author to the created subscribable (issue).

Since both subscription creation entries may share code, let's try to implement both cases in this same PR. 

**Proposed solution**
Using callbacks add a subscription for the comment author when a comment is created.
Uing callbacks, add a subscription for the subscribable author when the subscribable is created.
The subscription must be unique for a user and a subcribable.

### How to test
* Recreate the database, changes have been made in the migrations: `rails db:drop + rails db:create + rails db:migrate`
 * Create a user and an issue.
* Assert that a subscription has been created for that issue and that user.
*  Comment on that Issue with the user that created it
* Assert a comment is created, but not a new subscription (use count, the insert into subscriptions query may be seen in the logs, but should be stopped by the db unique index)
* Create another user
* With the console, comment on that Issue with the user that did not create it
* Assert that a comment and a subscription are created
* Comment again
* Assert that a comment is created but no subscription (use count, the insert into subscriptions query may be seen in the logs, but should be stopped by the db unique index)


dormi:
subscribed
unsubscribed

Note that we have this other "actions menu" PR to may be helpful here:
https://github.com/dradis/dradis-ce/pull/269

etdsoft:
@dormi I changed the index name to somewhat follow the standard convention, other than that, this seems ready, 👍

PR #281, opened 2018-07-04
Subscribe/Unscubscribe links

dormi:
### Spec
The user should be able to manually subscribe or unscubscribe to/from a subscribable item. 

**Proposed solution**
Add links to the subscription resource from a subscribable show page, so the user can edit her subscription for that item.

### How to test

- Visit an issue   
- Assert there is a link to subscribe to that issue.
- Subscribe and check subscriptions table on the db for a new record
- Unsubscribe and check in db the corresponding record is deleted

### DIFF
--- a/app/controllers/issues_controller.rb
+++ b/app/controllers/issues_controller.rb
@@ -32,6 +32,11 @@ class IssuesController < AuthenticatedController
     @first_evidence  = Evidence.where(node: @first_node, issue: @issue)
 
     load_conflicting_revisions(@issue)
+
+    @subscription = Subscription.find_by(
+                      user: current_user,
+                      subscribable_type: @issue.class.to_s,
+                      subscribable_id: @issue.id)
   end
 
   def new
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -1,0 +1,28 @@
+class SubscriptionsController < AuthenticatedController
+  include ProjectScoped
+
+  def create
+    subscription = Subscription.new(subscription_params)
+    subscription.user = current_user
+    subscription.save
+
+    redirect_to [@project, subscription.subscribable], notice: 'Subscribed!'
+  end
+
+  def destroy
+    subscription = Subscription.find_by(
+      user: current_user,
+      subscribable_type: subscription_params[:subscribable_type],
+      subscribable_id: subscription_params[:subscribable_id]
+    )
+    subscription.destroy
+
+    redirect_to [@project, subscription.subscribable], notice: 'Unsubscribed!'
+  end
+
+  private
+
+  def subscription_params
+    params.require(:subscription).permit(:subscribable_type, :subscribable_id)
+  end
+end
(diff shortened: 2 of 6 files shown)


app/controllers/issues_controller.rb
```diff
@@ -32,6 +32,11 @@ def show
     @first_evidence  = Evidence.where(node: @first_node, issue: @issue)
 
     load_conflicting_revisions(@issue)
+
+    @subscription = Subscription.find_by(
```
aapomm: Can we shorten this with the syntax: `@issue.subscriptions.where(user: current_user)`?

spec/features/issues_spec.rb
```diff
@@ -292,6 +292,9 @@
         let(:trackable) { @issue }
         it_behaves_like 'a page with an activity feed'
 
+        let(:subscribable) { @issue }
+        it_behaves_like 'a page sith subscribe/unsunscribe links'
```
aapomm: Typo here. (same with the shared examples)

palpatine

spec/support/subscription_shared_examples.rb
```diff
@@ -0,0 +1,25 @@
+# Define the following let variables before using these examples:
+#   subscribable: an instance of the subscribable model
+shared_examples 'a page sith subscribe/unsunscribe links' do
+  it 'subscribes and unsubscribes with the provided links' do
+    click_link 'Subscribe'
+    expect(page).to have_text 'Subscribed!'
+    expect(
+      Subscription.find_by(
```
aapomm: Same shorten as above

app/controllers/subscriptions_controller.rb
```diff
@@ -0,0 +1,28 @@
+class SubscriptionsController < AuthenticatedController
+  include ProjectScoped
+
+  def create
+    subscription = Subscription.new(subscription_params)
+    subscription.user = current_user
+    subscription.save
+
+    redirect_to [@project, subscription.subscribable], notice: 'Subscribed!'
```
aapomm: Probably add more info to the `notice` like `Subscribed to Issue SSL Vulnerability` etc. Same with unsubscribe.

app/controllers/issues_controller.rb
dormi: I was trying to avoid using something like this: https://github.com/dradis/dradis-ce/blob/master/app/models/issue.rb#L19

spec/features/issues_spec.rb
dormi: 😆

app/controllers/subscriptions_controller.rb
dormi: We are exactly viewing the item in question, let's to move on. 
In fact I am not sure of adding a :notice at all.

etdsoft:
Testing seems to be quite lightweight?

```
> Subscription.create subscribable: Issue.last, user: User.last
> Subscription.count
1
> Issue.last.delete
> Subscription.count
1
```

Also:

```
> Subscription.create subscribable: Issue.last, user: User.last
> Subscription.count
1
> User.last.delete
> Subscription.count
1
```

Also, creating an Issue doesn't end up in you getting subscribed to the issue, which is weird, added a ToDo for it.

Finally, the subscribe/unsubscribe seems to be working as expected, but I really think we need a way for user A to know who else is subscribed, otherwise someone that *should* be getting notifications may not be subscribed on the first place.

dormi:
captura de pantalla 2018-07-05 a las 13 33 14

PR #284, opened 2018-07-05
Subscribe/Unsubscribe links (2)

dormi:
**NOTE:** Rebuilding https://github.com/dradis/dradis-ce/pull/281

### Spec
The user should be able to manually subscribe or unscubscribe to/from a subscribable item. 

**Proposed solution**
Add links to the subscription resource from a subscribable show page, so the user can edit her subscription for that item.

### How to test

- Visit an issue   
- Assert there is a link to subscribe to that issue.
- Subscribe and check subscriptions table on the db for a new record
- Unsubscribe and check in db the corresponding record is deleted


etdsoft:
Looking food @dormi 👍

PR #285, opened 2018-07-05
Some final changes for the Project URL stuff

georgemillo:
A few things came up when merging the previous changes into Pro. This PR is mostly a "backport" of the changes in Pro to keep things as similar as possible and avoid future merge conflicts or other compatibility issues.

### DIFF
--- a/app/controllers/concerns/project_scoped.rb
+++ b/app/controllers/concerns/project_scoped.rb
@@ -18,7 +18,7 @@ module ProjectScoped
   #   https://github.com/airblade/paper_trail#metadata-from-controllers
   #
   def info_for_paper_trail
-    ;
+    { project_id: @project.id } if @project
   end
 
   def set_project
--- a/app/controllers/concerns/project_scoped.rb
+++ b/app/controllers/concerns/project_scoped.rb
@@ -2,27 +2,34 @@ module ProjectScoped
   extend ActiveSupport::Concern
 
   included do
-    before_action :set_project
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/project_scoped.rb
```diff
@@ -2,27 +2,34 @@ module ProjectScoped
   extend ActiveSupport::Concern
 
   included do
-    before_action :set_project
+    # Use prepend_before_action to keep things consistent with Pro:
+    prepend_before_action :set_project
```
etdsoft: why do we need to `prepend`?

app/controllers/concerns/project_scoped.rb
```diff
@@ -2,27 +2,34 @@ module ProjectScoped
...
+  # makes things clearer when dealing with CE/Pro merges in future.
   def info_for_paper_trail
-    ;
   end
 
+  def set_nodes
+    @nodes = Node.in_tree
+  end
+
+  # In Pro this method will load a Project using params[:project_id]. In
```
etdsoft: I'm not convinced about the need for this comment.

app/controllers/concerns/project_scoped.rb
georgemillo: In Pro it's necessary because, as explained in comments in that repo, `set_project` must be run before PaperTrail calls `info_for_paper_trail` (which happens in a before_action.)

In CE it doesn't make a difference, but I figured it's worth keeping things the same as Pro for consistency and future-proofing.

PR #288, opened 2018-07-06
Two small things I missed regarding Unicorn

georgemillo:
These should have been included in #287. Sorry!


etdsoft:
Can you please remove f5a243b? This is a standard Rails file that we keep for compatibility with the framework and future upgrades.

georgemillo:
done

PR #289, opened 2018-07-09
User Activity and General Presenter

aapomm:
### Spec
Currently, we have an ActivityPresenter that we use when we display activities. With the comments and the notifications coming into play, we can use some of the methods from the ActivityPresenter to display these items such as the avatar, partial, etc. to avoid reinventing the wheel.

**Proposed Solution**
We need to create a general/higher Presenter that the other presenters can inherit from with the reusable methods.

Additionally, to make the general Presenter code be more general we need to associate Activities with users (now that we have the proper User DB) instead of having a user email attribute.

### How to test
This is a refactor PR, but we need to make sure that the Recent Activities view is unchanged.

### Documentation
(Link to GitHub files or commits where this feature has been documented)


app/presenters/activity_presenter.rb
```diff
@@ -1,30 +1,6 @@
...
-                when 'Node'
-                  'fa-folder-o'
-                when 'Note'
-                  'fa-file-text-o'
-                else
-                  ''
-                end
-    h.content_tag :span, nil, class: icon_css
-  end
+  collection :trackable
```
dormi: collection sounds plural and this is a one to one relationship

aapomm: you are right. using `belongs_to` instead.

db/migrate/20180614094306_add_user_id_for_activities.rb
```diff
@@ -0,0 +1,23 @@
+class AddUserIdForActivities < ActiveRecord::Migration[5.1]
+  def change
+    add_reference :activities, :user, index: true
+
+    reversible do |dir|
+      dir.up do
+        Activity.find_each do |activity|
+          id = User.find_by_email(activity.read_attribute(:user)).id
```
georgemillo: what if the User doesn't exist in the DB anymore? This would cause an `undefined method on Nil` error

db/migrate/20180614094306_add_user_id_for_activities.rb
```diff
@@ -0,0 +1,23 @@
...
+          activity.update_attribute :user_id, id
+        end
+      end
+
+      dir.down do
+        Activity.find_each do |activity|
+          email = User.find_by_id(activity.user_id).email
```
georgemillo: same question here

db/migrate/20180614094306_add_user_id_for_activities.rb
georgemillo: Come to think of it, what if they still exist but they've changed their email address?

brianp:

PR #290, opened 2018-07-09
Notifications navbar

aapomm:
### Spec
Due to adding comments, we now need notifications to notify the parties involved.

We need a place to see these notifications.

**Proposed solution**
Add a notifications dropdown view that is loaded through AJAX.

### How to test
Use rails console to add sample notifications 
Assert that the dropdown displays the notifications

### Documentation
(Link to GitHub files or commits where this feature has been documented)


app/assets/javascripts/snowcrash/modules/notifications.js.coffee
```diff
@@ -0,0 +1,39 @@
...
+            $footer.hide()
+            $loading.show()
+
+          success: (data) ->
+            $('.js-notification-count').html("Notifications (#{data.count})")
+
+            if data.count == 0
+              $empty.show()
+            else
+              for notification in data.notifications
```
dormi: We may build the html server side.
Using here just a `remote: true` link + `format :js` on the controller.

Thoughts?

app/assets/javascripts/snowcrash/modules/notifications.js.coffee
```diff
@@ -0,0 +1,21 @@
+document.addEventListener 'turbolinks:load', ->
+  if $('.js-notifications-dropdown').length
+    $container = $('.js-notification-container')
```
dormi: let's use `data-` identifiers

app/views/notifications/_notifications_list.html.erb
```diff
@@ -0,0 +1,25 @@
+<% if notifications.any? %>
```
dormi: I would call this file `_feed` to match activities and comments

dormi:
red_dot
No notifications but I still see a red dot on the bell?

app/assets/javascripts/snowcrash/modules/notifications.js.coffee
```diff
@@ -0,0 +1,21 @@
...
+    $count     = $('.js-notification-count')
```
dormi: Why not render the whole dropdown serverside?
We could skip almost all this js.

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,12 @@
+class NotificationsController < AuthenticatedController
+  include ProjectScoped
+
+  def index
+    @notifications = current_user.notifications.newest.limit(7)
```
dormi: We should try to show all unread notifications?

app/views/layouts/snowcrash/_notifications.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="dropdown-menu">
```
dormi: As mentioned above, we could render the whole dropdown server side and use very little js in `index.js.erb`.

Then we should also use cache with the user and notifications in the key.

app/presenters/notification_presenter.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def partial_path
+    partial_paths.detect do |path|
+      lookup_context.template_exists? path, nil, true
+    end || raise("No partial found for notification in #{partial_paths}")
+  end
+
+  def partial_paths
+    [
+      "activities/#{notification.notifiable_type.underscore}/#{notification.action}",
```
dormi: Looks like at the end we are rendering activities.
But I think we want to render notifications.

Example: a notifiable may be a comment, the action may be a mention. We are not tracking activities for a mention, but we may create notifications for them.

Also, if we want to add notifications to mintcreek, we may want to include the project in the notification.

app/views/notifications/_comment.html.erb
```diff
@@ -0,0 +1,8 @@
+<% commentable = comment.commentable %>
+<% if commentable.title? %>
+  <%= presenter.verb %> on the
```
dormi: This verb should be `commented` or `added a comment` for comments?
captura de pantalla 2018-07-16 a las 10 59 37

dormi: Also the link may point to the exact comment https://github.com/dradis/dradis-ce/blob/comments-feature/app/controllers/comments_controller.rb#L11

app/assets/stylesheets/snowcrash/modules/notifications.scss
```diff
@@ -0,0 +1,78 @@
+.notifications {
+  .notifications-dot {
+    display: none;
```
dormi: I liked this dot.
Can't we make it work?

aapomm: I just added this dot to show how it looks on the navbar. The functionality should be added with the ActionCable feature.

georgemillo:
Can we make the "mark as read" button(s) update by AJAX rather than triggering a full page reload?

It's not just an aesthetic thing - reloading the page resets any text inputs, meaning you could lose any unsaved changes. For example, I could easily see someone working on editing an issue only to get distracted by their notification icon, then click "mark as read", inadvertently refreshing the page and losing their work:

notifications

georgemillo:
Shouldn't the "see all notifications" link be visible even when I have 0 comments?

screen shot 2018-07-17 at 12 29 11
screen shot 2018-07-17 at 12 29 17

The dropdown is just for *unread* notifications, right? Then "see all notifications" will take them to a page that shows *all* notifications, unread and read. (Am I understanding correctly that that's how it's supposed to work? So just because I have no *unread* notifications, doesn't mean I don't necessarily want to see the link to view all my notifications.

aapomm:
@georgemillo 
> Can we make the "mark as read" button(s) update by AJAX rather than triggering a full page reload?

Good idea! Though I plan on doing this on a separate PR.

> The dropdown is just for unread notifications, right?

My idea for this is that the dropdown should show all the notifications including the ones that were read (seems to be what Basecamp does)

georgemillo:
(See Basecamp discussion.) Sounds like we're clear on what this should look like. Just worth making it explicit here because I didn't mention it in Basecamp - the answer to my question is 'no, it *shouldn't* show the "See all notifications" link when there are no unread notifications in the dropdown - because this would mean I've *never* received a notification so there wouldn't be anything to see on "See all notifications" anyway.

georgemillo:
Weird bug when you resize the window with the notification panel open:

disappearing-navbar

georgemillo:
Actually i just realised this happens on `master` and is in fact a more general problem with the navbar, not specifically related to the notifications dropdown:

disappearing-navbar-2

georgemillo:
Looks like this one can be merged now? AFAIK all my comments above have either been addressed, or will/should be left to later PRs.

app/views/notifications/_comment.html.erb
```diff
@@ -0,0 +1,10 @@
+<% commentable = comment.commentable %>
```
dormi: @aapomm I just noticed: we must handle the case when the comments has been deleted ?

georgemillo: We create an Activity on comment creation, update and deletion... but I don't see why we need a *notification* for update and deletion.

I can't think of any service I use that sends me a notification when a post/comment/whatever is *deleted or edited* as opposed to just published/created.

If I'm right this partial can be much much simpler because e.g. we don't need to bother with `presenter.verb` as we already know what the verb will be ("created"!)

app/assets/javascripts/snowcrash/modules/notifications.js.coffee
```diff
@@ -0,0 +1,12 @@
+document.addEventListener 'turbolinks:load', ->
+  $dropdown = $('[data-id="js-notifications-dropdown"]')
```
etdsoft: let's stop using "js-" CSS classes for JS behaviors. Please update to: `$('[data-behavior~=notifications-dropdown]')`

app/assets/javascripts/snowcrash/modules/notifications.js.coffee
```diff
@@ -0,0 +1,12 @@
...
+
+  if $dropdown.length
+    $container = $('[data-id="js-notification-container"]')
```
etdsoft: probably can handle this via CSS selector (no js-)? not sure how this plays out w/ my data-behavior comment above

app/models/notification.rb
```diff
@@ -15,11 +17,17 @@ class Notification < ApplicationRecord
   # -- Scopes ---------------------------------------------------------------
   scope :unread,  -> { where(read_at: nil) }
   scope :read,    -> { where.not(read_at: nil) }
+  scope :newest,  -> { order(created_at: :desc) }
```
etdsoft: alpha order

app/views/layouts/snowcrash/_navbar.html.erb
```diff
@@ -14,6 +14,12 @@
           <li><%= link_to main_app.project_upload_manager_path(@project) do %><i class="fa fa-cloud-upload fa-lg"></i> Upload output from tool<% end %></li>
           <li><%= link_to main_app.project_export_manager_path(@project) do %><i class="fa fa-file-text-o fa-lg"></i> Export results<% end %></li>
           <li><%= link_to main_app.configurations_path do %><i class="fa fa-cog fa-lg"></i> Configuration<% end %></li>
+          <li class="notifications dropdown">
+            <%= link_to main_app.project_notifications_path(@project), class: 'dropdown-toggle', data: {id: 'js-notifications-dropdown', toggle: 'dropdown'}, remote: true do %>
```
etdsoft: indent

app/views/notifications/_comment.html.erb
etdsoft: What's the outcome of this conversation? I think @georgemillo has a point re: Notifications only needed on create?

app/views/notifications/_notifications.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <ul class="notification-container" >
+    <div class="notification-header">
+      Notifications
+      (<%= unread_count %>)
+      <div class="pull-right">
+        <a href="#">Mark all as read</a>
+      </div>
+    </div>
+    <% if notifications.any? %>
+      <%= render partial: 'feed', locals: {notifications: notifications} %>
```
etdsoft: {} spacing

app/models/notification.rb
```diff
@@ -15,11 +17,17 @@ class Notification < ApplicationRecord
...
 
   # -- Class Methods --------------------------------------------------------
 
   # -- Instance Methods -----------------------------------------------------
   def read?
     self.read_at
   end
+
+  def unread?
```
etdsoft: you can probably remove this and update the only place in the view where that's used with a call to `read?`

app/models/notification.rb
```diff
@@ -4,6 +4,8 @@ class Notification < ApplicationRecord
   belongs_to :recipient, class_name: 'User'
   belongs_to :notifiable, polymorphic: true
 
+  alias_method :user, :recipient
```
etdsoft: This seems to only be used by the presenter? I see how it may be interesting to have a common interface between Notification and Activity if we're going to create the general presenter, but that refactor should be part of the extraction of the presenter. 

In this PR I'd stick to :recipient throughout the code

app/views/notifications/_comment.html.erb
georgemillo: Looks like aaron [removed](https://github.com/dradis/dradis-ce/commit/db7dcb2d7533e6058e4b844a83498fbddf82d86e#diff-ecab1e17a4becfd34c67ccabc4181a83L36) the `verb` method from the presenter so my comment is outdated? Although we may come back to this when we start adding new types of notification.

C/f [my comment here](https://github.com/dradis/dradis-ce/pull/303/files#r203689765) which is relevant.

PR #291, opened 2018-07-09
Comments automatically update using ActionCable [WIP]

georgemillo:
Please note when testing or developing, because I lost hours to this issue yesterday: **if you restart the server during the course of development, you must also log out and log back in within Dradis.** Otherwise you get the following weird error in the Chrome console: `"WebSocket connection to 'ws://localhost:3000/cable' failed: WebSocket is closed before the connection is established."`

I'm not sure exactly what's going on here, but interestingly it only seems to happen with unicorn, not puma. I think we can get away with the 'log out and log back in' solution for now because afaik as I know it only affects developers, not the end user. But it needs more testing - my understanding of ActionCable and Websockets is still quite limited so hopefully I'll figure it out as I make more progress on this PR.

If this becomes a bigger issue or if anyone wants to learn more, I can write up a longer explanation of exactly what I've tried to fix it and what I've learned. But we have bigger fish to fry for now.


georgemillo:
This is almost ready (I think) but there's one major point that needs fixing: if you create/update a comment, it appears for everyone else as it should - but *everyone* will see the 'Edit'/'Delete' links for the new comment, even if they shouldn't be allowed to edit/delete it themselves.

I'll work on this first thing tomorrow, just noting it here as a "todo".

Also note that I had to add `puma` in the test environment to make the tests which involve websockets work, see the comments [here](https://github.com/dradis/dradis-ce/pull/291/commits/a88ed11218476ba985009d455fb2dc2475a4bc25#diff-e11bdaf9116be3911e9557b69e242370R43) for an explanation.

georgemillo:
Potentially of public interest: here are some ActionCable-related links I found helpful for figuring this out:

http://guides.rubyonrails.org/action_cable_overview.html (obviously)
https://www.youtube.com/watch?v=n0WUjGkDFS0 (tutorial by DHH)
https://blog.heroku.com/real_time_rails_implementing_websockets_in_rails_5_with_action_cable
https://blog.codeship.com/actioncable-the-missing-guide/
https://www.learnenough.com/action-cable-tutorial

georgemillo:
Some notes on the server setup here. I have other notes which I'll leave to a separate comment.

- As explained [here](https://github.com/dradis/dradis-ce/pull/291/commits/a88ed11218476ba985009d455fb2dc2475a4bc25#diff-e11bdaf9116be3911e9557b69e242370R43), the test env now uses `puma`. One annoying side-effect of this is that you now get a bunch of puma logging output when you run the tests:
    screen shot 2018-07-11 at 14 08 25

    There's probably some way to suppress this but I think we can live with it for now.
- In the long run it's probably a bad idea to use a different server for our test env than for dev and prod. I tried to replace `puma` with `unicorn` in the test env using [the instructions here](https://stackoverflow.com/questions/4585544/how-to-run-cucumber-selenium-using-thin-server-instead-of-default-webrick) but couldn't get it to work. Again, not ideal, but there are bigger problems for now.

georgemillo:
Okay, now on to the more important stuff regarding the actual implementation of the feature.

First of all, please note that this is the first major thing I've built with ActionCable so I'm still unsure about best practices.

When a comment is created or updated, I'm [rendering the partial on the server side](https://github.com/dradis/dradis-ce/pull/291/files#diff-ad7009a67ee4df1721dd8e449dffec56R36) and sending the HTML back through the websocket, where it's then added into the DOM by jQuery.

The problem is that `ActionCable.server.broadcast`, at least the way I'm using it, broadcasts the same message to *everyone*, including the same HTML. But not everyone should see the same comment, because the "Edit"/"Delete" links should only be visible if its *your* comment. And ActionCable can't customise the HTML per user because `broadcast` is being called from within the controller, which only knows about the `current_user`.

The least-bad solution I could think of was to remove [the `can?` stuff](https://github.com/dradis/dradis-ce/pull/291/files#diff-1aae7ed6c016b3e62cf25303b9a3f452L18) from the view as this is dependent upon `current_user`, and *always* include "Edit"/"Delete" etc in the rendered partial. Then on document.ready, [jQuery removes these links](https://github.com/dradis/dradis-ce/pull/291/files#diff-46e91fef6da8db1c6d2a068f4470bcbcR23) from where they don't belong. The `comment_channel` websocket JS class [does the same thing](https://github.com/dradis/dradis-ce/pull/291/files#diff-906caf1885793aeee5915611f6a778d5R30) when receiving a new or updated comment that it needs to add to the page.

This feels very hacky, not least because I had to add [this](https://github.com/dradis/dradis-ce/pull/291/files#diff-71a660186bdeee472a8683ee9759c79dR25) to the layout so that the JS can figure out which comments belong to the current user. I don't like it at all. There may be some smarter way of doing this with ActionCable, e.g. something that broadcasts different HTML templates to different users through their websockets. But after reading several ActionCable tutorials I can't figure it out.

Anyone have any suggestions and/or better ideas?

app/views/layouts/snowcrash.html.erb
```diff
@@ -19,7 +19,12 @@
 
   <%= csrf_meta_tag %>
 </head>
-<body class="<%= controller_name %> <%= action_name %>">
+<body
+  class="<%= controller_name %> <%= action_name %>"
+  <% if current_user %>
+    data-current-user-id="<%= current_user.id %>"
```
etdsoft: What about using:

https://github.com/secroots/dradis-pro/blob/develop/app/views/activities/_poller.html.erb

georgemillo: you mean put the `data` attribute there instead? Probably makes more sense.

Gemfile
```diff
@@ -107,9 +103,9 @@ gem 'rails-html-sanitizer', '~> 1.0.4'
 gem 'sqlite3'#,  '1.3.10'
 
 # Use Unicorn as the web server
-# FIXME: Switch to Puma for Rails 5
-gem 'unicorn',  '4.9.0', group: :production
-
+gem 'unicorn',  '4.9.0'
```
dormi: are we using an outdated version of `unicorn`? why?

Gemfile
```diff
@@ -188,6 +184,7 @@ group :development, :test do
 end
 
 group :test do
+  gem 'puma' # required for tests which involve websockets
```
dormi: is this solved with a newer version of `unicorn` ? probably not, but had to ask...

app/controllers/comments_controller.rb
```diff
@@ -7,23 +7,34 @@ class CommentsController < AuthenticatedController
   def create
```
dormi: In this controller I miss the `track_created`, `track_updtaed` and `track_destroyed` calls.

dormi: If I understand this correctly we are pushing `comment` related info through action cable.
I think the idea would be to make it generic and push `activity` related info.

app/assets/javascripts/channels/comment.coffee
```diff
@@ -0,0 +1,53 @@
...
+    @action    = data.action
+    @commentId = data.comment_id
+    @html      = data.html
+
+  # We need to make sure that the comment we've been notified about belongs
+  # on the currently-viewed page
+  isRelevant: ->
+    @$feed.length != 0
+
+  toggleActions: ->
```
dormi: we should find a way to skip this. If the user disables js or changes its html via browser console, he may see the forbidden links?

app/assets/javascripts/channels/comment.coffee
```diff
@@ -0,0 +1,53 @@
...
+    $comment = $("#comment_#{@commentId}")
+    unless $('body').data('currentUserId') == $comment.data('userId')
+      $comment.find('.actions, .edit_comment').remove()
+
+  updateCounters: ->
+    count = $(".comments-list div.comment").length
+    $("#comments_feed_no_comments").toggle(count == 0)
+    $("#comments-count-badge").text(count)
+
+  process: ->
```
dormi: Idea: remove custom js to update a specific page and reload the page with `TurboLinks` when `ActionCable` says so

Gemfile
georgemillo: yes, the most recent version of `unicorn` is `5.4.0`.  But `4.9.0.` is what we were using already so I didn't bother upgrading it (the only change here is that `unicorn` is no longer restricted to the production env only

app/assets/javascripts/channels/comment.coffee
georgemillo: If we're going to reload the entire page, what's the point in using ActionCable? We might as well just stick with polling since the JS would be simple (essentially, `if serverSaysWeMustRefresh then refresh()`) and ActionCable is big overhead.

Plus it would be very jarring to hit people with a full-page refresh that happens at a point they have no control over (i.e. when someone else comments).

app/assets/javascripts/channels/comment.coffee
georgemillo: That's the point - I know this is a crappy solution, but it's the best I could come up given my limited understanding of ActionCable.

Worth noting that even if the user can *see* links he shouldn't, they won't work because the user will be validated on the server-side (or at least my understanding is that this is being handled by `cancancan`.) That's probably worth writing a request spec for in any case.

Gemfile
georgemillo: Okay, after spending a bit more time on this, it turns out that `puma` is configured for Capybara [by default](https://github.com/teamcapybara/capybara/blob/2.15.1/lib/capybara.rb#L451-L454). `:unicorn` isn't in that file so you need to add it manually.

I tried removing `puma` from the Gemfile and sticking the following in `rails_helper` but it made the tests fail (`rspec spec/features/issue_pages/show_page_spec.rb` seem like the most relevant ones to try seeing as they're the ones which actually use ActionCable):

```ruby
Capybara.register_server :unicorn do |app, port, host, options={}|
  Rack::Handler::Unicorn.run(app, {Host: host, Port: port, Threads: "0:4", workers: 0, daemon: false}.merge(options))
end
Capybara.server = :unicorn
```

I'm probably passing the wrong settings to `Unicorn.run` (the above are just copied from what Capybara passes to `puma`) but I have no idea what to try instead.

Things I've tried that didn't work:

1. Updating `unicorn` to the latest version (5.4.0)
2. Updating `capybara` to the latest version (3.3.1)
3. Copying `config/unicorn.rb` from Pro to CE and passing it as an option (`config_file: ...`) to `Unicorn.run` (not surprised that this didn't work as that file is meant to set Unicorn up for a production environment.

I feel very uneasy using a different server for the production environment vs. the testing environment (although note that this is what we were doing before anyway, since production used unicorn while dev/test used WEBrick), especially since the difference between `puma` and `unicorn` can apparently have such a big effect on whether tests pass. But I don't know how we can fix this and I'm out my depth. I give up for now.

(Of course the other option is to switch to `puma` in dev/prod, as opposed to using `unicorn` in tests. I don't understand enough about the difference between the two servers to have a preference for which we use. The more important point is that we can't continue to use WEBrick in tests, or anywhere for that matter, if we want to start using ActionCable.)

dormi: > although note that this is what we were doing before anyway, since production used unicorn while dev/test used WEBrick

from this perspective, it does not seem that terrible if we use `puma`in dev/test and skip this by fix by now

georgemillo: Yeah, I think we can get away with it for now. Just wanted to give some more details about what I've tried as they may be helpful when trying to fix it in the future.

georgemillo:
Closed in favour of #306, but we may come back to it eventually.

PR #294, opened 2018-07-11
Notifications creation

dormi:
### Spec
We comment on a model, let's say an issue, that we didn't create.
Notifications must be stored on db, so the user that created that issue and the rest of subscribers may see it in their notifications list.

No notification should be created for the comment creator.

We may trigger the notifications creation when activity is created. But it must be done in the background.

**Proposed solution**
We may schedule bg job from the ActivityTracking concern, that creates the necessary notifications.

### How to test
If https://github.com/dradis/dradis-ce/pull/306 is merged:
  - Create a comment on an issue
  - Notifications should be created in db for all users subscribed to that issue except for the commenter

Else:
  - Specs should passs


spec/features/node_pages_spec.rb
```diff
@@ -3,7 +3,10 @@
 describe "node pages" do
   subject { page }
 
-  before { login_to_project_as_user }
+  before do
+    ActiveJob::Base.queue_adapter = :test
```
georgemillo: Why do this here? Couldn't it just go in `rails_helper`?

app/jobs/activity_tracking_job.rb
```diff
@@ -0,0 +1,23 @@
+class ActivityTrackingJob < ApplicationJob
+  queue_as :dradis_project
+
+  def perform(action:, trackable_id:, trackable_type:, user:)
+    trackable = trackable_type.constantize.find(trackable_id) rescue nil
```
aapomm: @dormi looks like an anti-pattern to rescue exceptions here. Might be better to use another method instead to check for existence?

spec/jobs/activity_tracking_job_spec.rb
```diff
@@ -0,0 +1,66 @@
+require 'rails_helper'
+
+describe ActivityTrackingJob do #, type: :job do
+
+  it 'it uses correct queue' do
```
aapomm: redundant `it`

spec/jobs/multi_destroy_job_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 describe MultiDestroyJob do #, type: :job do
 
-  it 'is uses correct queue' do
+  it 'it uses correct queue' do
```
aapomm: redundant `it`

spec/features/node_pages_spec.rb
dormi: @georgemillo you are right, changed.

app/jobs/activity_tracking_job.rb
```diff
@@ -0,0 +1,23 @@
+class ActivityTrackingJob < ApplicationJob
+  queue_as :dradis_project
+
+  def perform(action:, trackable_id:, trackable_type:, user:)
```
aapomm: just pass the user email here?

dormi: I would prefer to keep the user as it is. I think that in the future we are going to change user.email for user.id in most models, like in Activity, and I already have the user ready this way.
Also, we may need to pass the user to create other notifications, not sure yet.

georgemillo: I don't think you can pass the `User` in directly, or any "complex" Ruby object. In production* jobs use Resque, which means they'll be serialized using Redis, which can only store primitive types like strings and integers.

I've run into this issue before on another project and had to pass the record's `id` to the job, then inside the job load it again with `Record.find(id)`. (If you want to be really pedantic, reloading the record is probably a good idea anyway because the DB row may have changed since the job was queued and the job may need the most up-to-date data... unlikely to be an issue here but worth bearing in mind for future BG jobs.)

*or at least we use Resque in all envs in Pro? See `config/application.rb` in that repo. Not sure if this is the case in CE even though `resque` is included in CE's Gemfile. Whatever the case these changes are going to merged into Pro eventually so we should make them compatible with Resque now.

This is actually a great example of why it's a good idea to use the same ActiveJob adapter in all environments for the sake of [dev-prod parity](https://12factor.net/dev-prod-parity)... otherwise these kind of issues slip through the net.

dormi: Updated to use user_id

georgemillo:
This works, providing you merge #306 into it first. It gets a 👍 from me.

I'm just concerned that if we merge this into `notifications-feature` before that branch contains the work on #306, we'll cause ourselves headaches later when we try to merge everything properly. Not sure what the best way to handle this is, but my suggestion would be:

1. Finish #306 and merge it into `comments-feature`
2. merge `comments-feature` into `notifications-feature` so it has the latest changes
3. merge `notifications-feature` into `notifications-creation` and resolve any conflicts
4. merge #294 (i.e. merge `notifications-feature` back into `notifications-creation`)

Is this the best way to do it?

aapomm:
@georgemillo I also think that's the way to go

PR #303, opened 2018-07-17
Long running feature branch to add Notifications [WIP]

georgemillo:


db/migrate/20180705112109_create_notifications.rb
```diff
@@ -0,0 +1,20 @@
...
+      t.references  :actor, index: true, foreign_key: { to_table: :users }
+      t.references  :recipient, index: true, foreign_key: { to_table: :users }
+
+      t.timestamps
+    end
+
+    add_index :notifications,
+              [:notifiable_id, :notifiable_type, :recipient_id],
+              unique: true,
+              name: 'index_notifications_on_notifiable_and_recipient'
```
georgemillo: This index was [added in #294](https://github.com/dradis/dradis-ce/pull/294/files#diff-3f2b709644c202610cd166dc7224ad0aR17) but this point snuck past me when I was reviewing that PR:

Do we really want a `unique` index here? This would mean that the same person can't be notified twice about the same record - even for a different action. So say we want to notify people when an issue is both created *and* when it's updated, or whatever other action... this uniqueness constraint means we can't. 

I can't think of an obvious example where we'd definitely want to notify users about two different types of action to the same record - e.g. [as discussed here](https://github.com/dradis/dradis-ce/pull/290/files#r203348039) it would be weird to notify people about comment *updates* as well as comment creation... but I don't see why every record is necessarily like this. Plus in future we might want to notify people about some action that isn't a simple CRUD update.

Even if we can't think of any obvious cases now where this shouldn't be unique, I don't think we should shut ourselves off from the possibility.

georgemillo:
Just noticed that when I have this branch checked out and click "upload output from tool" I get an error:

screen shot 2018-07-19 at 16 42 26

This is obviously something to do with the project ID scoping changes. Did something sneak its way into this PR which shouldn't have? Or did something *not* get merged into here that *should* have?

PR #304, opened 2018-07-17
Long-running feature branch to add subscriptions [WIP]

georgemillo:
As of July 17th, my understanding is that the subscriptions feature is ready - we're just waiting to finish notifications so we can merge #303 into this one, then merge into comments and then finally `master`.


app/controllers/issues_controller.rb
```diff
@@ -32,6 +32,11 @@ def show
     @first_evidence  = Evidence.where(node: @first_node, issue: @issue)
 
     load_conflicting_revisions(@issue)
+
+    @subscription = Subscription.find_by(
```
georgemillo: On a purist note I think it's best to use `find_by!` instead of `find_by` if we're not excepting that it will ever return `nil`. That way if something goes wrong and it *does* return `nil`, the error gets caught right here instead of bubbling down to the view or wherever where it can be much harder to debug.

app/controllers/subscriptions_controller.rb
```diff
@@ -0,0 +1,28 @@
+class SubscriptionsController < AuthenticatedController
+  include ProjectScoped
+
+  def create
+    subscription = Subscription.new(subscription_params)
+    subscription.user = current_user
+    subscription.save
```
georgemillo: likewise here with `save!` vs `save`

app/views/subscriptions/_feed.html.erb
```diff
@@ -0,0 +1,6 @@
+<h3>Subscribers <span class="badge"><%= subscribable.subscriptions.count %></span></h3>
```
georgemillo: could it be worth adding a `counter_cache` on `subscriptions.count`?

PR #306, opened 2018-07-17
Comment activities 2

dormi:
### Spec
Add activity tracking to comments, as we have it for other resources

**Proposed solution**
~~Try to use different approach: use callbacks in the model that trigger a bg task to create the corresponding activity.~~

We are going to use the old way (track activity from controller), ~~and try to move to actioncable for the required page changes.~~

### How to test

- Comment on an Issue
- The main page, assert we can see the activity.
- Edit + delete the issue, check those activities too
- With another browser/user, visit the same issue and create/update/delete a comment. Assert that in the first browser wee see those changes.

### DIFF
--- a/app/models/concerns/commentable.rb
+++ b/app/models/concerns/commentable.rb
@@ -4,4 +4,10 @@ module Commentable
   included do
     has_many :comments, as: :commentable, dependent: :destroy
   end
+
+  def commentable_activities
+    Activity.where(trackable_type: self.class.to_s, trackable_id: self.id).or(
+      Activity.where(trackable_type: 'Comment', trackable_id: [self.comments.map(&:id)])
+    )
+  end
 end
--- a/app/models/concerns/commentable.rb
+++ b/app/models/concerns/commentable.rb
@@ -6,8 +6,11 @@ module Commentable
   end
 
   def commentable_activities
-    Activity.where(trackable_type: self.class.to_s, trackable_id: self.id).or(
-      Activity.where(trackable_type: 'Comment', trackable_id: [self.comments.map(&:id)])
+    self.activities.or(
+      Activity.where(
+        trackable_type: 'Comment',
+        trackable_id: self.comments.pluck(:id)
+      )
     )
   end
 end
--- a/app/views/activities/_comment.html.erb
+++ b/app/views/activities/_comment.html.erb
@@ -3,7 +3,7 @@
   <% if comment.commentable.title? %>
     <% title = comment.commentable.title %>
   <% end %>
-  <%= presenter.verb %> a comment on <%= link_to title, polymorphic_path([@project, comment.commentable], anchor: dom_id(comment)) %>.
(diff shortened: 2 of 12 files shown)


georgemillo:
@dormi I see the "person X added a comment" activity on the central project activity feed, but not on the activity feed for the issue itself?

screen shot 2018-07-18 at 10 58 06

screen shot 2018-07-18 at 10 58 12

app/views/activities/_comment.html.erb
```diff
@@ -0,0 +1,13 @@
+<% if comment %>
+  <% title = "#{comment.commentable.class.to_s.capitalize} ##{comment.commentable.id}" %>
+  <% if comment.commentable.title? %>
+    <% title = comment.commentable.title %>
+  <% end %>
+  <%= presenter.verb %> a comment on <%= link_to title, polymorphic_path([@project, comment.commentable], anchor: dom_id(comment)) %>.
```
georgemillo: How easy would it be to change this so instead of saying "Person X created a comment on Issue Y", it said "Person X commented on Issue Y"? The first one isn't wrong but it's a slightly weird way to say it.

I realise we're a bit restricted by the need to make the activities partials/presenters as reusable as possible, if making this change is going to take more than a few minutes then I wouldn't bother.

georgemillo: Also, the link here is a tad confusing:

screen shot 2018-07-18 at 11 05 48

When I view this I'm going to think "okay, what do I have to click to see the comment"? It's not obvious that the link "Issue 2" is a link to *the specific comment* (i.e. it has an anchor tag in the URL) rather than a link to the issue in general.

A more intuitive way to do it, I think, would be for the text to say "admin created [a comment](#) on [Issue 2](#)", where the first link has the anchor tag and the second one doesn't. Or if we change the wording as per my comment above: "admin [commented](#) on [Issue 2](#)"

dormi: Done this, but:

> I realise we're a bit restricted by the need to make the activities partials/presenters as reusable as possible

This is why I opted for "Person X added a [link:comment] to [link:issue]", so the update and create cases are identical.

georgemillo: 👍

app/models/concerns/commentable.rb
```diff
@@ -4,4 +4,10 @@ module Commentable
   included do
     has_many :comments, as: :commentable, dependent: :destroy
   end
+
+  def commentable_activities
+    Activity.where(trackable_type: self.class.to_s, trackable_id: self.id).or(
```
etdsoft: why this instead of `self.activities.or(` ?

app/models/concerns/commentable.rb
```diff
@@ -4,4 +4,10 @@ module Commentable
...
+      Activity.where(trackable_type: 'Comment', trackable_id: [self.comments.map(&:id)])
```
etdsoft: I'd use `self.comments.pluck(:id)`

georgemillo: I'd go one better and use `comment_ids`  ;)

http://guides.rubyonrails.org/association_basics.html#methods-added-by-has-many

PR #307, opened 2018-07-18
Comments must preserve linebreaks and white space

dormi:
This is the problem we want to fix:
preview-full-comments

### DIFF
--- a/app/assets/stylesheets/snowcrash/modules/_comments.scss
+++ b/app/assets/stylesheets/snowcrash/modules/_comments.scss
@@ -28,6 +28,13 @@
     border-radius: 5px;
     background-color: #f1f3f3;
 
+    .content {
+      padding: 2em;
+      font-family: inherit;
+      border: none;
+      font-size: 1.1em;
+    }
+
     .time {
       color: #888;
       font-size: 12px;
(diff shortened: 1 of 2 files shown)


georgemillo:
The linebreaks look fine and I agree that `<pre>` is the way to go - but I don't like how each comment now takes up much more space:

Before:
screen shot 2018-07-18 at 11 12 42

After:
screen shot 2018-07-18 at 11 12 54

Can we remove the extra padding, especially from the bottom?

dormi:
@georgemillo tried to fix it. 
Also looks like the comment bg color was a bit different. Fixed that too.

etdsoft:
what's wrong with:

http://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-simple_format

dormi:
@etdsoft 

> what's wrong with:
   http://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-simple_format

The white space (more than 1).
Using a `<pre>` we can keep line breaks and white space.

dormi:
closing in favor of https://github.com/dradis/dradis-ce/pull/313

PR #308, opened 2018-07-18
[WIP] Notifications trigger

aapomm:
Prerequisites:
- https://github.com/dradis/dradis-ce/pull/294
- https://github.com/dradis/dradis-ce/pull/290

### Spec
We want to have a way to notify the user of any unread notifications without the need for a page load.

**Proposed Solution**
Place an icon in the top header menu to show if we have unread notifications.

Use ActionCable to check if the notifications indicator in the top menu must be turned red (unread notifications available)

When that indicator is clicked, a dropdown menu w/ the notifications is going to be loaded via ajax. This menu is part of another task. 

### How to Test
1. Open up a project and read any notifications, if any.
2. In a separate browser, create a notification for the user.
3. Assert that the notifications indicator was triggered


app/presenters/notification_presenter.rb
```diff
@@ -0,0 +1,90 @@
...
+                when 'Evidence'
+                  'fa-flag'
+                when 'Issue'
+                  'fa-bug'
+                when 'Node'
+                  'fa-folder-o'
+                when 'Note'
+                  'fa-file-text-o'
+                else
+                  ''
```
georgemillo: You need to add a case for `'Comment'` - `fa-comment` is [already being used](https://github.com/dradis/dradis-ce/pull/306/files#diff-fdf1910eed85fd4ade99060ceb5397d9R15) by the activity presenter.

But I don't like how this complicated case statement is duplicated between the two presenters - maybe we could extract a shared method called `fa_icon_for_record` so e.g. `fa_icon_for_record('Evidence')` returns `fa-flag`? Then the code here would be replaced with `icon_css << fa_icon_for_record(notification.notifiable_type)`, with something similar in ActivityPresenter

app/assets/javascripts/cable/subscriptions/notifications_channel.js.coffee
```diff
@@ -0,0 +1,9 @@
+App.cable.subscriptions.create 'NotificationsChannel',
+  connected: ->
+    console.log('Subscribed to NotificationsChannel.')
+
+  rejected: ->
+    console.log('Error subscribing to NotificationsChannel.')
+
+  received: (data)->
+    $('[data-id="js-notifications-dot"]').toggleClass('hidden')
```
dormi: data-behavior ?

app/views/activities/poll.js.erb
```diff
@@ -91,4 +91,4 @@
 <% end %>
 
 ActivitiesPoller.lastPoll = <%= @this_poll %>;
-ActivitiesPoller.poll();
+<%# ActivitiesPoller.poll(); %>
```
georgemillo: was it a mistake to leave this commented out?

aapomm: yes it is. reverting...

app/views/notifications/_notifications.html.erb
```diff
@@ -1,10 +1,10 @@
-<% cache [current_user, @notifications] do %>
```
georgemillo: Hopefully this isn't too pedantic but can this partial be renamed to something more descriptive like "notifications/_dropdown"? `notifications` is such a generic name and it confused me a bit when I was looking at this PR in order to figure out the best approach for the `index` page. I'd prefer a name which tells me exactly what the partial is used for and where rather than this name, which doesn't give me many clues.

dormi:
while on this brach I've see:
captura de pantalla 2018-07-20 a las 12 38 13
May be when using ActionCable we are using different threads to access sqlite? Or maybe it was due to a bg job?

aapomm:
@dormi My guess is it's due to the bg job. We don't touch the DB with ActionCable in the case of notifications. Somewhat related: I had issue with SQLite and the subscription unique index. Were you experiencing that too?

config/cable.yml
```diff
@@ -1,8 +1,10 @@
 development:
-  adapter: async
+  adapter: redis
```
dormi: we are introducing redis as a dependency in CE

what are the options? I would prefer to skip this

app/views/notifications/_feed.html.erb
```diff
@@ -1,21 +1,3 @@
...
-          <%= notif_presenter.render_title %>
-        </div>
-
-        <div class="details">
-          <%= notif_presenter.avatar_with_link(30) %>
-        </div>
-      </div>
-    </li>
-  <% end %>
+  <%= render partial: 'item', locals: { notification: notification } %>
```
dormi: I think this partial can be called `notification` and this could be just `render notification`

app/views/notifications/_comment.html.erb
```diff
@@ -1,5 +1,6 @@
 <% commentable = comment.commentable %>
-<% comment_path = polymorphic_path([@project, commentable], anchor: dom_id(comment)) %>
+<% project = commentable.node.project %>
```
dormi: I don't get this `commentable.node.project`.
What if in the future we want to comment on a node?
I think this fixes the problem now, but this is not generic.
How to fix this?

app/models/concerns/notifiable.rb
```diff
@@ -8,13 +8,24 @@ module Notifiable
...
           recipient: recipient
         )
+
+        broadcast_to_user(notification, user)
       end
     end
   end
+
+  def broadcast_to_user(notification, user)
+    project = self.commentable.node.project
```
dormi: same comment as below, not sure about this way of getting the project.

app/views/notifications/_comment.html.erb
dormi: If we don't want to add `project_id` to `comments` or to `issues`, may be we could add a `project` method to the `Issue`model

app/models/concerns/notifiable.rb
```diff
@@ -8,13 +8,24 @@ module Notifiable
...
       recipients.each do |recipient|
-        Notification.create(
+        notification = Notification.create(
           action: action,
           actor: user,
           notifiable: self,
           recipient: recipient
         )
+
+        broadcast_to_user(notification, user)
```
dormi: I am an ActionCable newbie.
Are we broadcasting to the correct user? Should be `recipient` ?

app/controllers/concerns/notifications_reader.rb
```diff
@@ -0,0 +1,13 @@
+module NotificationsReader
+  protected
+
+  # Mark each notifications associated with the item as read
+  def read_item_notifications(item)
+    item.comments.each do |comment|
+      comment.notifications.where(recipient: current_user, read_at: nil).each(&:mark_as_read)
```
dormi: Look out, `mark_as_read` is now called `read!`

Gemfile
```diff
@@ -109,6 +109,7 @@ gem 'sqlite3'#,  '1.3.10'
 # Use Unicorn as the web server
 # FIXME: Switch to Puma for Rails 5
 gem 'unicorn',  '4.9.0', group: :production
```
dormi: are we using `unicorn` in dev too? is this `group: production` necessary?

config/cable.yml
aapomm: @dormi no way around this unfortunately. we need `redis` for `ActionCable`.

app/models/concerns/notifiable.rb
aapomm: You are correct.

app/views/notifications/_comment.html.erb
aapomm: I'm going with adding a `project` method (delegation) to the `Note` and `Issue` model same as in Pro

config/cable.yml
dormi: @aapomm https://edgeguides.rubyonrails.org/action_cable_overview.html#subscription-adapter

app/assets/stylesheets/snowcrash/modules/notifications.scss
```diff
@@ -59,6 +58,12 @@
           }
         }
 
+        .notification-read {
+          position: absolute;
```
dormi: we may use more bootstrap here. Like `pull-right` or using the `actions` class

app/assets/javascripts/cable/subscriptions/notifications_channel.js.coffee
```diff
@@ -0,0 +1,15 @@
...
+  rejected: ->
+    console.log('Error subscribing to NotificationsChannel.')
+
+  received: (data)->
+    $('[data-behavior~=notifications-dot]').removeClass('hidden')
+
+    $container = $('[data-behavior~=notifications-dropdown] + div')
+    $unreadCount = $('[data-behavior~=unread-count]')
+    if $container.is(':visible')
+      $('.notification-header').after(data.notification_html)
```
dormi: we may get to this element from `data-behavior~=notifications-dropdown` too ?
No need to add an extra `data-behavior`, but the `notifications-dropdown` may be enough to cover all this references

app/models/concerns/notifiable.rb
```diff
@@ -8,13 +8,24 @@ module Notifiable
...
         )
+
+        broadcast_to_user(notification, recipient)
       end
     end
   end
+
+  def broadcast_to_user(notification, user)
+    project = self.commentable.project
+    notification_html = NotificationsController.render(
```
dormi: could this `render` call be simplified? `render(notification)` ?
May be we can get the project from the notifications itself, not need to pass `notification_project`

georgemillo: More importantly, why are we calling controller methods, rendering HTML and broadcasting to websockets from within a *model*? These are clearly things that belong in the controller layer.

dormi: I don't think this is that terrible, but option would be to not render anything and with js force a click to trigger the dropdown reload via ajax

Gemfile
```diff
@@ -108,7 +108,8 @@ gem 'sqlite3'#,  '1.3.10'
 
 # Use Unicorn as the web server
 # FIXME: Switch to Puma for Rails 5
```
georgemillo: I think this comment can be removed now

app/assets/stylesheets/snowcrash/modules/notifications.scss
```diff
@@ -37,7 +36,7 @@
       li.notification {
         position: relative;
```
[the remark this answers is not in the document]
georgemillo: It might not be directly relevant to your comment, but don't forget that the "notification" CSS class is also used (or at least will be used) by the `notifications#index` page. (See #315)

Gemfile
georgemillo: This commented-out part of the Gemfile can also be removed:

https://github.com/dradis/dradis-ce/blob/874a6492a1294fdd9e6439c9f35209aa0fa887f5/Gemfile#L11-L13

app/controllers/concerns/notifications_reader.rb
```diff
@@ -0,0 +1,13 @@
+module NotificationsReader
+  protected
+
+  # Mark each notifications associated with the item as read
+  def read_item_notifications(item)
+    item.comments.each do |comment|
+      comment.notifications.where(recipient: current_user, read_at: nil).each(&:read!)
```
georgemillo: This generates one sql UPDATE call for every notification, which could potentially be a lot:

screen shot 2018-07-24 at 13 07 33

You can reduce this to one SQL call by using `update_all`, but this has the disadvantage of not running any ActiveRecord validations/callbacks etc.

At the very least I think this `each` block should be wrapped in a transaction.

georgemillo: Oh and `comment.notifications.where(recipient: current_user, read_at: nil)` can be shortened to `comment.notifications.unread.where(recipient: current_user)

spec/factories/notifications.rb
```diff
@@ -3,6 +3,7 @@
     notifiable { |notification| notification.association :issue }
     association :actor, factory: :user
     association :recipient, factory: :user
+    read_at :nil
```
georgemillo: Watch out - this says `:nil` (i.e. a symbol), not `nil`.

But `read_at nil` is redundant anyway - you can leave the entire line out and FactoryBot will set it to `nil` by default

georgemillo:
I don't think the "mark as read" button should be visible once the notification has already been marked as read:

screen shot 2018-07-24 at 14 44 51

It's confusing to have a button that doesn't do anything when clicked.

aapomm:
@georgemillo I think it's to simplify the CE deployment. I'm not too familiar with the CE production environment to say if this is acceptable. cc @etdsoft @dormi

georgemillo:
We already use redis in CE though, or at least it's [included in the Gemfile](https://github.com/dradis/dradis-ce/blob/647d87b4f0b6fe946fcce906b5aa9f8dd65c0f8b/Gemfile#L34). (That's on `master`.) Is it not actually being used already?

On general principle I'd prefer using the same adapter in CE and Pro, in all environments, for the sake of consistency and dev/prod parity. But if introducing `redis` as a dependency (if it's not already a dependency??) is too much overhead, maybe it's not worth it.

I guess the real question is "why can't the `async` adapter be used in production?" I imagine that it's not built to handle large amounts of concurrent requests at scale. But maybe this doesn't matter for an app like Dradis, since it's meant to be deployed locally at a relatively small scale rather than there being a single instance online which zillions of people access at once. I don't know.

dormi:
We are also using `sqlite` in CE, which doesn't seem production ready.
The point would be to have a system that is easy to set up, so contributors have an easier entry path.

In the installable packages we built (like travelling ruby...), we may decide to change those tools.

Also, I think we are not currently using `redis` anywhere in CE.

georgemillo:
> Also, I think we are not currently using redis anywhere in CE.

I think you're right. The only other obvious place that might use redis is BG jobs, but we're using the async adapter for ActiveJob even in production:

```
$ RAILS_ENV=production rails runner "puts ApplicationJob.queue_adapter.class"
ActiveJob::QueueAdapters::AsyncAdapter
```

georgemillo:
(I think if we're not using `redis` in CE we should remove it from the Gemfile, or at least put a comment explaining it's only used in Pro, to make sure no-one else gets confused by this point in future.)

georgemillo:
I'm still getting the "DB locked" exception that Xavi talked about above. It happens quite often when I try to submit a comment (although this might not be the only thing that would trigger it.)

In Xavi's screenshot above, the error is triggered by the line `@issue = Issue.find(params[:id])`, but I've also seen it triggered by `notifications.update_all(read_at: Time.now)` in `notifications_reader.rb`.

I imagine that the lock is being caused by the BG job, which doesn't always finish in time to release the lock before the main app redirects to `issues#show` and tries to access the DB again?

The real problem is that we're trying to use SQLite for more than it's built for. I imagine if we switched to mysql or whatever these problems would disappear, which isn't a difficult change to make EXCEPT we can't reasonably expect people to migrate their own data to a different DB. [The answers to this StackOverflow question](https://stackoverflow.com/questions/7154664/ruby-sqlite3busyexception-database-is-locked?noredirect=1&lq=1) suggest that increasing the `timeout` config value in `database.yml` would be one workaround, although it feels very hacky to me (and I worry it wouldn't catch every case which might raise a `BusyException`. Plus this problem is only going to magnify in future as we add more types of BG job.)

@etdsoft what are your thoughts on this?

app/controllers/comments_controller.rb
```diff
@@ -10,6 +10,7 @@ def create
     @comment.user = current_user
     if @comment.save
       track_created(@comment)
+      broadcast_comment
```
dormi: I have doubts about this:
We create notifications on a bg job when calling `track_created`.
Then in the next line we call `broadcast_comment` which calls `@comment.notifications`.
Those `notifications` may still not exist, depending on the bg job.

dormi: I think the first one, rendering from outside the controller, at least should work.

I also think that if we have moved creating activities and notifications to a bg job, it also makes sense to send those notifications from the bg job.

app/views/notifications/_notification.html.erb
```diff
@@ -15,13 +15,15 @@
...
       </div>
     </div>
-    <%= link_to project_notification_path(@project || notification_project, notification),
-                title: 'Mark as read',
-                class: 'notification-read',
-                data: { behavior: 'notification-read' },
-                remote: true,
-                method: :put do %>
-      <i class="fa fa-times-circle-o fa-lg" aria-hidden="true"></i>
+    <% if !notification.read? %>
```
dormi: we may use https://api.rubyonrails.org/v5.2/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to_if, but...

app/jobs/activity_tracking_job.rb
```diff
@@ -20,6 +20,27 @@ def perform(action:, trackable_id:, trackable_type:, user_id:)
...
+    if trackable.respond_to?(:notify)
+      trackable.notify(action)
+      broadcast_notifications(trackable)
+    end
+  end
+
+  private
+
+  def broadcast_notifications(trackable)
+    project = trackable.commentable.project
```
dormi: not all trackables have a commentable ?
If in the future we want to create notifications when an issue is created, trackable would be an issue but an issue does not have a commentable

app/controllers/notifications_controller.rb
```diff
@@ -9,4 +9,38 @@ def index
...
+
+  def update
+    @notifications =
+      if params[:id] == 'all'
+        current_user.notifications
+      elsif params[:id]
+        current_user.notifications.where(id: params[:id])
+      end
+
+    @notifications.each(&:read!)
```
etdsoft: If we have 10 notifications this is going to cause 10 separate `UPDATE` calls. Is that what we want?

app/models/concerns/notifiable.rb
```diff
@@ -8,7 +8,7 @@ module Notifiable
   def create_notifications(action:, recipients:)
     ActiveRecord::Base.transaction do
       recipients.each do |recipient|
-        Notification.create(
+        notification = Notification.create(
```
etdsoft: why is this change needed?

app/views/notifications/_feed.html.erb
```diff
@@ -1,21 +1,3 @@
 <% notifications.each do |notification| %>
```
etdsoft: do we need the `_feed` partial? You can pass a notifications Array to `render` and rails will find the right partial.

app/controllers/issues_controller.rb
```diff
@@ -37,6 +38,8 @@ def show
                       user: current_user,
                       subscribable_type: @issue.class.to_s,
                       subscribable_id: @issue.id)
+
+    read_item_notifications(@issue)
```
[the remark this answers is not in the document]
aapomm: ✅

PR #309, opened 2018-07-18
Mentions

dormi:
### Spec
This task can be divided in 2:

- when creating notifications for a "comment created" activity, we must detect if the content of that comment has "@mentions". If it does we must:
   - subscribe the mentioned users to the comment commentable
   - create mentioned notifications for them, instead of created notifications
- on the front end, it would be nice to show a list of users to mention when clicking @ on the comment input

**Proposed solution**
For the UX part, make a list of available tools for this job.
The size and number of dependencies may help us decide the one.

### How to test
Make a comment on a issue, mentioning another user not subscribed to the issue.
Assert the user is subscribed and a "mention notification" created for him.

mention


spec/models/comment_spec.rb
```diff
@@ -15,4 +15,14 @@
       Comment.create(commentable: issue, content: 'rspec content', user: user)
     end.to change { Subscription.count }.by(1)
   end
+
+  describe '#mentions' do
```
georgemillo: Why does this need to be a public method? I don't see it being used anywhere except `subscribe_mentions` which itself is private?

dormi: But wait, this feels to hacky, and I think it makes sense being able to call `comment.mentions`.
Reverted my last change and only made `subscribe_mentions` private

georgemillo:
Kinda weird behaviour, but I suspect it's a problem with the `tribute` library rather than anything in your PR: if I select a name from the dropdown, but continue typing without hitting "Enter", the dropdown stays visible... then when I do hit "Enter", I lose everything I've typed since:

mentions

Really the dropdown should disappear if you hit Space, that's how GitHub does it for example:

github-mentions

georgemillo:
What's the point of showing the avatar in the dropdown list when everyone in CE has the same avatar anyway?

screen shot 2018-07-19 at 10 37 31

It's probably not worth removing the avatar from here just so there's one less difference to maintain between CE and Pro, but I figure I might as well mention it here in case anyone disagrees.

dormi:
@georgemillo I think I've been able to fix this white space problem

georgemillo:
Mentioning another user doesn't seem to subscribe them? Am I missing something?

subscribers

(In that gif I'm logged in as "alice", the 1 existing subscriber is alice, when I mention @bob in a comment no new subscription is created.)

bob:
Yes, I'm subscribed. Thanks a lot!

app/models/comment.rb
```diff
@@ -38,8 +38,34 @@ def create_subscription
...
+        users << mentioned_user
+      end
+    end
+
+    users
+  end
+
+  private
+  
+  def subscribe_mentions
```
etdsoft: name? `subscribe_mentioned`

app/models/comment.rb
```diff
@@ -38,8 +38,34 @@ def create_subscription
...
+    subscribers_by_mention = []
+    mentions.each do |mention|
+      Subscription.subscribe(user: mention, to: commentable)
+      subscribers_by_mention << mention
+    end
+
+    subscribers_by_mention
```
etdsoft: there is something weird here, `subscribers_by_mention` is just a duplicate of the return value of `#mentions`.

Why not avoid this and have `create` L40 first subscribe and then call `#mentions` again to pass that to `create_notifications`. Should `#mentions` be memoized? Possibly dependent on the value of `:content` having changed?

spec/jobs/activity_tracking_job_spec.rb
```diff
@@ -62,5 +62,30 @@
         )
       }.to change { Notification.count }.by(2)
     end
+
+    it 'creates notifications when a comment has mentions' do
```
etdsoft: this doesn't belong here, the reason I know is because this PR doesn't touch ActivityTrackingJob.

This code is testing the implementation of `Comments#notify` and it should belong in the corresponding spec.

app/views/comments/_mentionable_users.html.erb
```diff
@@ -0,0 +1,12 @@
+<%# this is picked up by javascripts/snowcrash/modules/comments.js.coffee %>
+<%= tag.div id: "mentionable-users",
```
georgemillo: can this be cached? Otherwise we're making the same DB call to load all Users on every single page load.

georgemillo:
Just gave this another review - other than my comment above about caching, it all seems fine.

georgemillo:
Seems good now 👍

etdsoft:
This is now merged. Thanks to @bob for your support!

PR #315, opened 2018-07-19
Notifications index page

georgemillo:
Some points of note:

- On this new page you can mark notifications as read. Users should also be able to do the same thing from the notifications *navbar*, which @aapomm has already started working on in #308. #308 is still unfinished and contains a lot of stuff that isn't relevant here, so rather than trying to merge in #308 (which will surely cause a huge mess), I've just duplicated the relevant pieces of that PR here. It's not a lot of code - just two new routes and controller actions - so hopefully this won't cause too much trouble when merging #308, but it's worth bearing in mind.
- Marking notifications as "read" from the index page triggers a full page reload. This isn't great UI - AJAX would be better - but since Aaron is already building an AJAX interface for the dropdown I figure it's best to keep things simple for now in this PR, then once the dropdown stuff is all finished we can think about re-using that code to make the index page better. 
- This PR introduces a few new references to `@project`, all of which are marked with `TODO` comments because they should be replaced with `current_project` once `project-id-scopes` is merged.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -2,11 +2,16 @@ class NotificationsController < AuthenticatedController
   include ProjectScoped
 
   def index
-    @notifications = current_user.notifications.newest.limit(20)
-    @unread_count = @notifications.unread.count
-
     respond_to do |format|
-      format.js
+      format.html do
+        @notifications = current_user.notifications.newest.page(params[:page])
+        @unread_count  = @notifications.unread.count
+      end
+      format.js do
+        @notifications = current_user.notifications.newest.limit(20)
+        # TODO how can we avoid this repetition of '@unread_count = ...'?
+        @unread_count  = @notifications.unread.count
+      end
     end
   end
 end
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -7,12 +7,12 @@ class NotificationsController < AuthenticatedController
     respond_to do |format|
       format.html do
         @notifications = notifs.page(params[:page])
-        @unread_count  = @notifications.unread.count
       end
       format.js do
         @notifications = notifs.limit(20)
-        # TODO how can we avoid this repetition of '@unread_count = ...'?
-        @unread_count  = @notifications.unread.count
+        # NB the unread count is not the same as @notifications.count because
+        # @notifications a) includes read notifs and b) is capped at 20
+        @unread_count  = current_user.notifications.unread.count
       end
     end
   end
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -2,13 +2,15 @@ class NotificationsController < AuthenticatedController
   include ProjectScoped
 
   def index
+    # TODO this doesn't fix all the N+1 query problems, needs a closer look:
+    notifs = current_user.notifications.newest.includes(:actor, notifiable: [:user, :commentable])
     respond_to do |format|
       format.html do
-        @notifications = current_user.notifications.newest.page(params[:page])
+        @notifications = notifs.page(params[:page])
         @unread_count  = @notifications.unread.count
       end
       format.js do
-        @notifications = current_user.notifications.newest.limit(20)
+        @notifications = notifs.limit(20)
         # TODO how can we avoid this repetition of '@unread_count = ...'?
         @unread_count  = @notifications.unread.count
       end
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -16,4 +16,23 @@ class NotificationsController < AuthenticatedController
       end
     end
   end
+
+  def read
+    current_user.notifications.find(params[:id]).mark_as_read
+    @has_unread = current_user.notifications.unread.any?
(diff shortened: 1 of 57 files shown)


app/controllers/notifications_controller.rb
```diff
@@ -2,11 +2,37 @@ class NotificationsController < AuthenticatedController
...
+      format.js do
+        @notifications = notifs.limit(20)
+        # NB the unread count is not the same as @notifications.count because
+        # @notifications a) includes read notifs and b) is capped at 20
+        @unread_count  = current_user.notifications.unread.count
+      end
+    end
+  end
+
+  def read
```
dormi: would it be possible to use `update` method here?
passing the `id` or a `all` as a parameter.

If we ever want to mix api and web controllers (not sure about it), it looks more restful.

app/models/notification.rb
```diff
@@ -23,4 +23,9 @@ class Notification < ApplicationRecord
   def read?
     self.read_at
   end
+
+  def mark_as_read(time = Time.now)
```
dormi: what do you think of calling this method `read!`?

app/controllers/notifications_controller.rb
```diff
@@ -2,11 +2,37 @@ class NotificationsController < AuthenticatedController
   include ProjectScoped
 
   def index
-    @notifications = current_user.notifications.newest.limit(20)
-    @unread_count = @notifications.unread.count
-
+    # TODO this doesn't fix all the N+1 query problems, needs a closer look:
```
dormi: What needs to be improved?

app/models/notification.rb
georgemillo: I prefer this name as well - it fits better with ruby conventions.

app/controllers/notifications_controller.rb
georgemillo: If you look in the server output it looks like some N+1 queries are going on, I couldn't figure out an immediate fix so decided to leave it until later - but forgot to come back to it! Will try and fix now.

app/models/notification.rb
georgemillo: Just note that #308 makes heavy use of this method and it's not merged yet - @aapomm , do you support this change?

georgemillo:
Regarding whether to use `update` instead of `read`, I'll leave that with @aapomm because like I mentioned above he's the one who wrote that code originally and I just C&Ped it from #308. So if Aaron wants to change it, I'll let him fix it on 308 first then I'll copy the changes into here.

dormi:
captura de pantalla 2018-07-20 a las 11 44 00
That dot in front of "You have no..." could be removed

app/controllers/notifications_controller.rb
```diff
@@ -2,11 +2,37 @@ class NotificationsController < AuthenticatedController
...
+    notifs = current_user.notifications.newest.includes(:actor, notifiable: [:user, :commentable])
```
dormi: I would call this variable `notifications`, or `_notifications`.

dormi:
I think that here: https://github.com/dradis/dradis-ce/blob/notifications-index-page/app/presenters/notification_presenter.rb#L67
we have already merged a bug.
This should be the `notification.actor`, not the `.recipient`.
I see: 
captura de pantalla 2018-07-20 a las 12 00 59
But the use that made those comments is not `xavi`

georgemillo:
Yes, good catch.

(Note to self: the same bug appears on [L52](https://github.com/dradis/dradis-ce/blob/notifications-index-page/app/presenters/notification_presenter.rb#L52) of the same file.)

georgemillo:
@dormi @aapomm this is ready for another review

spec/features/notifications/index_page_spec.rb
```diff
@@ -40,7 +40,7 @@ def create_notification(opts = {})
     expect(page).not_to have_content 'You have no notifications yet'
   end
 
-  example 'marking all notifications as read' do
+  example 'marking all notifications as read', :js do
```
dormi: I didn't know that `js` could be enabled like this in specs.
In other specs I see `js: true`.

dormi: let's be consistent

spec/features/notifications/index_page_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    create_notification(recipient: u1)
+    create_notification(recipient: u2)
+
+    visit project_notifications_path(@project)
+
+    expect(page).to have_selector '.notification', count: 2
+    expect(page).not_to have_content 'You have no notifications yet'
+  end
+
+  example 'marking all notifications as read', :js do
```
dormi: I didn't know that spces could be marked as `js` like this.
I would be consistent with other specs with: `js: true`

app/views/notifications/_dropdown.html.erb
```diff
@@ -18,8 +18,9 @@
     <% end %>
   </ul>
   <% if notifications.any? %>
-    <div class="footer text-center">
-      <a href="#">See all notifications</a>
+    <div class="footer text-center" data-id="js-footer">
```
dormi: we don't need this `data-id` ?

PR #317, opened 2018-07-20
Notifications trigger rest

dormi:
This is a PR to use more restful methods in `NotificationsController`.

I am not 100% convinced about using the `update` method for both a member call (with `id`) and what would be a collection call (`id = 'all'`).

Also in this PR we are trying to use remote: true links + `js.erb` views. This part looks interesting to me.


aapomm:
@dormi This works 👍. 
Conflict notes in the `notifications-trigger` branch:

- Added a `notification_project` local variable to the `_feed` partial
- Renamed `_notifications` to `_dropdown`

Also, do you think we should update the notification count when "reading" the notifications in the dropdown?

PR #318, opened 2018-07-23
scope 'find' calls to current_project

georgemillo:
While working on removing `set_project_scope` from Pro, I uncovered some more cases where `find` calls are not correctly scoped to the `current_project`. I've fixed them in Pro and pushed to `project-id-scopes` - this PR backports the changes to CE to keep the controller code identical.

Probably best not to merge this one until the `set_project_scope` stuff is done on Pro, in case I find any more changes like this that need to be backported. No sense doing it in more than one PR.


georgemillo:
Okay, _now_ this should be ready.

PR #319, opened 2018-07-23
csrf error

dormi:
We do these 2 checks on our requests:

- check csrf token (applies to POST requests).
- check for valid user session

When the user session expires, csrf token check also fails, because there is no info to check the incoming csrf token.

So, if we open Dradis, go to a page that sends a POST request (create issue, evidence...) and wait enough time (or delete the cookie manually), we see this error.

In this PR we force checking for a valid session before checking the csrf token. 

### How to test:
-  Create an issue and stay on that issue page
-  Delete the session cookie from the browser to simulate a session timeout
-  From the issue page/evidence tab, create evidence using the "Add evidence" link
-  We should be logged out, no error should raise

### DIFF
--- a/app/controllers/authenticated_controller.rb
+++ b/app/controllers/authenticated_controller.rb
@@ -7,7 +7,7 @@
 # Any third-party code (e.g. plugins) should inherit from this class instead of
 # calling the authentication filters directly
 class AuthenticatedController < ApplicationController
-  before_action :login_required
+  prepend_before_action :login_required
   before_action :set_paper_trail_whodunnit
   # before_action :render_onboarding_tour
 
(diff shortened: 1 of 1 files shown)


app/controllers/authenticated_controller.rb
```diff
@@ -7,7 +7,7 @@
 # Any third-party code (e.g. plugins) should inherit from this class instead of
 # calling the authentication filters directly
 class AuthenticatedController < ApplicationController
-  before_action :login_required
+  prepend_before_action :login_required
```
etdsoft: Do you think that it is clear why we're using `prepend_` here? I'm concerned about someone reverting this in the future.

PR #328, opened 2018-08-02
Add new projects teaser page

etdsoft:
If the URL :project_id parameter is changed to anything other than `1` show a teaser link.


tabathad:
Review in basecamp https://3.basecamp.com/3336503/buckets/4682975/todos/1218817942

PR #336, opened 2018-08-08
Release 4.0

aapomm:


georgemillo:
Note: before releasing the next CE version, we need to make sure we update the RedCloth dependency to 4.3.2, as 4.3.1 has a security vulnerability.

The problem for now is that the most recent version of `dradis-html_export` depends on RedCloth 4.3.1 exactly, meaning we can't update the dependency in CE without updating the dradis-html_export dependency to. So what we need to do is:

- Merge https://github.com/dradis/dradis-html_export/pull/13, which contains the RedCloth update (as well as a version bump)
- Publish the new version of dradis-html_export to Rubygems
- Update CE so it depends on RedCloth `~> 4.3.2`, *and* the new version of dradis-html-export

georgemillo:
Actually while I'm at it I might as well fix the version number and the branch name. Unfortunately I don't think it's possible to rename the branch without closing and opening a new PR. So moving to https://github.com/dradis/dradis-ce/pull/354

PR #346, opened 2018-08-20
Fix 2 bugs with the UI on issues#index

georgemillo:
Bug 1: if you update an issue's tag while the "tags" column is hidden, bad things happen:

issue-tag

Fix: update the JS so it hides the updated "tag" cell if necessary.

Bug 2: updating an issue's tag doesn't change the tag colour in the sidebar. This is caused by an incorrect jQuery selector string and is a one-line fix.

To test:
- Open a project, create some issues, and go to the "all issues" page
- Hide the "Tags" column with the button at the top
- Select an issue, click the Tag button that appears at the top, and change the tag
- Confirm that the table itself is not changed, but the tag icon next to the issue in the sidebar updates to the new colour.
- Without reloading the page, show the "Tags" column again and confirm that the updated tag is displayed.

### DIFF
--- a/app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
+++ b/app/assets/javascripts/snowcrash/modules/issues/table.js.coffee
@@ -8,13 +8,15 @@ class IssuesTable extends ItemsTable
     that = this
     $target = $(event.target)
     tagColumnIndex = @columnIndices['tags']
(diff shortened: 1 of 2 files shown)


georgemillo:
Crap, I accidentally merged this into `master` instead of `release-3.10`. Will open a new PR

PR #351, opened 2018-08-23
Add current_user to issues cache key

aapomm:
Currently, the show page does not update when logging with a new user due to caching. We want to re-cache the view when a new `current_user` is set.


georgemillo:
I've switched this PR's base branch to `release-4.0`, but it required a rebase and a force push: please make sure you pull the updated history before testing/updating.

PR #364, opened 2018-09-19
Escape html in comments

dormi:


aapomm:
Looks good to me 👍

app/views/comments/_comment.html.erb
```diff
@@ -30,7 +30,7 @@
     </div>
 
     <div class="content">
-      <%= simple_format(comment.content) %>
+      <%= simple_format(h(comment.content)) %>
```
etdsoft: according to [the docs](url
https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-simple_format), `simple_format` already does this.

PR #365, opened 2018-09-24
Add comments to notes

dormi:


app/helpers/comments_helper.rb
```diff
@@ -0,0 +1,27 @@
+module CommentsHelper
+  def comment_path(comment)
+    polymorphic_path(
+      commentable_resource_array(comment.commentable),
+      anchor: dom_id(comment)
+    )
+  end
+
+  def commentable_path(comment)
```
etdsoft: looks like these to methods could be merged if you make `anchor` an optional argument?

etdsoft: I'm not a fan of the name though, it makes it look like a Rails route, but it isn't.

etdsoft:
My main concern is around the hiding of the polymorphic route helper, in a way that's different to what we've done for Activities.

app/controllers/comments_controller.rb
```diff
@@ -12,28 +12,36 @@ def create
...
 
   private
 
   def comment_params
     params.require(:comment).permit(:content, :commentable_type, :commentable_id)
   end
+
+  def redirect_to_comment(comment)
+    if comment.persisted? && !request.referrer.nil?
+      request.env['HTTP_REFERER'] += "##{dom_id(comment)}"
```
dormi: I want to use an `anchor` with `redirect_back`, and I came up with this.
Better ideas?

dormi:
> My main concern is around the hiding of the polymorphic route helper, in a way that's different to what we've done for Activities.

Now we have moved the methods that return the path to a comment to the presenters.
Still calling it `comment_path` but now it reads `presenter.comment_path` in the views.

dormi:
Maybe it is time to go transform comments_controller to ajax.
If what we have is acceptable we could note it down in case we can fit that in this cycle.

PR #366, opened 2018-09-24
Add subscriptions to notes

dormi:
### Spec
Since we're adding comments to notes, we also need to add subscriptions to notes.

### Proposed solution
Add subscriptions to notes:

- Add subscription to the model
- Add subscribe/unsubscribe buttons to the view

### How to test

1. Add a note in a project
2. Subscribe to the note
3. Assert that you are receiving notifications for the note when it is commented.
4. Unsubscribe to the note
5. Assert that you are not receiving notifications for the note


app/controllers/notes_controller.rb
```diff
@@ -32,6 +32,11 @@ def create
   # Retrieve a Note given its :id
   def show
     @activities = @note.activities.latest
+    @subscription = Subscription.find_by(
```
aapomm: @dormi We'll be seeing a lot of this `find_by` method called the same way. Do you think it's worth it to add an instance method to the `Subscribable` module that does something like this?
e.g.
```
def current_subscription(user)
  Subscription.find_by(
    user: user,
    subscribable_type: self.class.to_s,
    subscribable_id: self.id
  )
end
```

app/controllers/subscriptions_controller.rb
```diff
@@ -6,7 +6,7 @@ def create
     subscription.user = current_user
     subscription.save
 
-    redirect_to [current_project, subscription.subscribable], notice: 'Subscribed!'
+    redirect_to helpers.subscribable_path(subscription), notice: 'Subscribed!'
```
etdsoft: shouldn't this be a redirect_back?

app/controllers/subscriptions_controller.rb
```diff
@@ -17,7 +17,7 @@ def destroy
     )
     subscription.destroy
 
-    redirect_to [current_project, subscription.subscribable], notice: 'Unsubscribed!'
+    redirect_to helpers.subscribable_path(subscription), notice: 'Unsubscribed!'
```
etdsoft: same

app/models/concerns/subscribable.rb
```diff
@@ -12,4 +12,12 @@ def create_subscription
       Subscription.subscribe(user: user, to: self)
     end
   end
+
+  def subscription_for(user:)
```
etdsoft: 👍

app/models/note.rb
```diff
@@ -25,7 +25,7 @@
 #
 # This behaviour is extensively used by import/export plugins such as WordExport.
 class Note < ApplicationRecord
-  include Commentable, HasFields, RevisionTracking
+  include Commentable, HasFields, RevisionTracking, Subscribable
```
etdsoft: same comment re: convention

app/views/issues/show.html.erb
```diff
@@ -40,34 +40,7 @@
...
-                        subscribable_type: @issue.class.to_s,
-                        subscribable_id: @issue.id
-                      }
-                    ),
-                    class: 'action-link',
-                    method: :post do %>
-                <i class="fa fa-bell"></i> Subscribe
-              <% end %>
-            <% end %> -
+            <%= render 'subscriptions/button', subscribable: @issue %> -
```
etdsoft: possibly `actions` is a better name?

app/models/concerns/subscribable.rb
dormi: ^ @aapomm's idea 👍

app/controllers/subscriptions_controller.rb
```diff
@@ -6,7 +6,7 @@ def create
     subscription.user = current_user
     subscription.save
 
-    redirect_to [current_project, subscription.subscribable], notice: 'Subscribed!'
+    redirect_back fallback_location: root_path, notice: 'Subscribed!'
```
etdsoft: at least redirect to the `current_project` with an eye on Pro sync

PR #367, opened 2018-09-24
Add mentions/notifications to notes

dormi:
### Spec
Since we're adding comments to the view, should also add mentions.

### Proposed solution
Add mentions to the notes with the mentionable users.

### How to test
1. Create a note in a project.
2. With user_1,  comment on the note with a mention to user_2
3. Check that user_2 gets a notification for that comment


aapomm:
@dormi looks good to me 👍. I think we can merge this.

app/controllers/notes_controller.rb
```diff
@@ -38,6 +38,7 @@ def show
       subscribable_id: @note.id
     )
     load_conflicting_revisions(@note)
+    @mentionable_users = User.all
```
etdsoft: time to extract a Concern?

app/views/notifications/_comment.html.erb
```diff
@@ -1,5 +1,5 @@
 <% commentable = comment.commentable %>
-<% comment_path = polymorphic_path([commentable.project, commentable], anchor: dom_id(comment)) %>
+<% comment_path = comment_path(comment) %>
```
etdsoft: don't we need the project or the :anchor any more?

spec/jobs/activity_tracking_job_spec.rb
```diff
@@ -60,20 +60,23 @@
     end
 
     it 'creates notifications when a comment is created' do
-      commentable = create(:issue)
-      create_list(:subscription, 2, subscribable: commentable)
-      trackable = create(:comment, commentable: commentable)
-      project = commentable.node.project
+      commentables = [:issue, :note]
```
etdsoft: I'm not sure about this code, the specs seems to be trying to do too much?

Maybe split in different scopes, or into a shared example or something else?

Otherwise by the end of the cycle we're going to have a mammoth spec for :issue, :evidence, :note, :card, etc.

app/views/notifications/_comment.html.erb
aapomm: This is using the method from `CommentsHelper` and not the Rails one.

dormi:
I forgot something that I think belongs to this PR too: https://github.com/dradis/dradis-ce/pull/367/commits/ad223097d7ef1f2e4151a8bb26a5fe126c60d818

app/views/notifications/_comment.html.erb
dormi: This is added in https://github.com/dradis/dradis-ce/pull/365

app/controllers/concerns/mentions.rb
```diff
@@ -0,0 +1,9 @@
+module Mentions
+  extend ActiveSupport::Concern
+
+  protected
+
+  def find_mentionable_users
```
etdsoft: can we make this a before_filter here for the relevant actions (e.g. `#show`)?

That way we skip cluttering the controller code.

app/controllers/concerns/mentions.rb
```diff
@@ -0,0 +1,9 @@
+module Mentions
```
etdsoft: Not convinced about the name convention of our controller concerns, it's all over the place:

- include ActivityTracking
- include ContentFromTemplate
  include ContentFromTemplate
  include ConflictResolver
  include ConflictResolver
  include Mentions
  include MultipleDestroy
  include MultipleDestroy
  include NotificationsReader
  include NotificationsReader
  include ProjectScoped
  include ProjectScoped

app/controllers/notes_controller.rb
```diff
@@ -3,7 +3,9 @@
 class NotesController < NestedNodeResourceController
   include ConflictResolver
   include MultipleDestroy
+  include Mentions
```
etdsoft: :alpha

app/controllers/notes_controller.rb
```diff
@@ -34,6 +36,8 @@ def show
     @activities = @note.activities.latest
     @subscription = @note.subscription_for(user: current_user)
     load_conflicting_revisions(@note)
+    find_mentionable_users
+    read_item_notifications(@note, current_user)
```
etdsoft: can this be, as part of a separate PR, pushed up to the concern?

app/views/notifications/_comment.html.erb
etdsoft: discussed there

spec/jobs/activity_tracking_job_spec.rb
```diff
@@ -59,22 +59,8 @@
...
-          action: 'create',
-          project_id: project.id,
-          trackable_id: trackable.id,
-          trackable_type: trackable.class.to_s,
-          user_id: trackable.user.id
-        )
-      }.to change { Notification.count }.by(2)
-    end
+    include_examples 'creates notifications for comments in a', :issue
+    include_examples 'creates notifications for comments in a', :note
```
etdsoft: as discussed in the call: this PR doesn't introduce any changes to the ActivityTracking code, so it's weird that we have to change its spec.

Another smell is that the responsibility for ensuring the behaviour of :note is put in the ActivityTracking spec, when it should be in some note-related spec. Otherwise this is only going to get uglier as we add more notifiable models.

dormi: In fact I think we are already testing this in the right place, the `comment` model: https://github.com/dradis/dradis-ce/blob/master/spec/models/comment_spec.rb#L40

app/controllers/concerns/mentions.rb
dormi: noted

app/controllers/notes_controller.rb
dormi: will do

dormi: https://github.com/dradis/dradis-ce/pull/377

app/controllers/issues_controller.rb
```diff
@@ -2,13 +2,15 @@ class IssuesController < AuthenticatedController
...
   include ConflictResolver
+  include Mentions
   include MultipleDestroy
   include NotificationsReader
   include ProjectScoped
 
   before_action :find_issuelib
   before_action :find_issues, except: [:destroy]
 
+  before_action :find_mentionable_users, only: [:show]
```
etdsoft: why not move this to the concern directly?

app/controllers/notes_controller.rb
```diff
@@ -2,9 +2,12 @@
 # resource.
 class NotesController < NestedNodeResourceController
   include ConflictResolver
+  include Mentions
   include MultipleDestroy
   include NodesSidebar
+  include NotificationsReader
 
+  before_action :find_mentionable_users, only: [:show]
```
etdsoft: ditto

PR #373, opened 2018-10-01
Evidence subscriptions

aapomm:
Reviewer notes: this PR has some commits from #366 and #372. Ideally those should be looked into first.

### Spec
Since we're adding comments to evidence, we should also add subscriptions.

**Proposed solution**
Add subscriptions to evidences:
  - Add subscription to the Evidence
  - Add subscribe/unsubscribe to the evidence view

### How to test

1. Add an evidence in a project
2. Subscribe to the evidence
3. Assert that you are receiving notifications for the evidence when it is commented.
4. Unsubscribe to the evidence
5. Assert that you are not receiving notifications for the evidence


app/models/concerns/subscribable.rb
```diff
@@ -12,4 +12,12 @@ def create_subscription
       Subscription.subscribe(user: user, to: self)
     end
   end
+
+  def current_subscription(user)
```
dormi: I want to use this concern method too.
But I don't see the `current` in the name `current_subscription` 🤔 
What about:
`def user_subscription(user)`
`def subscription_for(user: user)`
`def find_subsbcription(user)`
...

dormi: I vote for `subscription_for(user:)`, and I am open to suggestions, or to use `current_subscription`

aapomm: I'm going with `subscription_for`

dormi:
@aapomm I think the subscriptions feed is missing in `evidence#show`

dormi:
Also, I noticed that here https://github.com/dradis/dradis-ce/pull/373/files#diff-f251accfc96ca2940f035c06c490255aR23 we are missing comments and subscriptions as part of the cache, as we did in issues.

aapomm:
@dormi Added the feed in evidence and updated the cache keys for both notes and evidence.

app/controllers/evidence_controller.rb
```diff
@@ -10,6 +10,7 @@ class EvidenceController < NestedNodeResourceController
   def show
     @issue      = @evidence.issue
     @activities = @evidence.activities.latest
+    @subscription = @evidence.subscription_for(user: current_user)
```
etdsoft: alignment

etdsoft:
Minor comment, please merge when ready.

PR #377, opened 2018-10-05
Move before actions to concerns

dormi:
We have detected a couple of lines that can be moved to an already existent concern:

 - In IssuesController and NotesController, this line:
```
before_action :find_mentionable_users, only: [:show]
```
has been moved to the `Mentions` controller concern (renamed to `Mentioned` to scratch some consistency. )

- In the same controller, this line:
```
read_item_notifications(@issue, current_user)
```
has been moved to the `NotificationsReader` concern.

- And also in the same controllers, this other line:
```
@subscription = @issue.subscription_for(user: current_user)
```
has been moved to a new `Subscribed` controller concern.


app/controllers/issues_controller.rb
```diff
@@ -2,18 +2,19 @@ class IssuesController < AuthenticatedController
...
   include ProjectScoped
 
   before_action :find_issuelib
   before_action :find_issues, except: [:destroy]
 
-  before_action :find_mentionable_users, only: [:show]
   before_action :find_or_initialize_issue, except: [:import, :index]
   before_action :find_or_initialize_tags, except: [:destroy]
 
+  include Subscribed
```
etdsoft: really? down here? can't we insert the filter in the right spot? If not, we need a comment, otherwise someone refactoring this code will move this line with the other includes.

app/controllers/notes_controller.rb
```diff
@@ -2,15 +2,16 @@
...
+  include Mentioned
   include MultipleDestroy
   include NodesSidebar
   include NotificationsReader
 
-  before_action :find_mentionable_users, only: [:show]
   before_action :find_or_initialize_note, except: [:index, :new, :multiple_destroy]
   before_action :initialize_nodes_sidebar, only: [:edit, :new, :show]
 
+  include Subscribed
```
etdsoft: ditto

PR #381, opened 2018-10-10
Comments api

aapomm:
Reviewer notes: update the base branch when #372 is merged.

### Spec
We want to have an API endpoint for comments. This task should implement the comments#index API.

**Proposed solution**
We'll pattern the comment API endpoint from basecamp.

First, we add two new attributes to the the commentable items API: comments_count and comments_url. Comments URL will point to the commentable's comment API endpoint. Ex: api/issues/:issue_id/comments.

Second, we add the comments endpoint. Using the comments_url in the first step, it should list the commentable's comments in an array. 

For this task, we should implement the groundwork for the comments API and the comments index.

### How to test

1. Add an issue to the project.
2. Add comments to the issue.
3. Fetch the issue using the API. Assert that the comments_count and comments_url appears.
4. Fetch the issue's comments using the comments_url. Assert that the comments were correctly fetched.

### DIFF
--- a/app/models/concerns/commentable.rb
+++ b/app/models/concerns/commentable.rb
@@ -13,8 +13,4 @@ module Commentable
       )
     )
   end
-
-  def comments_count
-    self.comments.count
-  end
 end
--- a/app/models/concerns/commentable.rb
+++ b/app/models/concerns/commentable.rb
@@ -13,4 +13,8 @@ module Commentable
       )
     )
   end
+
+  def comments_count
+    self.comments.count
+  end
 end
--- a/engines/dradis-api/app/controllers/dradis/ce/api/v1/comments_controller.rb
+++ b/engines/dradis-api/app/controllers/dradis/ce/api/v1/comments_controller.rb
@@ -22,16 +22,10 @@ module Dradis::CE::API
       private
 
       def set_commentable
-        @commentable =
-          if params[:issue_id]
-           Issue.find(params[:issue_id])
-          elsif params[:note_id]
-            Note.find(params[:note_id])
-          elsif params[:evidence_id]
-            Evidence.find(params[:evidence_id])
-          else
-            raise 'Polymorphic model missing!'
-          end
+        commentable_klasses = %w[issue note evidence]
+        if klass = commentable_klasses.detect { |ck| params[:"#{ck}_id"].present? }
+          @commentable = klass.camelize.constantize.find params[:"#{klass}_id"]
+        end
       end
 
       def set_comment
--- a/engines/dradis-api/app/controllers/dradis/ce/api/v1/comments_controller.rb
+++ b/engines/dradis-api/app/controllers/dradis/ce/api/v1/comments_controller.rb
@@ -24,7 +24,8 @@ module Dradis::CE::API
       def set_commentable
         commentable_klasses = %w[issue note evidence]
         if klass = commentable_klasses.detect { |ck| params[:"#{ck}_id"].present? }
-          @commentable = klass.camelize.constantize.find params[:"#{klass}_id"]
+          @commentable =
+            current_project.send(klass.pluralize).find(params[:"#{klass}_id"])
         end
       end
 
--- a/engines/dradis-api/app/controllers/dradis/ce/api/v1/comments_controller.rb
+++ b/engines/dradis-api/app/controllers/dradis/ce/api/v1/comments_controller.rb
@@ -1,0 +1,36 @@
+module Dradis::CE::API
+  module V1
+    class CommentsController < Dradis::CE::API::V1::ProjectScopedController
+      before_action :set_comment, only: [:show, :update, :destroy]
+      before_action :set_commentable, only: [:index, :create]
+
+      def index
+      end
+
+      def show; end
+
+      def create
+      end
+
+      def update
+      end
+
+      def destroy
+      end
+
+      private
+
+      def set_commentable
+        regex = /(?<=\/api\/)([a-z]+)(?=\/)/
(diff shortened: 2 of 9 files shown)


engines/dradis-api/app/controllers/dradis/ce/api/v1/comments_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      def update
+      end
+
+      def destroy
+      end
+
+      private
+
+      def set_commentable
```
aapomm: This complexity is caused by the fact that we route the comments list to the same controller. ex: 

```
/api/evidence/:evidence_id/comments
/api/issues/:issue_id/comments
/api/notes/:note_id/comments
```
Let me know if there's an simpler way of doing this.

engines/dradis-api/config/routes.rb
```diff
@@ -1,8 +1,17 @@
...
+  concern :with_comments do
+    resources :comments, only: [:index, :create]
+  end
+
   scope path: :api do
     defaults format: 'json' do
       scope module: :v1, constraints: Dradis::CE::API::RoutingConstraints.new(version: 1, default: true) do
-        resources :issues
+        resources :issues, concerns: :with_comments
+        resources :evidence, only: [], concerns: :with_comments
```
aapomm: Another resource line is added for notes and evidence to prevent an unnecessarily long path. Ex:
```
/api/nodes/1/notes/1/comments
```
vs
```
/api/notes/1/comments
```

engines/dradis-api/app/views/dradis/ce/api/v1/evidence/_evidence.json.jbuilder
```diff
@@ -4,3 +4,5 @@ json.issue do |json|
   json.title evidence.issue.title
   json.url dradis_api.issue_url(evidence.issue_id)
 end
+json.comments_count evidence.comments_count
```
dormi: Use a partial for those 2 lines used in issues, evidence and notes.

I am thinking if the need of this partial could be automatically detected (something like `respond_to?(:comments)`) and then automatically added. Would be nice, but how? 🤔

engines/dradis-api/config/routes.rb
```diff
@@ -1,8 +1,17 @@
...
+        resources :notes, only: [], concerns: :with_comments
```
dormi: what about this `only: []` ?

app/models/concerns/commentable.rb
```diff
@@ -13,4 +13,8 @@ def commentable_activities
       )
     )
   end
+
+  def comments_count
+    self.comments.count
```
dormi: I would skip this method. (`comments_count` vs `comments.count`...)

engines/dradis-api/config/routes.rb
dormi: I am good with the long path version, I feel its is more consistent.
If I see `/api/notes/1/comments`, makes me think that I can access the note with: `/api/notes/1`, but that is not true.

engines/dradis-api/config/routes.rb
aapomm: We don't want to add a shallow routes for evidence and notes. ex: `/api/notes/1/`. Although I actually prefer this shallow path than the one with `nodes`, but this isn't the PR for that.

engines/dradis-api/config/routes.rb
aapomm: @dormi that would mean that when deleting comments, you'd need 3 id's (node, commentable, comment). Is that okay? Or should we have an `/api/comments` path?

engines/dradis-api/app/controllers/dradis/ce/api/v1/comments_controller.rb
```diff
@@ -0,0 +1,42 @@
...
```
dormi: I had this idea to make this less dependent on the commendable type:
https://github.com/dradis/dradis-ce/pull/385/files#diff-6cead7405362cf0e4214e6479ed22755R44

engines/dradis-api/config/routes.rb
```diff
@@ -1,11 +1,17 @@
 Dradis::CE::API::Engine::routes.draw do
```
dormi: I added some changes to this file in this PR that points to this same branch: https://github.com/dradis/dradis-ce/pull/385/files#diff-0f2e3d02b44405293aedb7d1ee196292

dormi:
specs fixed in this PR https://github.com/dradis/dradis-ce/pull/385

engines/dradis-api/app/controllers/dradis/ce/api/v1/comments_controller.rb
etdsoft: we're basing our design in Basecamp, and I believe this `if/elsif/elsif` would make DHH cringe.

I'm not sure if there is a better solution, but a quick search on StackOverflow yielded this:

https://stackoverflow.com/a/14959484

engines/dradis-api/config/routes.rb
```diff
@@ -1,11 +1,17 @@
...
+  concern :with_comments do
+    resources :comments, only: [:index, :create]
+  end
+
   scope path: :api do
     defaults format: 'json' do
       scope module: :v1, constraints: Dradis::CE::API::RoutingConstraints.new(version: 1, default: true) do
-        resources :issues
+        resources :issues, concerns: :with_comments
```
etdsoft: :alpha

brianp:

PR #383, opened 2018-10-16
Enhance mentions display

dormi:
When comments have mentions, we are just displaying the Email in the rendered comment, like: `@user@example.com`.

This PR tries to enhance that rendering.

With a view helper, when we are rendering a comment and detect a mention, we search for the user an instead of rendering the string above, we render some improved version of it using the user info (avatar, name with styles, etc...)

improved_mentions_ce


brianp:
We've added mentions enhancements. Which perform similar functionality. Closing this PR.

PR #384, opened 2018-10-18
Improve cache keys in commentable#show pages

dormi:
### Spec
The resources `comments`, `subscriptions` and the `current_user` can be removed from the cache key in `commentable#show` view

### Proposed solution

1. Separate comments and subscriptions feed to its own cache fragment.
2. Remove current user from cache keys and use js to make specific user content visible.

### How to test

- enable cache: `rake dev:cache`

- visit a note/evidence/issue with user A
- visit the same page with user B, check that the page renders from cache
   (you may insert `<%= Time.now %>` here: https://github.com/dradis/dradis-ce/pull/384/files#diff-9c0beaca63c1fc3597ad03d826b4f02eR30)

- change subscription status with user A. Reload with user B: user B must keep the subscriptions status, and page must still be cached.

- add a comment with user A
- reload page with user B. Check that comments for user B are loaded from cache
  (you may insert `<%= Time.now %>` here: https://github.com/dradis/dradis-ce/pull/384/files#diff-1aae7ed6c016b3e62cf25303b9a3f452R4)

- Add a comment for user B. Assert that each user can only see the edit/delete links for the comments that he/she posted.


app/views/comments/_feed.html.erb
```diff
@@ -1,16 +1,19 @@
...
 
-    <div data-notice="no-comments" class="comment"
-      <% if commentable.comments.any? %>
-        style="display: none;" <%# may be un-hidden later by JS %>
-      <% end %>
-    >
-      There have been no comments yet.
+      <div data-notice="no-comments" class="comment"
+        <% if commentable.comments.any? %>
+          style="display: none;" <%# may be un-hidden later by JS %>
```
aapomm: Can we use the `hidden` class here instead of adding a `style` attribute?

app/assets/stylesheets/snowcrash/modules/_subscriptions.scss
```diff
@@ -0,0 +1,3 @@
+.subscription-link {
```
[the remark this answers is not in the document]
dormi: 👍

app/views/comments/_feed.html.erb
dormi: 👍

etdsoft:
Each comment div needs to be cached as well (w/ `comment` as the key). That way if we have to re-create the entire feed for some reason, then most of the blocks will come from the cache.

app/controllers/concerns/commented.rb
```diff
@@ -0,0 +1,17 @@
+module Commented
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :set_comments, only: :show
```
etdsoft: I'm worried about L12. It seems that it'd be safer to let CanCan load the resource and hook after that? Although it seems we're not using cancan for load_and_authorize

Maybe the concern is more trouble than it's worth?

L12 makes the assumption that `:find_or_initialize_X` is available, which will break in Pro's CardsController.

Alternatively maybe the concern can skip the filter and just provide a `def comments` method that's also a `helper_method` available to views. That way when it's called in the views, we're reasonably certain that the `instance_variable_get("@#{resource}")` is going to work, w/o making any assumptions about how it was found or initialised.

app/controllers/notes_controller.rb
```diff
@@ -73,13 +74,14 @@ def destroy
...
   # are going to be working with based on the :id passed by the user.
   def find_or_initialize_note
-    if params[:id]
-      @note = @node.notes.find(params[:id])
-    elsif params[:note]
-      @note = @node.notes.new(note_params)
-    else
-      @note = @node.notes.new
-    end
+    @note ||=
```
etdsoft: `app/controllers//nested_node_resource_controller.rb` also provides a find_or_initialize method, it would be good it that's updated to the new `||=` convention you're following in this PR

app/views/comments/_comment.html.erb
```diff
@@ -13,20 +13,19 @@
...
-                      data: { confirm: 'Are you sure?' },
-                      class: 'text-error' do %>
-            <i class="fa fa-trash"></i> Delete
-          <% end %>
-        </div>
-      <% end %>
+      <% # this is not visible, it is shown later with js %>
+      <div class="actions">
+        <%= link_to 'javascript:void(0)',
+                    data: {toggle_comment: "on"} do %>
```
etdsoft:
```suggestion
                    data: { toggle_comment: 'on' } do %>
```

app/views/layouts/snowcrash.html.erb
```diff
@@ -11,6 +11,9 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=0, minimum-scale=1.0, maximum-scale=1.0">
   <meta name="apple-mobile-web-app-capable" content="yes">
   <meta name="apple-mobile-web-app-status-bar-style" content="black">
+  <% if current_user %>
```
etdsoft: Is this really needed? Can you ever be in SnowCrash layout w/o signing in first?

app/controllers/concerns/commented.rb
dormi: I like (and used) the concern that just adds helper methods idea.

app/controllers/notes_controller.rb
dormi: Old this memoization, now that `find_or_initialize_X` is no longer being called in the concern, is not necessary, because `find_or_initialize_X` is only going to be called once, as before.
Reverting these changes.

PR #386, opened 2018-10-19
Fix NotificationsReader

dormi:
### Spec
When we get a notification and we visit it clicking the notifications dropdown link, the notification is marked read, but the red dot alert stays, until we reload the page again.
￼
red_dot_stays

### Proposed solution
`NotificationsReader` conocern cannot run in the bg.
If we visit the notified resource, we mark it read at visit time, not later.

### How to test
  Even if this is a CE PR, but can only be seen when changing the bg queue to use redis.
  So change it or use a Pro instance if available.
  Then:
  - create a an issue with user A
  - visit the "All issues" the with user A
  - with user B comment on that issue
  - assert that user A get the red dot on the top bell wo/ doing anything
  - visit the issue with user A, assert the red dot disappears

### DIFF
--- a/app/assets/javascripts/cable/subscriptions/notifications_channel.js.coffee
+++ b/app/assets/javascripts/cable/subscriptions/notifications_channel.js.coffee
@@ -10,4 +10,7 @@ App.cable.subscriptions.create 'NotificationsChannel',
     $(document).off('.notifications')
 
   received: (data)->
-    $('[data-behavior~=notifications-dot]').removeClass('hidden')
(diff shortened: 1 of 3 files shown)


aapomm:
@dormi this is working, but I think the correct solution for this is to instead send out a websocket broadcast that updates the red dot, with the job still running in the background.

aapomm:
@dormi this is working, but I think the correct solution for this is to instead send out a websocket broadcast that updates the red dot, with the job still running in the background.

PR #387, opened 2018-10-25
Commentable path

dormi:
We noticed during the API implementation, that we may need to know the full url for a comment not only in our views.

Before we had an ugly method under our activity/notification presenter to handle that.

In this PR those methods are removed.

The `Commentable` model concern introduces a default `path` method that returns an array with the nested resources needed to build the url for that commentable.

Every commentable model may or not overwrite that default method.

The PR points to the API branch to keep the context of the discussion.

### DIFF
--- a/app/models/concerns/commentable.rb
+++ b/app/models/concerns/commentable.rb
@@ -13,4 +13,8 @@ module Commentable
       )
     )
   end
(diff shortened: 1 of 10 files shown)


brianp:

PR #396, opened 2018-11-08
Fix evidence redirects

dormi:
When we edit or delete evidence from an issue, we must stay in the issue (and not be redirected to the node).

This must be tested when the issue has 1 evidence for a single node, and > 1 evidence for a single node.


app/controllers/evidence_controller.rb
```diff
@@ -115,7 +115,13 @@ def destroy
       if @evidence.destroy
         track_destroyed(@evidence)
         format.html {
-          redirect_to [current_project, @node],
+          path =
```
etdsoft: Wouldn't a `redirect_back` call work in this context?

https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_back

dormi: This sounds like the way to go, thanks.

I tried it but I found the following problem:
Evidence can be deleted from 3 places:
1.  from the issue evidence tab
2. from the node evidence tab
3. from the evidence show page itself (under node)
When using `redirect_back` in case 3, we find an `evidence not found` error, since the evidence does not exist anymore.

So I reverted the change, using again the `back_to` param in delete links.

@aapomm do you see a more elegant way to deal with this?

etdsoft: I'd still use `redirect_back` (check the source) and set `request.headers["Referer"]` to `nil` if they match the `Evidence#show` URL for the recently deleted, that way it'll go to the fallback.

I'd even prefer an:

```
if request.headers["Referer"] == Evidence#show
  redirect_to node
else
  redirect_back
end
```

Than having to add (and remember to add in the future) the extra `:back_to` param

app/views/issues/nodes/_evidence.html.erb
```diff
@@ -33,7 +33,7 @@
               <h3>
                 Instance #<%= i+1 %> for <%= link_to node.label, project_node_path(node.project, node) %> -
                 <span class="actions">
-                  <%= link_to edit_project_node_evidence_path(node.project, node, instance), class: 'action-link' do %>
+                  <%= link_to edit_project_node_evidence_path(node.project, node, instance, back_to: :issue), class: 'action-link' do %>
```
etdsoft: it seems this change is no longer needed?

etdsoft:
```suggestion
                  <%= link_to edit_project_node_evidence_path(node.project, node, instance), class: 'action-link' do %>
```

app/controllers/evidence_controller.rb
```diff
@@ -115,8 +115,13 @@ def destroy
       if @evidence.destroy
         track_destroyed(@evidence)
         format.html {
-          redirect_to [current_project, @node],
-            notice: "Successfully deleted evidence for '#{@evidence.issue.title}.'"
+          notice = "Successfully deleted evidence for '#{@evidence.issue.title}.'"
+          fallback_location = [current_project, @node]
```
etdsoft:
```suggestion
          fallback_location = project_node_path(current_project, @node)
```

app/views/issues/nodes/_evidence.html.erb
dormi: this `back_to` param solves the evidence edit from issue case, when we have > 1 evidence for the issue.
When we have just 1 evidence the param was already there https://github.com/dradis/dradis-ce/pull/396/files#diff-05c2eb1fbc4727068bb8269ff00dfbe3R8

PR #397, opened 2018-11-12
Fix comments form cache

dormi:
### Spec
We found a bug on the way we are using cache for comment forms.
We are caching the form, even the csrf token.
So if we logout and login again the token is the cached one and the form does not work.

This is fix tries to invalidate the cache by session.
This leaves us closer to where we were before (we were invalidating the cache for different users), may be a bit worse sine on a new session the cached pages will not be there.

The real fix here will probably be moving the comments create and edit forms to Ajax.

### How to test
- enable cache
- visit issues#show (so the page is cached)
- logout + login
- visit the same page
- try to comment and assert that it works
- try also with evidence#show and notes#show

### DIFF
--- a/app/views/evidence/show.html.erb
+++ b/app/views/evidence/show.html.erb
@@ -20,7 +20,7 @@
 
 
 <div class="tab-content">
-  <% cache ['evidence-evidence-tab', @evidence, @evidence.comments, @evidence.subscriptions, current_user] do %>
+  <% cache ['evidence-evidence-tab', @evidence, @evidence.comments, @evidence.subscriptions, session.id] do %>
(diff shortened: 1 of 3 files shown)


etdsoft:
It seems to me that we're trying to cache too much in a single chunk. Intuitively the solution for this would be that the comment feed is left inside the cached chunk but the form is left outside.

Now in order to do that, instead of having a single cache chunk for the entire tab of the #show view, we could have different chunks: one for the content of the model itself, one for the comment feed, one for the activity feed, etc.  The overall performance shouldn't suffer as most of the chunks won't change at the same time. And we can leave any truly dynamic parts, like forms and such, out of any chunks.

PR #399, opened 2018-11-13
Unescape links in the escape filter

aapomm:
The Textile links are displaying incorrectly. Example text:
```
#[Sources]#
"OWASP Top 10: A6 - Sensitive data exposure":https://www.owasp.org/index.php/Top_10_2013-A6-Sensitive_Data_Exposure

"OWASP Top 10: A6 Sensitive data exposure":https://www.owasp.org/index.php/Top_10_2013-A6-Sensitive_Data_Exposure
```
Current display:
screen shot 2018-11-13 at 7 28 56 pm

It should display like:
screen shot 2018-11-13 at 7 29 47 pm

This is because the `DradisEscapeHTMLFilter` class is escaping the quotes to `&quot;` making it fail when parsed as Textile.

Note: this PR only fixes the display/preview and not how the links export.

### How to test:
1. Use the sample text above for an issue
2. Save the issue / click on Preview
3. Assert that the content is correctly displayed


lib/html/pipeline/dradis_escape_html_filter.rb
```diff
@@ -13,13 +13,19 @@ class Pipeline
...
+        callto feed svn urn aim rsync tag ssh sftp rtsp afs file
+      }
+
       def call
         text = ERB::Util.html_escape(@text)
 
-        # Match the text between bc. and bc.. following the textile rules
+        # Match the text under bc./bc.. and links, following the textile rules
         regex = Regexp.union(
           /(?<=bc. )(.*?)(?=(\r\n|\n){2})/m,
```
dormi: Now that we are here, I think this this should be: 
```
/(?<=bc\. )(.*?)(?=(\r\n|\n){2})/m
```
(using `\.` instead of just `.` after the `bc`)
Also on the next line for the `bc..`.

If not we are matching undesired cases:
http://rubular.com/r/WGCnnGTAr3

dormi:
Here https://github.com/dradis/dradis-ce/pull/399/files#diff-19869d50981411c48297e5e81956fcd0R33 we may now use a more descriptive variable name instead of `bc`. Maybe be `escaped` or `to_unescape`

lib/html/pipeline/dradis_escape_html_filter.rb
```diff
@@ -13,18 +13,24 @@ class Pipeline
     #
     # This filter does not write any additional information to the context hash.
     class DradisEscapeHTMLFilter < TextFilter
+      PROTOCOLS = %w{
```
etdsoft: let's no shoot ourselves in the foot here, `http` and `https` are more than enough for our use case. I'd also hard-code them in the regex in L28

lib/html/pipeline/dradis_escape_html_filter.rb
```diff
@@ -13,18 +13,24 @@ class Pipeline
...
         text = ERB::Util.html_escape(@text)
 
-        # Match the text between bc. and bc.. following the textile rules
+        # Match the text under bc./bc.. and links, following the textile rules
         regex = Regexp.union(
-          /(?<=bc. )(.*?)(?=(\r\n|\n){2})/m,
-          /(?<=bc.. )(.*?)(?=(bc\.|bc\.\.|p\.|\z))/m
+          /(?<=bc\. )(.*?)(?=(\r\n|\n){2})/m,
+          /(?<=bc\.\. )(.*?)(?=(bc\.|bc\.\.|p\.|\z))/m,
+          /&quot;.*&quot;:(?:#{PROTOCOLS.join('|')})\:\/\/.+/
```
etdsoft: - we probably need to use the `(.*?)` formula, no?
- don't we need the `/m`?
- see above re: hard-coding the supported protocols here

etdsoft:
Looks like `DradisEscapeHTMLFilter` needs a bunch of specs?

lib/html/pipeline/dradis_escape_html_filter.rb
aapomm: I don't think we'd need the `/m` since there can be no multiline links.

PR #400, opened 2018-11-13
Use Ajax in comments

dormi:
### Spec

We are using cache in our comments feed.
But as comments are implemented now, this means that we may be caching the comments edit and create forms we we are not very careful with cache keys: https://github.com/dradis/dradis-ce/pull/397 
In this PR we intend to remove forms from any cache fragment rendering them via ajax.

For consistency we are moving all comment CRUD actions to Ajax 

### How to test
Add, Edit and Delete one ore more comments.
Assert everything keeps working.

### DIFF
--- a/app/assets/javascripts/snowcrash/modules/comments.js.coffee
+++ b/app/assets/javascripts/snowcrash/modules/comments.js.coffee
@@ -1,18 +1,12 @@
 document.addEventListener "turbolinks:load", ->
 
-  # hide/show comment editable textarea
-  $("[data-toggle-comment]").click ->
+  # hide comment editable textarea when clicking cancel
+  $('.comment').on('click', "[data-behavior~=cancel-comment]", ->
     element = $(this)
     comment = element.closest(".comment")
-    content = comment.find(".content")
-    update  = comment.find(".edit_comment")
-    value   = element.data('toggle-comment')
-    if value == "on"
-      content.hide()
-      update.show()
-    else if value == "off"
-      content.show()
-      update.hide()
+    comment.find(".content").show()
+    comment.find("form").hide()
+  )
 
   if $('[data-behavior~=mentionable]').length
     # initialize mentions (https://github.com/zurb/tribute)
--- a/app/views/comments/_form.html.erb
+++ b/app/views/comments/_form.html.erb
@@ -1,6 +1,6 @@
-<%= form_for [current_project, comment] do |f| %>
+<%= form_with model: [current_project, @comment] do |f| %>
   <%= f.text_area :content,
-                  value: comment.content,
+                  value: @comment.content,
                   rows: 5,
                   required: 'required',
                   data: { behavior: 'mentionable' } %>
@@ -8,5 +8,5 @@
   <%= link_to 'Cancel',
               'javascript:void(0)',
               class: 'cancel-link',
-              data: {toggle_comment: "off"} %>
+              data: {behavior: "cancel-comment"} %>
 <% end %>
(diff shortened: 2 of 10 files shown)


app/views/comments/destroy.js.erb
```diff
@@ -0,0 +1,4 @@
+$("#<%= dom_id(@comment) %>").remove()
```
aapomm: should we have a `can? :destroy` line here?

dormi: I don't think it is worth it.
If we are rendering this view it is bc we went through `CommentsController#destroy`, so we have that permission.

Now that you mention it, I am not sure if it is worth it to keep https://github.com/dradis/dradis-ce/pull/400/files#diff-25ea4ccfaab28d70bcffa9056829dd06R1.
I will keep it bc it does not hurt, but we could remove it in my opinion.

aapomm: Good point

aapomm:
@dormi mentions don't work in the comment edit form. Is this intended?

dormi:
> @dormi mentions don't work in the comment edit form. Is this intended?

Yes, I think it is: https://github.com/dradis/dradis-ce/blob/master/app/models/comment.rb#L40
(and it was).
If we want to change it, let's move to another PR.

etdsoft:
> > @dormi mentions don't work in the comment edit form. Is this intended?
> 
> Yes, I think it is: https://github.com/dradis/dradis-ce/blob/master/app/models/comment.rb#L40
> If we want to change it, let's move to another PR.

@dormi what @aapomm means is @- mentions don't work in Edit

app/assets/javascripts/snowcrash/modules/comments.js.coffee
```diff
@@ -1,18 +1,12 @@
 document.addEventListener "turbolinks:load", ->
 
-  # hide/show comment editable textarea
-  $("[data-toggle-comment]").click ->
+  # hide comment editable textarea when clicking cancel
+  $('.comment-feed').on('click', "[data-behavior~=cancel-comment]", ->
```
etdsoft: can we use `data-behavior` in both selectors?

app/views/comments/_form.html.erb
```diff
@@ -1,12 +1,12 @@
...
+                  value: @comment.content,
                   rows: 5,
                   required: 'required',
                   data: { behavior: 'mentionable' } %>
   <%= f.submit "Update comment", class: "btn btn-primary" %> or
   <%= link_to 'Cancel',
               'javascript:void(0)',
               class: 'cancel-link',
-              data: {toggle_comment: "off"} %>
+              data: {behavior: "cancel-comment"} %>
```
etdsoft:
```suggestion
              data: { behavior: 'cancel-comment' } %>
```

app/assets/javascripts/snowcrash/modules/mentions.js.coffee
```diff
@@ -0,0 +1,15 @@
...
+    # initialize mentions (https://github.com/zurb/tribute)
+    unless @tribute?
+      @tribute = new Tribute(
+        allowSpaces: ->
+          false
+        menuItemTemplate: (item) ->
+          '<img src="' + item.original.avatar_url + '" width="24px" height="24px" > ' + item.string
+        noMatchTemplate: ->
+          ''
+        values: $('#mentionable-users').data('users')
```
etdsoft: should we remove this from the markup and move it into a `<meta>`?

PR #410, opened 2018-12-14
When flash messages present, add turbolinks disable cache meta

dormi:
Create 2 issues.
Update one of them, so the "Issue updated" message is displayed.
Click on the sidebar to go to the other issue.
Click on the sidebar to visit the first issue again. Note that the "Issue updated" message is displayed for a second, then disappears when the page is really loaded.

This is due to the turbo links cache: https://github.com/turbolinks/turbolinks#understanding-caching

And the proposed solution consists on adding a `meta` when flash objects are present.

### DIFF
--- a/app/views/layouts/snowcrash.html.erb
+++ b/app/views/layouts/snowcrash.html.erb
@@ -11,6 +11,9 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=0, minimum-scale=1.0, maximum-scale=1.0">
   <meta name="apple-mobile-web-app-capable" content="yes">
   <meta name="apple-mobile-web-app-status-bar-style" content="black">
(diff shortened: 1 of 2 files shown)


app/views/layouts/snowcrash.html.erb
```diff
@@ -11,6 +11,9 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=0, minimum-scale=1.0, maximum-scale=1.0">
   <meta name="apple-mobile-web-app-capable" content="yes">
   <meta name="apple-mobile-web-app-status-bar-style" content="black">
+  <% if flash.any? %>
```
aapomm: @dormi can you add comments as to why this condition is here?

dormi: 👍

PR #411, opened 2018-12-21
Add sort to attachments

aapomm:
### Spec
Currently, attachments are ordered arbitrarily in the attachments list in the nodes view.

**Proposed solution**
Use the `sort` method to sort the attachments.

### How to test

1. Visit a node
2. Upload a number of attachments
3. Assert that they are sorted alphabetically


etdsoft:
Can you please update CHANGELOG and merge?

PR #421, opened 2019-01-16
Try to fix attachments spec

dormi:
There is a spec failing randomly:

```
Failures:

1) Attachments API as authenticated user POST /api/nodes/:node_id/attachments returns 422 when no file saved
Failure/Error: expect(File.exist?(Attachment.pwd.join(node.id.to_s))).to be false

  expected false
       got true
# ./engines/dradis-api/spec/requests/dradis/ce/api/v1/attachments_spec.rb:147:in `block (4 levels) in <top (required)>'

Finished in 10 minutes 44 seconds (files took 9.18 seconds to load)
1610 examples, 1 failure, 72 pending

Failed examples:

rspec ./engines/dradis-api/spec/requests/dradis/ce/api/v1/attachments_spec.rb:142 # Attachments API as authenticated user POST /api/nodes/:node_id/attachments returns 422 when no file saved
```

With this change we try to assert that every spec always find an empty attachments folder.

### DIFF
--- a/engines/dradis-api/spec/requests/dradis/ce/api/v1/attachments_spec.rb
+++ b/engines/dradis-api/spec/requests/dradis/ce/api/v1/attachments_spec.rb
@@ -42,18 +42,17 @@ describe "Attachments API" do
   context "as authenticated user" do
     include_context "authenticated API user"
 
-    def empty_attachments
-      node_attachments = Attachment.pwd.join(node.id.to_s)
-      until Dir["#{node_attachments}/*"].count == 0 do
-        FileUtils.rm_rf(Attachment.pwd.join(node.id.to_s))
+    before(:each) do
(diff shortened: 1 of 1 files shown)


etdsoft:
It seems that the problem is a specific test not cleaning after itself (here or elsewhere in the specs). I'd prefer a fix that addresses that. However, I can see the simplicity of the 'delete' all approach.

That being said, why not just delete the folder, instead of looping through the contents? Either by deleting the folder itself or by deleting '*' in one go - `FileUtils.rm_rf(Dir['dir/to/remove/*'])`

dormi:
Deleting the folder itself caused other failures in the spec. Looks like this line https://github.com/dradis/dradis-ce/blob/master/app/models/attachment.rb#L68 is not recreating the folder fast enough.
I moved to deleting the contents in one go.

PR #426, opened 2019-01-18
Add start of line (^) modifier for bc.. terminals

aapomm:
### Spec
The regex terminates the `bc..` block even if the new textile block elements are not in a newline.

### How to test:
1. Add an issue with the content:
```
#[Description]#
bc.. (C) Copyright 1985-2001 Microsoft Corp.    
D:\Program Files\microsoft_corp>
printf "Hello World"
```

2. Click on the preview tab.
3. Confirm that the characters are not escaped when displayed.


dormi:
before:
captura de pantalla 2019-02-06 a las 11 35 56

after:
captura de pantalla 2019-02-06 a las 11 36 26

so 👍

PR #430, opened 2019-01-23
[WIP] Fix revision spec

aapomm:
Check revisions confirmation using insensitive case to catch the message.


spec/support/revision_shared_examples.rb
```diff
@@ -77,7 +77,7 @@
         expect(page).not_to have_content item_type.to_s
       end
       expect(model.class.find_by_id(model.id)).not_to be_nil
-      expect(page).to have_content 'Recovered'
+      expect(page).to have_content(/recovered/i)
```
dormi: do we have to use a regex? I mean, it is upcase or downcase, right? and we know it. Using the right format should just work?

PR #431, opened 2019-01-29
Nodes#tree - sort in an IP-meaningful way

etdsoft:
This PR causes the nodes loaded via Ajax for the nodes tree to be sorted in an IP-meaningful way.

To test:

1. Create a node
2. Click 'Add subnode' and 'add multiple', then paste the following:

```
127.0.0.10
127.0.0.2
127.0.0.12
127.0.0.11
127.0.1.10
127.0.10.10
127.0.1.1
127.0.1.11
```

Before (left, alpha sorting) and after (right, IP sorting) this PR:

screen shot 2019-01-29 at 13 50 26

### DIFF
--- a/app/controllers/nodes_controller.rb
+++ b/app/controllers/nodes_controller.rb
@@ -107,7 +107,7 @@ class NodesController < NestedNodeResourceController
   # Lazy-load the nodes' tree via Ajax calls to this action.
   def tree
     # TODO: Do we want to use :find_or_initialize_node or skip it and :include children?
-    @children = @node.children
+    @children = @node.children.to_a.sort_by!{ |node| node.label.split('.').map(&:to_i) }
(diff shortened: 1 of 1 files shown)


etdsoft:
@dormi you can try it out!

dormi:
Sorry, what I mean is that I'm not sure it is working properly:
captura de pantalla 2019-01-29 a las 22 21 50

etdsoft:
You're right, we'd have to sort IP-looking nodes as one group, then non-IP looking ones, and merge the results, seems like a more complicated effort than originally expected.

Interestingly we're using this algo to sort in the ContentService:
https://github.com/dradis/dradis-plugins/blob/master/lib/dradis/plugins/content_service/nodes.rb#L11-L17

PR #432, opened 2019-02-04
Warden - fix :failure_app config

etdsoft:
When providing the wrong credentials, we had a Warden config problem that resulted in an exception thrown instead of the login form being re-rendered.

This PR fixes the warden config to handle bad creds properly.

See:
  https://github.com/wardencommunity/rails_warden/issues/30


etdsoft:
Closing in favour of https://github.com/dradis/dradis-ce/pull/439/

PR #436, opened 2019-02-19
Fix textile stacking

aapomm:
### Spec
Currently, when editing an item, moving back in browser history and forwards stacks the textile toolbar in the view. This happens because the jquery data that we use to track the presence of the modified textarea does not persist when going through the history, but the modified textarea itself does (thanks to turbolinks).

image001

**Proposed solution**
Use a flag (`data-textiled`) in the DOM itself to check for the modified textarea.

### Steps to replicate: 

1. Open and Issue
2. Click edit
3. Use the back button in your browser and then the forward button to send you back to the edit page
4. Confirm that the "write/preview" tabs duplicate with every back/forward:


vendor/assets/javascripts/jquery.textile.js
```diff
@@ -46,15 +46,27 @@
...
 
   Plugin.prototype = {
-    init: function () {
+    init: function ( exists ) {
       // Place initialization logic here
       // You already have access to the DOM element and the options via the instance,
       // e.g., this.element and this.options
-      this._buildContainer();
+      if (exists) {
+        this.options.$help    = $('.textile-help');
```
dormi: is this code "duplicated" ?

Looks like it is also performed by `_buildContainer`.
So I was thinking if we could do something like:
```
if (!exists) {
  this._buildContainer();
}

...[this code]...
```

vendor/assets/javascripts/jquery.textile.js
```diff
@@ -46,15 +46,27 @@
     this._defaults = defaults;
     this._name = pluginName;
 
-    this.init();
+    this.init(this.$element.data('textiled'));
   }
 
   Plugin.prototype = {
-    init: function () {
+    init: function ( exists ) {
```
dormi: can we get the `exists` (check `data-textilized`) value here in the constructor and skip passing it?

CHANGELOG
```diff
@@ -14,6 +14,7 @@ v3.12 (April 2019)
  - Disable turbolinks cache when displaying flash messages
  - Sort attachments in alphabetical ASCII order
  - Fix methodology checklist edit error
+ - Fix textile toolbar stacking
```
dormi: ... and breadcrumbs

vendor/assets/javascripts/jquery.textile.js
```diff
@@ -74,45 +77,49 @@
...
+
+      this.$element.attr('data-textiled', 'true');
     },
     _buildToolbar: function() {
       var button;
 
       // Write
       button = $('<a class="btn-write active" href="javascript:void(null);"><span>Write</span></a>');
-      button.click( $.proxy( function(evt) { this._onBtnWrite(evt); }, this));
+      button.click( function(evt) { this._onBtnWrite(evt) }.bind(this) );
```
dormi: this change you are doing, from `$.proxy` to `.bind(this)`, I think it is more readable. Is that the reason for the change?

aapomm: They are functionally the same but `$.proxy` relies on jquery while `.bind` is available on plain javascript.

etdsoft: why are all of these `$.proxy` changes needed?

aapomm: Not needed as they are functionally the same, but `$.proxy` relies on jquery while `.bind` is available on plain javascript.

PR #439, opened 2019-02-21
Warden failure app (v2)

dormi:
### Spec 

This is another attempt to https://github.com/dradis/dradis-ce/pull/432

When:
- we login with invalid credentials
- we make a POST/PATCH request and the session is expired 
We are redirected to the warden failure app. But since the request triggered in these cases check the csrf token, the failure app raises an error.

In this PR we try to set up a unique warden failure app, that does not check csrf token.
This failure app then just redirect to login for html requests, and for other requests (js/json) just returns 404 (not found)

### How to test
* Switch to `warden-failure-app-2` and confirm you now see a form when trying to login with bad credentials.
* In CE, create a node with subnodes (so when clicking the triangle to unfold the parent node a js request is generated). 
* Alter the app session timeout to be smaller. To do so edit this line: https://github.com/dradis/dradis-ce/blob/master/config/initializers/warden_99_timeout.rb#L30 so it looks like: 
  ` if last_request_at && last_request_at <= 10.seconds.ago`
and restart the server
* Login and visit a node so the attachments dropbox is visible. Wait > 10 seconds so the session expires.
 * When we try to upload an attachment, we should see an error like this:
captura de pantalla 2019-02-21 a las 20 41 39
* When we try to unfold a node in the sidebar we must see:
captura de pantalla 2019-02-21 a las 21 18 51
    
* Reload the page and assert we see the login form
* Login and visit an issue edit page, but don't save it. Wait > 10 seconds so the session expires and submit the form. We should be redirected to the login page.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -11,7 +11,6 @@ Rails.application.routes.draw do
   # Sign in / sign out
   get '/login'  => 'sessions#new'
   get '/logout' => 'sessions#destroy'
-  match '/timeout' => 'sessions#timeout', via: :all
   resource :session
 
   # ------------------------------------------------------------ Project routes
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -23,7 +23,7 @@ Rails.application.routes.draw do
   resources :projects, only: [:show] do
     resources :activities, only: [] do
       collection do
-        get :poll, defaults: { format: :js }
+        get :poll, constraints: { format: /js/ }
       end
     end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -11,6 +11,7 @@ Rails.application.routes.draw do
   # Sign in / sign out
   get '/login'  => 'sessions#new'
   get '/logout' => 'sessions#destroy'
+  match '/timeout' => 'sessions#timeout', via: :all
   resource :session
 
   # ------------------------------------------------------------ Project routes
@@ -23,7 +24,7 @@ Rails.application.routes.draw do
   resources :projects, only: [:show] do
     resources :activities, only: [] do
       collection do
-        get :poll, constraints: { format: /js/ }
+        get :poll, defaults: { format: :js }
       end
     end
 
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -8,8 +8,7 @@ module Authentication
   end
 
   def access_denied
-    # specify 'main_app' explicitly in case we're inside an engine.
-    redirect_to(main_app.login_path, alert: 'Please sign in first. Access denied.')
+    throw :warden, message: 'Please sign in first. Access denied.'
   end
 
   # Proxy to the authenticated? method on warden
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -3,6 +3,7 @@ class SessionsController < ApplicationController
   before_action :ensure_setup,          only: :new
   before_action :ensure_not_setup,      only: [:init, :setup]
   before_action :ensure_valid_password, only: :setup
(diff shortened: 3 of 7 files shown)


config/routes.rb
```diff
@@ -11,6 +11,7 @@
   # Sign in / sign out
   get '/login'  => 'sessions#new'
   get '/logout' => 'sessions#destroy'
+  match '/timeout' => 'sessions#timeout', via: :all
```
etdsoft: is this ever used?

config/routes.rb
```diff
@@ -23,7 +24,7 @@
   resources :projects, only: [:show] do
     resources :activities, only: [] do
       collection do
-        get :poll, constraints: { format: /js/ }
+        get :poll, defaults: { format: :js }
```
etdsoft: this doesn't seem related? Why is it needed?

dormi: When I was checking with the browser console what was going on, I noticed that polling requests when the session expired tried to be redirected.
That is because rails was processing them as */* and in the failure app they were redirected as html format.
With this the poll requests is processed as js in the failure app the get a 404, because that is what we give to any html format.
Before, when a poll was received for and the session expired:
captura de pantalla 2019-02-21 a las 21 56 05
```
Processing by SessionsController#failure as */*
  Parameters: {"last_poll"=>"1550786143", "project_id"=>"1"}
Redirected to http://localhost:3000/login
Completed 302 Found in 1ms (ActiveRecord: 0.0ms)
```
And now it is:
captura de pantalla 2019-02-21 a las 21 57 24
```
Processing by SessionsController#failure as JS
  Parameters: {"last_poll"=>"1550786228", "project_id"=>"1"}
Completed 404 Not Found in 1ms (ActiveRecord: 0.0ms)
```

config/routes.rb
dormi: no, you are right. Removed.

config/routes.rb
dormi: ^ wait, scratch this please. I am seeing a different result when I try this on `master` branch now. The `poll` request gets the html login form now but I see no redirection.

dormi: Ok, I figured it out:

Now in `master` branch, when we get an expired `poll` request, it is processed by the current failure app `SessionController.new`, so it tries to render the login form with a `Session Expired` message:
captura de pantalla 2019-02-25 a las 10 39 30
(it is a js request and nobody sees that)

In this PR branch, we are introducing a new failure app: `SessionsController.failure`, that redirects expired html requests to login forms. That's why when we got an expired `poll` request, we were seeing a redirect:
captura de pantalla 2019-02-21 a las 21 56 05
The `poll` request format was not `js`, was `*/*`, so in the failure app it was handled as html.

With this change: `get :poll, defaults: { format: :js }` we are making the `poll` requests be processed as `js` requests:
captura de pantalla 2019-02-21 a las 21 57 24
Then when we get an expired `js` request we are just returning `head :unknown`.

etdsoft: I think the spirit of the original line was: only allow JS calls to this endpoint.

The problem was the client-side poller wasn't hitting `/poll.js` but `/poll`.

With this change, we're making any calls hitting `/poll` to be considered JS calls. Which works, however, is that what we want? I think that the other way to fix it (making the JS poller actually hit the JS endpoing) is preferable / clearer / more consistent with other JS calls (tree navigation for instance, console logger)?

In any case, we can move this to a separate PR, as the rest of the Warden-related code is ready to merge.

dormi: I understand what you say, and I think it was my other option to do it. I think it can be done like this: https://github.com/dradis/dradis-ce/pull/439/commits/3a4e524b4360b66b5f4030268e5186a2cc3212eb

PR #447, opened 2019-03-20
Add auto-save JS that uses localStorage on textarea textchange events

etdsoft:
This PR uses `localStorage` to save the in-progress value in your Issues#form so if your browser crashes or you accidentally close your tab, you can restore the contents of the editor from the browser's cache.


app/assets/javascripts/snowcrash/modules/auto_save.js
```diff
@@ -0,0 +1,49 @@
+document.addEventListener( "turbolinks:load", function(){
+  console.log('auto_save loaded');
+
+  document.querySelectorAll("[data-behavior~=auto-save]").forEach( function(item){
```
aapomm: Why not use jquery selector?

```suggestion
  $("[data-behavior~=auto-save]").each( function(index){
```
then use `$(this)` for the item.

app/assets/javascripts/snowcrash/modules/auto_save.js
```diff
@@ -0,0 +1,49 @@
...
+    $(item).on("textchange", function(event, previousText) {
+      console.log('textchange');
+      console.log(event);
+
+      clearTimeout(timer)
+
+      timer = setTimeout(function(){
+        console.log("Saving to localStorage...");
+        console.log(event.currentTarget.value);
+        if (typeof Storage !== "undefined" && Storage !== null) {
```
aapomm: Why not check for `Storage` before creating a timeout instead of seeing a console message every text change?

app/assets/javascripts/snowcrash/modules/auto_save.js
```diff
@@ -0,0 +1,49 @@
...
+      }
+    } else {
+      console.log("No data in localStorage for " + key);
+    }
+
+    var timer;
+
+    $(item).on("textchange", function(event, previousText) {
+      console.log('textchange');
+      console.log(event);
```
aapomm: `console.log` seems to be flooding the console. We should limit it otherwise we'll miss the important errors.

app/assets/javascripts/snowcrash/modules/auto_save.js
etdsoft: because I can see a future in which we're going to move away from jQuery... I think that if we keep this in mind when adding little snippets, we'll have an easier time when it's time to pull the plug.

app/assets/javascripts/snowcrash/modules/auto_save.js
etdsoft: you're right, none `console.log` calls should survive the review.

PR #461, opened 2019-05-19
Merge nodes functionality

nikobozi:
### Summary

This PR addresses [this issue](https://trello.com/c/xs0puZDz). 
In the node's actions bar there is a new `Merge` action that moves associated: `Notes`, `Evidence` `Attachments` and `Activities` from the source node into the target node.
After the merge, the source node is being deleted

### Other Information

- Action items are now grouped inside a drop down since there is many of them and they would otherwise go in 2 rows
- Since there is lots of data reassigning between nodes, the operation is wrapped inside a `ActiveRecord::Base.transaction` so that there is a rollback in case something goes wrong

### Demo
merge-demo

### Copyright assignment

### DIFF
--- a/app/assets/stylesheets/snowcrash/modules/nodes.scss
+++ b/app/assets/stylesheets/snowcrash/modules/nodes.scss
@@ -46,3 +46,14 @@ td.services-extras-datum-output {
   // layout:
   word-break: break-all;
 }
+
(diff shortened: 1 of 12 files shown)


etdsoft:
I'm going to close this for now, and we'll review it / reopen it on on Thu.

PR #463, opened 2019-05-24
Correct diff behavior for first revision

### LINKED ISSUE #462: Diff for first revision displays incorrectly
### Steps to reproduce

Go into an issue which has no revision history (just the initial creation). Make a small edit. Now go to view the revision history and there is an unhelpful diff which doesn't just highlight the changed line but interprets all the content as having changed. Does does not seem to happen for later revisions, only for the first one.

### Expected behavior
The same changes should give the exact same diff result regardless of what revision the change was part of.

### Actual behavior
It behaves differently on the first revision.

### System configuration
**Dradis version**: 3.12

**Ruby version**: 2.4.1

**OS version**: MacOS 10.13.6

em77:
### Summary

Fixes #462. The problem was that the first version uses `\n` for a newline and successive versions use `\r\n` so I made the first version use the same format to get the desired results.

### Copyright assignment

### DIFF
--- a/app/models/diffed_revision.rb
+++ b/app/models/diffed_revision.rb
@@ -11,7 +11,7 @@ class DiffedRevision
   def diff
     @diff ||= Differ.diff_by_line(
                 after[content_attribute],
-                before[content_attribute]
+                before_content
               )
   end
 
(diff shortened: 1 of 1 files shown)


em77:
My first crack at fixing this only fixed it on sample issues but it should now give the same results for user-created issues.

PR #464, opened 2019-05-24
Tag crud

NuclearMachine:
### **Spec**
Currently Tags for the project are created at seed time:
https://github.com/dradis/dradis-ce/blob/master/db/seeds.rb

They can be used across the app (Project#show, Issues#index, Issues#form), but there is no way to modify / add / delete them without firing the database.

We need a better way for managing this.

### **Proposed solution**

A new Tags#Index with CRUD functionalities could be used to manage the list of tags. It's input fields should not allow invalid characters not currently supported by the model. 

### **How to test**

**Creating new tags**
Clicking the `New Tag` button should open a modal where the user can input the Tag details and submit.
new

**Updating existing tags**
A user can edit a Tag's `color` or `name` by clicking their respective fields. Tag `name`s are automatically submitted when the input field loses its `focus` (by clicking anywhere in the webpage other than the `focused` field) or when the user presses `enter`, `!` or `space`. Similarly, tag `color`s are automatically submitted when the user finishes selecting a color. 
update

**Deleting existing tags**
A user can delete a Tag by clicking the `delete` button by the options field. An alert confirmation should appear first before deletion.
delete

**Accessing the Index**
A user should be able to access the **Tags#Index** by going the **Issues#Index** view and clicking the Tags button right beside the toolbar. Tags should be paginated and only limited to `12` per page.
access

**Searching the Index**
Inputting a Tag's `name` or `color` attribute in the `Tags#Index`'s search field should show that specified tag (or any similar tags). The `results` info should show the proper tag count and queried input. A helpful text warning the user should also appear when none are found. 
searching

### Bonus
Currently, assigning tags to an `Issue` requires the user to click a drop-down. If the user decided to create more than `15` tags, it would look very odd to see in a drop-down (given the limited page height) and time consuming to search. 

**Assigning tags**
To assign a `tag` to an existing Issue, enter its edit page and click the tag button right below (where the tag drop-down) used to be. Search for the tags name or color, then click. 
selecting

**Searching Tags**
Searching tags in the Issues form is done via `AJAX`. A loading icon should appear upon `AJAX` submission and only disappear once the script is finished filling the results tab. Since it uses the same controller method as `Tags#Index`, it is currently paginated to `12` per page. 
searchingTags


etdsoft:
I'm going to close this for now, and we'll review it / reopen it during the call.

PR #467, opened 2019-06-04
Delete multiple Evidence from Issue#show page

em77:
### Summary

Currently if an Issue is associated with multiple Nodes or multiple Evidence, there is no easy way to bulk-delete the evidence from the Issues#show view, you have to either open each Evidence individually or open each Node and bulk-delete all Evidence for that Node.

This PR's aim is to make it possible bulk-delete Evidence from Issue#show.

### Other Information

There is a problem I ran into in the past few hours while going over this that I didn't have time to find a solution for yet. If the user clicks on a node other than the pre-selected node, the full desired checking functionality breaks for selecting individual checkboxes (the Delete button will not appear). Using the Select All checkbox still selects all the checkboxes and it is possible to first do that and then deselect individual checkboxes but it will not deselect the Select All checkbox. Refreshing the page is the only way to get it to work again for the pre-selected node after that. These problems don't exist for the pre-selected node.

At first I thought it might be a conflict with another event listener but it looks like no listeners are bound to the checkboxes which are not initially displayed so I think it may be related to their "visibility" when the page is first loaded. What do you all think?

### Copyright assignment

### DIFF
--- a/app/assets/javascripts/snowcrash/modules/issues/evidence.js.coffee
+++ b/app/assets/javascripts/snowcrash/modules/issues/evidence.js.coffee
@@ -1,40 +1,106 @@
 document.addEventListener "turbolinks:load", ->
-  if $('body.issues.show').length
-    $('#evidence-tabs li').first().addClass('active')
-    $('#evidence-tabs .tab-pane').first().addClass('active')
+  return unless $('body.issues.show').length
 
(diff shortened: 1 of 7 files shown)


etdsoft:
Thanks! I'm going to close this for now, and we'll review it / reopen it during the call.

PR #469, opened 2019-06-06
Combine node feature

brianp:
## Spec

At the moment when you create a Node in the tree you can only move it around to a different part of the tree.

There are situations in which you create a Node, for example for IP 10.0.0.1 and add some data to it, and then you realise that a node already existed for that IP somewhere in the tree.

There is no way to move the Notes, Evidence and Attachments associated with a node to another node.

## Proposed solution

A new "Merge" action should be added to the node's actions bar that would move associated: Notes, Evidence, Attachment and Activities from the source node into the target node.

Upon clicking that action, a modal similar to the one triggered by the Move action will be displayed so the user can chose the target node for the combine operation.

The source node will be deleted after the combine.

## How to test
1. Create new nodes to test with by using the `+` button to open a new node modal.
    image
2. In the modal select "Add multiple" and add two new node names. One on each line.
    image
3. Navigate to you duplicated node.
4. Add a new note, piece of evidence, and an attachment.
5. Now utilize the new "burger" menu on the top right of the screen and select "Merge"
    burger-menu
6. Select the node you wish to Merge the current duplicate into, and than press "Merge"
    image
7. Now you should be redirected to the Node you had selected to merge into.
8. The note, evidence, and attachment from the duplicate will now be present on your current node.

### Successful merge
 - In this example `Subnode - 1.1.1.1` contains no additional notes, evidence, or attachments.
 - `Subnode Dup - 1.1.1.1` contains a single piece of evidence, a note, and an attachment.
 - `Subnode Dup - 1.1.1.1`'s evidence, note, and attachment will all be merged over to `Subnode - 1.1.1.1`
 - The page will redirect to `Subnode - 1.1.1.1`
dradis-combine-node-success

### Failed merge
dradis-combine-node-error

### DIFF
--- a/app/assets/stylesheets/snowcrash/base.scss
+++ b/app/assets/stylesheets/snowcrash/base.scss
@@ -312,6 +312,14 @@ body {
 
     .node-header {
 //      height: $breadCrumbHeight;
+
+      // Styling of the breadcrumbs menu items conflicts when trying to match
+      // dropdown style of the main navigation. The dropdown ends up always
+      // pulling right, off the screen. This pulls it left, and makes it visible
+      .dropdown-menu-more {
+        left: auto;
+        right: 0;
+      }
     }
     .breadcrumb {
       margin-bottom: 0;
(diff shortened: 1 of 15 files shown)


etdsoft:
Thanks! I'm going to close this for now, and we'll review it / reopen it during the call.

brianp:
Closing in favour of #479 so CI can run.

PR #473, opened 2019-07-04
Consolidate `Import Issues` and `Issues` headers in the sidebar into one header.

MattBudz:
This PR adds the down-chevron from `Import Issues` into the `Issues` header located in the `All Issues` view.


app/views/shared/_sidebar_collection.html.erb
```diff
@@ -1,14 +1,16 @@
...
       <%= render 'shared/add_item_dropdown',
         name: name,
         new_path: new_path,
-        category_id: category_id,
-        menu_class: 'pull-right'
+        category_id: category_id
       %>
+      <% if name =='Issues' %>
+        <%= render 'import_box' %>
+      <% end%>
```
brianp: Missing space between `end` and `%`
ex. `<% end %>`

app/views/shared/_sidebar_collection.html.erb
```diff
@@ -1,14 +1,16 @@
...
+    <h3><%= local_assigns[:index_path] ? link_to(name, index_path) : name %></h3>
     <div class="options">
       <%= render 'shared/add_item_dropdown',
         name: name,
         new_path: new_path,
-        category_id: category_id,
-        menu_class: 'pull-right'
+        category_id: category_id
       %>
+      <% if name =='Issues' %>
```
brianp: Missing space after `==` before `'Issues'`
ex. `<% if name == 'Issues' %>`

brianp: I don't _love_ that we're checking a string to see if we want to add additional partials. I'm not sure I have a better idea, as much as a different idea.

**issues/_sidebar.html.erb**
```erb
  name: 'Issues',
  collection: @issues,
  new_path: main_app.new_project_issue_path(current_project),
  category_id: nil  do
    render('import_box')
  end
%>
```

**shares/_sidebar_collection.html.erb**
```erb
...
    <div class="options">
      <%= render 'shared/add_item_dropdown',
        name: name,
        new_path: new_path,
        category_id: category_id
      %>
      <%= yield %>
    </div>
...
```

If no block is passed the yield is simply ignored. If a block is passed it'll render the given block. In this case rendering an additional partial.
I like it because it keeps the knowledge at the caller, and the sidebar doesn't need to know about what it is rendering.
It could also be a little off putting to come across in the code, so I think a second opinion would be good.

MattBudz: When I implement these 2 code blocks in the respective files, the down-caret doesn't show up and the import box doesn't render. Maybe I missed something?

brianp: We clarified the issue was the usage of `<% yield %>` where it should be `<%= yield %>`. Got it working now.

app/views/shared/_sidebar_collection.html.erb
```diff
@@ -1,14 +1,14 @@
...
+    <h3><%= local_assigns[:index_path] ? link_to(name, index_path) : name %></h3>
     <div class="options">
       <%= render 'shared/add_item_dropdown',
         name: name,
         new_path: new_path,
-        category_id: category_id,
-        menu_class: 'pull-right'
+        category_id: category_id
       %>
+      <%= yield %>
```
brianp: This line might get just a bit safer if we added a conditional.
```erb
<%= yield if block_given? %>
```

MattBudz:
This did not have the desired functionality so I am closing this PR.

New PR is [here](https://github.com/dradis/dradis-ce/pull/474)

PR #474, opened 2019-07-17
Collapsable issues

MattBudz:
Currently, when viewing issues, the list of issues in the sidebar can get quite long, causing the "Import issues" dropdown to be far down in the view. 

This PR adds the capability to collapse the issues list, which results in quick and easy access to the "Import issues" dropdown. The list of issues, when expanded, is now limited in size and is scrollable.

### DIFF
--- a/app/assets/javascripts/snowcrash/modules/issues.js.coffee
+++ b/app/assets/javascripts/snowcrash/modules/issues.js.coffee
@@ -3,10 +3,3 @@
 //= require snowcrash/modules/issues/merge
 //= require snowcrash/modules/issues/table
 //= require snowcrash/modules/issues/tag-input
-
-document.addEventListener "turbolinks:load", ->
-  if $('body.issues, body.nodes, body.cards').length
-    $('.import-toggle').click ->
-      $this = $(this)
-      $this.find('i').toggleClass('fa-chevron-down fa-chevron-up')
-      $($this.data('target')).find("input[type='text']:first").focus()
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/snowcrash/modules/issues.js.coffee
```diff
@@ -5,7 +5,7 @@
 //= require snowcrash/modules/issues/tag-input
 
 document.addEventListener "turbolinks:load", ->
-  if $('body.issues').length
+  if $('body.issues, body.nodes, body.cards').length
```
etdsoft: @MattBudz if we're making this a global feature, it needs to be extracted out of `issues.js`. It also needs to be based on data-behavior, so we don't have perpetuate this ugliness.

app/assets/stylesheets/snowcrash/modules.scss
```diff
@@ -148,3 +148,10 @@ table.table-ajax-inputs {
     }
   }
 }
+
+#issues, #notes, #evidence, #tasks {
```
etdsoft: @MattBudz 
- What is the purpose of this code?
- Why do we have to repeat it for each of these views?
- Looks like a behaviour we want in our layout (e.g. the content panel of every view)? In which case we could move it to the right file and reference the right class?

I'm concerned about us adding a new view further down the line and forgetting we have to hard-code it here too.

MattBudz: This allows the collection to be vertically scrolled through in the sidebar. 
I agree that listing ID's is not the best approach. I have changed this to use a common class.

PR #475, opened 2019-07-24
Add tunnel to node properties

aapomm:
Related to the nmap change: https://github.com/dradis/dradis-nmap/pull/14


aapomm:
Not needed.

PR #479, opened 2019-07-25
Merge node feature

brianp:
## Spec

At the moment when you create a Node in the tree you can only move it around to a different part of the tree.

There are situations in which you create a Node, for example for IP 10.0.0.1 and add some data to it, and then you realise that a node already existed for that IP somewhere in the tree.

There is no way to move the Notes, Evidence and Attachments associated with a node to another node.

## Proposed solution

A new "Merge" action should be added to the node's actions bar that would move associated: Notes, Evidence, Attachment and Activities from the source node into the target node.

Upon clicking that action, a modal similar to the one triggered by the Move action will be displayed so the user can chose the target node for the combine operation.

The source node will be deleted after the combine.

## Noteworthy changes since the original PR

- Tests cleanup after themselves instead of cleaning attachments after *every* test via the test_helper.
- Attachment merging method has changed (description below)
- Properties merge utilizing the existing property management!
- The node merger service no longer accepts a block, and deletes the source node
- Removed the strange result type from the service

### Attachment merging method

Previously the method was to *Move* existing attachments over to the target
node. If an error occurred we *Moved* them back to the source.

Once we account for possible naming collisions we're no longer just moving a
file but we're changing it. If an error occurs we would have to track the changes
rename the file, and move it back.

Instead we've implemented a new solution. *Copy* the source files to the target.
Track the files we've moved in *this* merge. If an error occurs simply remove
all the newly copied files. Source files stay exactly where they are until the
source node is deleted.

## How to test
1. Create new nodes to test with by using the `+` button to open a new node modal.
    image
2. In the modal select "Add multiple" and add two new node names. One on each line.
    image
3. Navigate to you duplicated node.
4. Add a new note, piece of evidence, and an attachment.
5. Now utilize the new "burger" menu on the top right of the screen and select "Merge"
    burger-menu
6. Select the node you wish to Merge the current duplicate into, and than press "Merge"
    image
7. Now you should be redirected to the Node you had selected to merge into.
8. The note, evidence, and attachment from the duplicate will now be present on your current node.

### Successful merge
 - In this example `Subnode - 1.1.1.1` contains no additional notes, evidence, or attachments.
 - `Subnode Dup - 1.1.1.1` contains a single piece of evidence, a note, and an attachment.
 - `Subnode Dup - 1.1.1.1`'s evidence, note, and attachment will all be merged over to `Subnode - 1.1.1.1`
 - The page will redirect to `Subnode - 1.1.1.1`
dradis-combine-node-success

### Failed merge
dradis-combine-node-error

### DIFF
--- a/app/assets/stylesheets/snowcrash/base.scss
+++ b/app/assets/stylesheets/snowcrash/base.scss
@@ -310,17 +310,7 @@ body {
     overflow-y: auto;
     padding: 0 $mainContentPadding;
 
-    .node-header {
-//      height: $breadCrumbHeight;
 
-      // Styling of the breadcrumbs menu items conflicts when trying to match
-      // dropdown style of the main navigation. The dropdown ends up always
-      // pulling right, off the screen. This pulls it left, and makes it visible
-      .dropdown-menu-more {
-        left: auto;
-        right: 0;
-      }
-    }
     .breadcrumb {
       margin-bottom: 0;
       padding: 0.5em 1em;
--- a/app/assets/stylesheets/snowcrash/base.scss
+++ b/app/assets/stylesheets/snowcrash/base.scss
@@ -312,6 +312,14 @@ body {
 
     .node-header {
 //      height: $breadCrumbHeight;
+
+      // Styling of the breadcrumbs menu items conflicts when trying to match
+      // dropdown style of the main navigation. The dropdown ends up always
+      // pulling right, off the screen. This pulls it left, and makes it visible
+      .dropdown-menu-more {
+        left: auto;
+        right: 0;
+      }
     }
     .breadcrumb {
       margin-bottom: 0;
--- a/app/services/nodes/merger.rb
+++ b/app/services/nodes/merger.rb
@@ -22,6 +22,7 @@ class Nodes::Merger
       move_descendents
       reset_counter_caches
       move_attachments
+      source_node.destroy
 
       return []
     end
@@ -46,7 +47,6 @@ class Nodes::Merger
 
     def reset_counter_caches
       Node.reset_counters target_node.id, :children_count
-      Node.reset_counters source_node.id, :children_count
     end
 
     def move_attachments
--- a/app/services/nodes/merger.rb
+++ b/app/services/nodes/merger.rb
@@ -9,7 +9,7 @@ class Nodes::Merger
   }.freeze
 
   def self.call(target_node, source_node)
-    new(target_node, source_node).call(&block)
+    new(target_node, source_node).call
   end
 
   def initialize(target_node, source_node)
@@ -17,13 +17,11 @@ class Nodes::Merger
     @target_node = target_node
   end
 
-  def call(&block)
+  def call
     Node.transaction do
       move_descendents
       reset_counter_caches
       move_attachments
-
-      yield if block
 
       return []
     end
--- a/app/services/nodes/merger.rb
+++ b/app/services/nodes/merger.rb
@@ -1,13 +1,6 @@
 # frozen_string_literal: true
 
 class Nodes::Merger
-  SOURCES = {
-    activities: :trackable_id,
-    children: :parent_id,
-    evidence: :node_id,
-    notes: :node_id
-  }.freeze
-
   def self.call(target_node, source_node)
     new(target_node, source_node).call
   end
@@ -32,16 +25,23 @@ class Nodes::Merger
 
     undo_attachments_move
 
-    return [e.message]
+    []
   end
 
   private
 
     attr_accessor :target_node, :source_node, :source_attachments
 
+    DESCENDENT_RELATIONSHIPS = {
+      activities: :trackable_id,
+      children: :parent_id,
+      evidence: :node_id,
+      notes: :node_id
+    }.freeze
+
     def move_descendents
-      SOURCES.each do |relation, column|
-        source_node.send(relation).update_all(column => target_node_id)
+      DESCENDENT_RELATIONSHIPS.each do |relation, column|
+        source_node.send(relation).update_all(column => target_node.id)
       end
     end
 
--- a/app/services/nodes/merger.rb
+++ b/app/services/nodes/merger.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+class Nodes::Merger
+  SOURCES = {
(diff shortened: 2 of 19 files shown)


app/services/nodes/merger.rb
```diff
@@ -0,0 +1,72 @@
...
+    @source_node = source_node
+    @target_node = target_node
+  end
+
+  def call
+    Node.transaction do
+      move_descendents
+      reset_counter_caches
+      move_attachments
+      update_properties
```
brianp: Update properties before moving attachments

app/assets/stylesheets/snowcrash/base.scss
```diff
@@ -312,6 +312,14 @@ body {
...
     .node-header {
 //      height: $breadCrumbHeight;
+
+      // Styling of the breadcrumbs menu items conflicts when trying to match
+      // dropdown style of the main navigation. The dropdown ends up always
+      // pulling right, off the screen. This pulls it left, and makes it visible
+      .dropdown-menu-more {
+        left: auto;
+        right: 0;
+      }
```
MattBudz: I don't think making a new class for this is necessary. Simply adding a `pull-right` class to the `<ul>` element will achieve this.

app/views/layouts/snowcrash/_breadcrumb.html.erb
```diff
@@ -10,10 +10,17 @@
...
   <div class="span6">
     <ul class="nav nav-pills pull-right">
-      <li><a href="#modal_add_child_node" tabindex="-1" data-toggle="modal"><i class="fa fa-plus"></i> Add subnode</a></li>
-      <li><a href="#modal_delete_node" class="text-error" tabindex="-1" data-toggle="modal"><i class="fa fa-trash"></i> Delete</a></li>
-      <li><a href="#modal_rename_node" tabindex="-1" data-toggle="modal"><i class="fa fa-edit"></i> Rename</a></li>
-      <li><a href="#modal_move_node" tabindex="-1" data-toggle="modal"><i class="fa fa-mail-forward"></i> Move</a></li>
+      
+      <li class="dropdown">
+        <a href="#" class="dropdown-toggle" data-toggle="dropdown" data-behavior="nodes-more-dropdown"><i class="fa fa-ellipsis-h fa-lg"></i></a>
+        <ul class="dropdown-menu dropdown-menu-more">
```
MattBudz: I suggest using `pull-right` in place of `dropdown-menu-more`.

app/services/nodes/merger.rb
```diff
@@ -0,0 +1,72 @@
...
+    @source_node = source_node
+    @target_node = target_node
+  end
+
+  def call
+    Node.transaction do
+      move_descendents
+      reset_counter_caches
+      update_properties
+      move_attachments
```
etdsoft:
```suggestion
      copy_attachments
```

app/services/nodes/merger.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    def update_properties
+      source_node.properties.each do |key, value|
+        target_node.set_property key, value
+      end
+
+      target_node.save
+    end
+
+    def move_attachments
```
etdsoft:
```suggestion
    def copy_attachments
```

app/services/nodes/merger.rb
```diff
@@ -0,0 +1,72 @@
...
+      self.moved_attachments = []
```
etdsoft:
```suggestion
      self.copieded_attachments = []
```

app/services/nodes/merger.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      source_node.attachments.each do |attachment|
+        moved_attachments << attachment.copy_to(target_node)
```
etdsoft:
```suggestion
        copied_attachments << attachment.copy_to(target_node)
```

app/services/nodes/merger.rb
```diff
@@ -0,0 +1,72 @@
...
+      update_properties
+      move_attachments
+
+      Node.destroy(source_node.id)
+    end
+  rescue StandardError => e
+    Rails.logger.error 'Node merge error occured, attempting to rectify attachments.'
+    Rails.logger.error e.backtrace
+
+    undo_attachments_move
```
etdsoft:
```suggestion
    undo_attachments_copy
```

app/services/nodes/merger.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    def move_attachments
+      self.moved_attachments = []
+
+      source_node.attachments.each do |attachment|
+        moved_attachments << attachment.copy_to(target_node)
+      end
+    end
+
+    def undo_attachments_move
```
etdsoft:
```suggestion
    def undo_attachments_copy
```

app/services/nodes/merger.rb
```diff
@@ -0,0 +1,72 @@
...
+      return unless moved_attachments&.any?
```
etdsoft:
```suggestion
      return unless copied_attachments&.any?
```

app/services/nodes/merger.rb
```diff
@@ -0,0 +1,72 @@
...
+      moved_attachments.each(&:delete)
```
etdsoft:
```suggestion
      copied_attachments.each(&:delete)
```

app/views/layouts/snowcrash/_breadcrumb.html.erb
```diff
@@ -23,6 +30,7 @@
 <%= render partial: 'nodes/modals/delete' %>
 <%= render partial: 'nodes/modals/rename' %>
 <%= render partial: 'nodes/modals/move' %>
+<%= render partial: 'nodes/modals/merge' %>
```
etdsoft: :alpha (L31,32,33)

spec/features/node_pages/polling_spec.rb
```diff
@@ -261,7 +261,10 @@ def within_move_node_nodal
   end
 
   def show_move_node_modal
-    click_link "Move" unless move_modal_visible?
+    if !move_modal_visible?
+      find('[data-behavior~=nodes-more-dropdown]').click
+      click_link "Move"
```
etdsoft:
```suggestion
      click_link 'Move'
```

PR #481, opened 2019-08-02
fix project name overflow

MattBudz:
**Spec**
Currently, in a project, the main-sidebar items are hidden/covered if the project name is too long or the browser window is reduced enough. 

**Proposed solution**
Put the navbar brand and nav items into two columns within the same row. Limit the brand column size and hide any overflow


etdsoft:
Seems to be working, however I noticed some responsive quirks:

In `master`:
master-01

In this PR:
pr481-01

break point around width 1276px

PR #483, opened 2019-08-05
Cleanup attachments after lint tests

brianp:
# Spec
Factorybot runs a linter before the suite runs. When the attachment factory is created it creates a file in the attachments folder and leaves it there abandoned. This file can create conflicts with attachment tests in the suite. Ensure it's cleaned up.


brianp:
Closing because the source branch is from a forked repo.

PR #488, opened 2019-08-22
Services Extras Node Merging Bug

brianp:
### Summary

A guard clause was being bypassed due to a ruby string =/= symbol issue. The merge code was written to use the existing set property method. Which was not designed to support the merging of `services` or `services_extras`. Once applied against real world use cases this caused node merge problems that after merging caused nodes to no longer be accessible in the app.

This branch both fixes the initial guard clause bug and introduces a new property merge method which supports the merging on `services` and `service_extras`.

### Other Information
I did a scan of plugins to check over `set_property` usage to see if anything would be impacted by fixing the guard clause. Nothing obvious jumped out.

### Steps to test

1. Create a completely blank Node (Node #1)
2. Create another Node (Node #2) and use the following properties: 
```json
{
  "ip": "127.0.0.1",
  "os": "Linux 4.9",
  "services": [
    {
      "protocol": "tcp",
      "port": 22,
      "state": "open",
      "product": "OpenSSH",
      "name": "ssh",
    }
  ],
  "services_extras": {
    "tcp/22": [
      {
        "source": "nmap",
        "id": "tunnel",
        "output": "n/a"
      }
    ]
  }
}
```
3. Merge the two together with success!


app/models/concerns/node_properties.rb
```diff
@@ -24,9 +24,9 @@ def self.included(base)
   # Sets a property, storing value as Array when needed
   # and taking care of duplications
   def set_property(key, value)
-    if [:services, :services_extras].include?(key) # let's get defensive
+    if [:services, :services_extras].include?(key.to_sym) # let's get defensive
       raise ArgumentError, "don't use set_property for :services or "\
```
etdsoft: Can't you replace these with singles too?

app/models/concerns/node_properties.rb
```diff
@@ -124,4 +124,23 @@ def raw_properties=(value)
   rescue JSON::ParserError => exception
     @raw_properties = value
   end
+
+  def merge_properties(source_properties)
```
etdsoft: If we are going down this route (see next comment), maybe add one of those ugly comment/separator things to split this method from the `raw properties` block.

app/models/concerns/node_properties.rb
```diff
@@ -124,4 +124,23 @@ def raw_properties=(value)
...
+    non_service_properties = source_properties.reject { |k, v| [:services, :services_extras].include?(k.to_sym) }
+
+    non_service_properties.each do |key, value|
+      set_property key, value
+    end
+
+    properties[:services] = [properties[:services], source_properties[:services]].flatten.uniq.reject(&:blank?)
```
etdsoft: I can't say I like this approach more than the elegant (albeit) broken version that we had before (i.e. just going through `each` and calling `set_propety`.

Is there a middle ground we can live with?

```
   source_properties.each do |key, value|
     if [:services, :services_extras].includes?(key.to_sym)
       value.each do |data|
         set_service data
       end
     else
      set_property key, value
     end
  end
```

I'm not against having a `merge_properties` method in the model (although I'm not against having the logic in the controller). But it seems that the responsibility of ensuring de-duplication at the service / service_extras level should belong in `set_service`?

I think that code along those lines would be easier to understand.

The code we have here, I don't doubt it works, but it's somewhat difficult to follow intuitively?

The relationship between :services and :services_extra is interesting (in a bad way). A plugin may call `set_service` and the method may decide to add :services_extra too. Maybe calling `set_service` with `each` :services_extra` value is idempotent already?

brianp: > The relationship between :services and :services_extra is interesting (in a bad way). A plugin may call set_service and the method may decide to add :services_extra too. Maybe calling set_service with each :services_extra` value is idempotent already?

This is where it did all go south.

The thing about merging the properties is calling `set_service` from another nodes properties is done at time when `services`, and `services_extras` no longer relate at the iterator level. Calling `set_service` with `services_extras` is simply an improper data format, as it lacks any service information.

Maybe the route would be a `set_services_extras` method and a switch from set properties to handle all 3 unique types.

- standard properties
- services
- services_extras

brianp: I did find a nicer way to pass service extras through `set_service` like we discussed. What I've found though is an inconsistent behaviour of our `set_` methods, which makes it hard to re-use `set_service` with actual services. The current behaviour may not be what's desired though so:

When using `set_property` if the property being set already exists, then we expand the value to contain multiple values, removing duplicates.

When using `set_service` if the service already exists, identified by port and protocol as the key then we *update* the existing service with any other attributes. 
The agreed pattern for merging services was that in the event of services that already exist based on port and protocol we want to allow having duplicates in the event the additional data is different.
This becomes impossible to do via `set_service` because if we attempt to pass a duplicate node & protocol with different information `set_service` will identify this as an update.

The inconsistency here is between `set_property` and `set_service`. One adds new values, the other tries to update existing values.
I propose we make them similar. Have `set_service` accept new values with existing port and protocol, with other fields. Remove true dups (all fields the same). And remove the ability to *update* existing services.

The questions is, will removing the ability to update an existing service break existing usage in anyway?

aapomm: @brianp I'm all for consistency, but I'm worried that the `set_service` change you are suggesting will impact the other parts of the app. If this will cascade to a bunch of other features, I suggest we keep it simple for this PR, and add a separate task for the `set_service` consistency refactor. Otherwise, the refactor sounds good to me.

app/models/concerns/node_properties.rb
```diff
@@ -67,7 +67,7 @@ def has_any_property?
   #
   # Anything else will be saved under the `services_extras` key
   def set_service(data)
-    data.symbolize_keys!
+    data = data.symbolize_keys
```
dormi: is this for better reading or this was not working?
should we indent the `=` in this line to match the next lines?

dormi: I see the commit message now `HashWithIndifferentAccess doesn't have the bang operatore symbolize keys.`

PR #489, opened 2019-08-26
Port kits to CE

dormi:
### Summary

Currently we have a `setup:welcome` task in CE.
We also have a `setup:kit` task in Pro. This task extracts a zip file with a specific file tree and applies it.
 
The CE task imports a bunch of templates under `lib/tasks/templates`.
The Pro task imports a bunch of templates from a specified zip file.

We want to port the Pro task (`setup:kit`) to CE.

**Proposed solution**
In  CE, rename the `setup:welcome` to something like `setup:kit:welcome`.
This renamed task will first zip the contents of `lib/task/templates`, then call the ported `setup:kit` to apply them.


app/models/log.rb
```diff
@@ -7,13 +7,14 @@ def set_uid
 
   def write(trace = nil, &block)
     text = trace.nil? ? yield : trace
-    Log.create!(attributes.merge(text: text))
+    Log.create!(attributes.except('id').merge(text: text))
```
dormi: all changes in this model are just porting the Pro version to CE
The Pro version fixes this case:
```
l = Log.new.write('hello')
l.write 'hello 2' # raises unique id error
```

app/jobs/kit_import_job.rb
```diff
@@ -0,0 +1,125 @@
...
+    }.each do |plugin|
+      reports_dir = "#{temporary_dir}/kit/templates/reports/#{plugin}"
+      FileUtils.cp_r(
+        "#{reports_dir}/.",
+        "#{report_templates_dir}/#{plugin}/"
+      ) if File.exist?(reports_dir)
+    end
+  end
+
+  def import_report_template_properties;end
```
aapomm: I'm not a fan of removing the functionality in this method for the CE version. This may cause some issues when syncing this to Pro. Can we instead include the functionality here and in line 21 we have something like:
```
import_report_template_files if defined?(Dradis::Pro)
```
This way we'd have the same versions in CE and Pro and merging will be easier.

app/jobs/kit_import_job.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  def import_rules;end
```
aapomm: see above comment

app/models/configuration.rb
```diff
@@ -30,11 +30,26 @@ def self.max_deleted_inline
...
+    create_with(value: Rails.root.join('../../shared/templates/notes/').to_s)
+      .find_or_create_by(name: 'admin:paths:note_templates').value
+  end
+
   def self.paths_templates_plugins
     create_with(value: Rails.root.join('templates', 'plugins').to_s)
       .find_or_create_by(name: 'admin:paths:templates:plugins').value
   end
 
+  def self.paths_templates_projects
```
aapomm: Are we even using this directory for CE outside of the welcome task?

lib/tasks/thorfile.rb
```diff
@@ -120,38 +120,34 @@ def seed
...
-      if Methodology.pwd.exist?
-        say 'Methodology templates folder already exists. Skipping.'
-      else
-        template 'methodology.xml', Methodology.pwd.join('owasp2017.xml')
+      # zip lib/tasks/welcome_kit
+      directory_to_zip = Rails.root.join('lib', 'tasks', 'welcome_kit')
+      temporary_zip = Tempfile.new
+
+      entries = Dir["#{directory_to_zip}/**/**"] - ['.', '..']
+      Zip::File.open(temporary_zip.path, Zip::File::CREATE) do |zipfile|
```
aapomm: Zipping a file and consequently unzipping it seems clunky to me. Can we have the welcome templates ready as a `.zip` file already? For example, all the templates that we need are found in: `lib/tasks/welcome.zip`.

If that's not possible, another solution is to allow the `setup:kit` task to handle directories instead of a zip file using an option.

spec/jobs/kit_import_job_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      )
+      allow(Methodology).to receive(:pwd).and_return(
+        Pathname.new(Configuration.paths_templates_methodologies)
+      )
+
+      described_class.new.perform(file: file, logger: Log.new.write('Testing...'))
+    end
+
+    after(:all) do
+      Rails.root.join("tmp", "rspec")
```
aapomm:
```suggestion
      Rails.root.join('tmp', 'rspec')
```

spec/jobs/kit_import_job_spec.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe KitImportJob do
+  describe '#perform' do
+
+    before(:each) do
+      file     = File.new(Rails.root.join('spec', 'fixtures', 'files', 'templates', 'kit.zip'))
+      tmp_dir = Rails.root.join("tmp", "rspec")
```
aapomm:
```suggestion
      tmp_dir = Rails.root.join('tmp', 'rspec')
```

lib/tasks/thorfile.rb
```diff
@@ -120,38 +120,34 @@ def seed
       puts "[  DONE  ]"
     end
 
+    desc "kit SETUP_PACKAGE", "import files and projects from a specified kit file"
```
aapomm:
```suggestion
    desc 'kit SETUP_PACKAGE', 'import files and projects from a specified kit file'
```

lib/tasks/thorfile.rb
```diff
@@ -120,38 +120,34 @@ def seed
...
+    method_option :file, :type => :string, :desc => "full path to a zip file"
```
aapomm:
```suggestion
    method_option :file, :type => :string, :desc => 'full path to a zip file'
```

lib/tasks/thorfile.rb
```diff
@@ -120,38 +120,34 @@ def seed
...
+    def kit(file)
+      puts "** Importing package..."
```
aapomm:
```suggestion
      puts '** Importing package...'
```

lib/tasks/thorfile.rb
```diff
@@ -120,38 +120,34 @@ def seed
...
+      KitImportJob.perform_now(file: file, logger: default_logger)
+      puts "[  DONE  ]"
```
aapomm:
```suggestion
      puts '[  DONE  ]'
```

lib/tasks/thorfile.rb
```diff
@@ -120,38 +120,34 @@ def seed
...
+    end
+
     desc "welcome", "adds initial content to the repo for demonstration purposes"
```
aapomm:
```suggestion
    desc 'welcome', 'adds initial content to the repo for demonstration purposes'
```

app/models/log.rb
aapomm: Can you explain this more since `Log.new.write('hello')` seems to work for both CE and Pro on my end. Also, how is this related to the PR?

dormi: If you apply the lines above together (`write` on the result of a `write`), you should see an error in CE.
Here we are just porting the Log model code from Pro to CE, since that problem was fixed in Pro before but forgot to port it to CE.
We need it fixed in this PR because we are passing to the bg job as a logger the result of a `write`:  https://github.com/dradis/dradis-ce/pull/489/files#diff-5d55ddbe9a8496e5b77d56dbb8004988R34 (to keep consistency with Pro)

app/models/configuration.rb
dormi: No, I don't think so.
But it makes our Pro and CE codes converge. The `configuration` model now looks the same both in `Pro` and `CE`.

app/jobs/kit_import_job.rb
dormi: I have a valid point here but I am unsure about this change.
We would be referencing a class that does not exist in the codebase (`ReportTemplateProperties`).
My idea was that keeping the method but making it a no-op would be clear enough at merge time.

lib/tasks/thorfile.rb
dormi: I did it this way by definition of the task 
> Then setup:welcome task will first zip the contents of lib/task/templates, then call the ported setup:kit to apply them
We had agreed on that somewhere.

The good thing of zipping it is that we have the welcome kit unzipped on `lib/tasks/weilcome_kit`, and its easier to see changes using git.

app/jobs/kit_import_job.rb
aapomm: I just realized are are already doing the defined check for `Dradis::Pro`, so I was thinking that we were planning this approach beforehand.

My main concern here is the consistency since in `dradis-plugins` we do already something similar. In [content_service/base.rb](https://github.com/dradis/dradis-plugins/blob/master/lib/dradis/plugins/content_service/base.rb), we do a defined check for `Dradis::Pro`:
```
    include Boards        if defined?(Dradis::Pro)
    include Categories
    include ContentBlocks if defined?(Dradis::Pro)
    include Evidence
    include Issues
    include Nodes
    include Notes
    include Properties    if defined?(Dradis::Pro)
```
and in those modules we are referencing classes that do not exist in the CE codebase too.

lib/tasks/thorfile.rb
brianp: > We had agreed on that somewhere.
I'm late to the party on this. If it is what we decided to do I think it's fine to leave.

I don't understand though- why the repo already has access to all the resources it needs for the kit but can't utilize them in their current form. Creating the kit, and then unpacking the kit does seem like a weird process.

I do get the idea of utilizing the process of kit unpacking because it already exists and is easier to apply. Just seems weird we have all the resources needed to do it already present. If we're committing them, why not also commit the package?

app/jobs/kit_import_job.rb
```diff
@@ -0,0 +1,125 @@
...
+    @temporary_dir = Dir.mktmpdir
+
+    FileUtils.cp file, temporary_dir
+    unzip
+    import_methodology_templates
+    import_note_templates
+    import_project_package
+    import_project_templates
+    import_report_template_files
+    if defined?(Dradis::Pro)
```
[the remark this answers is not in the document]
dormi: yes

app/jobs/kit_import_job.rb
dormi: Ok, restored these methods

PR #490, opened 2019-08-28
Pass params the usual way when redirecting to export plugin controllers

dormi:
### Summary

When we redirect to the export plugin controller, we were using the session to pass params.
But they can be passed as normal params too.

In the case of the the html_export plugin, the rendered html opens in a new tab, which is nice.
But problems arise if we have another tab opened with another project html exported.
When we refresh the browser tabs, since we are getting the params from the session, we always render using the last export params we stored in the session. 
And that could mean rendering a different project with a different template.

Passing params as normal params fixes this.

### Copyright assignment


CHANGELOG
```diff
@@ -88,6 +89,7 @@ v3.15 (November 2019)
    - Medium: Authenticated author mentioning an existing user outside of the
      project will subscribe that user to the note/issue/evidence
    - Upgraded gems: nokogiri (CVE-2019-13117)
+>>>>>>> master
```
brianp, line 92: Merge conflict present.

```suggestion
```

CHANGELOG
```diff
@@ -4,6 +4,7 @@ v3.17 (Month 2020)
     - Activity Feed now shows board name and link
     - No mandatory due date
   * Comments can now have Textile markup
+  * Export plugins also get template as a request param
```
brianp, line 7: This entry doesn't really explain what the change is

```suggestion
  * Prevent HTML exports from rendering with the wrong template when multiple tabs are open
```

aapomm:
Implemented for v4.8.0

PR #494, opened 2019-09-10
Port Boards: db tables and models

dormi:
### Specs
Port the db migrations so we can store boards, lists and cards in the database.
Port the models (board, list, card and related files) required to operate with these tables.
Make sure we port the model specs and related files (factories...) to make the specs go green.

### How to test

- Specs should go green
- We should be able to open the console and run successfully this code:
```
b = Board.create!(name: "Test board", node: Project.new.methodology_library)
l = List.create!(name: "Test list", board: b)
c = Card.create!(name: "Test card", list: l)
```

### Copyright assignment

### Check List

### DIFF
--- a/app/models/board.rb
+++ b/app/models/board.rb
@@ -21,7 +21,6 @@ class Board < ApplicationRecord
   def project=(new_project); end
 
   # -- Callbacks ------------------------------------------------------------
-  after_save :set_default_node
 
   # -- Validations ----------------------------------------------------------
   validates :name, presence: true, length: { maximum: DB_MAX_STRING_LENGTH }
@@ -45,10 +44,6 @@ class Board < ApplicationRecord
   end
 
   private
-
-  def set_default_node
-    self.node = Project.find(1).methodology_library
-  end
 
   def validate_node_has_no_other_board
     if self.node.try(:type_id) != Node::Types::METHODOLOGY \
--- a/app/models/board.rb
+++ b/app/models/board.rb
@@ -1,0 +1,59 @@
+class Board < ApplicationRecord
+  include ActsAsLinkedList
+  list_item_name :list
+
+  # -- Relationships --------------------------------------------------------
+  belongs_to :node
+  has_many :activities, as: :trackable
+  # NOTE: Rails >= 5.1 raises error when a `has_many :y, through: x`
+  # relationship is defined before the corresponding `has_many :x` relationship.
+  # (https://github.com/rails/rails/pull/27485).
+  # So we define `lists` before `cards`
+  has_many :lists, dependent: :destroy
+  has_many :cards, through: :lists
+
+  def project
+    # dummy project; this makes Boards's interface more similar to how it is
+    # in Pro and makes it easier to deal with board in URL helpers
+    @project ||= Project.new
+  end
+
+  def project=(new_project); end
+
+  # -- Callbacks ------------------------------------------------------------
+  after_save :set_default_node
+
+  # -- Validations ----------------------------------------------------------
+  validates :name, presence: true, length: { maximum: DB_MAX_STRING_LENGTH }
+  validates :project, presence: true
+  validate  :validate_node_has_no_other_board
+
+  # -- Scopes -----------------------------------------------------------------
+
+  # -- Class Methods ----------------------------------------------------------
+
+  # -- Instance Methods -------------------------------------------------------
+  def recovered_list
+    list = self.lists.find_by(name: 'Recovered')
+    if !list
+      list = self.lists.new(name: 'Recovered')
+      list.previous_id = self.last_list.try(:id)
+      list.save
+    end
+
+    list
+  end
+
+  private
+
+  def set_default_node
+    self.node = Project.find(1).methodology_library
+  end
+
+  def validate_node_has_no_other_board
+    if self.node.try(:type_id) != Node::Types::METHODOLOGY \
+    && Board.where.not(id: self.id).where(node_id: self.node_id).any?
+      errors.add(:node, 'already has a board')
+    end
+  end
+end
--- a/app/models/card.rb
+++ b/app/models/card.rb
@@ -1,0 +1,91 @@
+class Card < ApplicationRecord
+  include Commentable
+  include HasFields
+  include Subscribable
+
+  dradis_has_fields_for :description
+
+  has_paper_trail on: [:destroy]
+
+  # -- Relationships --------------------------------------------------------
+  belongs_to :list, touch: true
+  belongs_to :previous_card,
+             foreign_key: :previous_id,
+             class_name: 'Card',
+             optional: true
+  has_and_belongs_to_many :assignees, class_name: 'User'
+  has_many :activities, as: :trackable
+
+  delegate :project, to: :list
+  delegate :board, to: :list
+
+  # -- Callbacks ------------------------------------------------------------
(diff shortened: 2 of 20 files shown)


app/models/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  belongs_to :node
+  has_many :activities, as: :trackable
+  # NOTE: Rails >= 5.1 raises error when a `has_many :y, through: x`
+  # relationship is defined before the corresponding `has_many :x` relationship.
+  # (https://github.com/rails/rails/pull/27485).
+  # So we define `lists` before `cards`
+  has_many :lists, dependent: :destroy
+  has_many :cards, through: :lists
+
+  def project
```
dormi: these 2 methods are the new thing in CE. And there is no `belongs_to :project` in CE.

app/models/card.rb
```diff
@@ -0,0 +1,91 @@
+class Card < ApplicationRecord
```
dormi: No CE differences

app/models/concerns/acts_as_linked_list.rb
```diff
@@ -0,0 +1,64 @@
+module ActsAsLinkedList
```
dormi: No CE differences

app/models/list.rb
```diff
@@ -0,0 +1,36 @@
+class List < ApplicationRecord
```
dormi: No CE differences

app/models/node.rb
```diff
@@ -18,10 +18,11 @@ module Types
   acts_as_tree counter_cache: true, order: :label
 
   # -- Relationships --------------------------------------------------------
-  has_many :notes, dependent: :destroy
+  has_many :activities, as: :trackable
```
dormi: Reordered to be like in Pro
Added `has_many :boards`

db/migrate/20161114111520_add_project_id_to_versions.rb
```diff
@@ -1,4 +1,4 @@
-class AddProjectIdToVersions < ActiveRecord::Migration[5.0]
+class AddProjectIdToVersions < ActiveRecord::Migration[5.1]
```
dormi: We changed the ported migrations to 5.1, as we did with the rest of migrations long ago. But this one was still in 5.0. Ported too.

db/migrate/20170210085327_create_boards.rb
```diff
@@ -0,0 +1,10 @@
+class CreateBoards < ActiveRecord::Migration[5.1]
```
dormi: Ported to 5.1

db/migrate/20170210085327_create_boards.rb
```diff
@@ -0,0 +1,10 @@
...
+  def change
+    create_table :boards do |t|
+      t.string :name
+      t.references :node, foreign_key: { on_delete: :cascade }
```
dormi: Instead of adding this in  a later migration like we had to do in Pro, here we can add this reference when creating boards.
This line replaces: `t.integer :project_id, index: true`

db/migrate/20170210092702_create_lists.rb
```diff
@@ -0,0 +1,11 @@
+class CreateLists < ActiveRecord::Migration[5.1]
```
dormi: Ported to 5.1, no other differences with Pro

db/migrate/20170210092712_create_cards.rb
```diff
@@ -0,0 +1,13 @@
+class CreateCards < ActiveRecord::Migration[5.1]
```
dormi: Ported to 5.1, no other differences

db/migrate/20170210102351_create_join_table_card_user.rb
```diff
@@ -0,0 +1,8 @@
+class CreateJoinTableCardUser < ActiveRecord::Migration[5.1]
```
dormi: Ported to 5.1, no other differences

db/schema.rb
```diff
@@ -23,6 +23,33 @@
     t.index ["trackable_id", "trackable_type"], name: "index_activities_on_trackable_id_and_trackable_type"
   end
 
+  create_table "boards", force: :cascade do |t|
```
dormi: The schema.rb has some changes a part from including the new tables.
Some columns are reverted from `bigint` to `integer`.
But this happens if we generate the scheme with `database.yml` pointing to sqlite.
And since in CE we are using sqlite, seems like the correct thing to do ?
CI seems ok with this change.

spec/factories/boards.rb
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
```
dormi: No differences

spec/factories/cards.rb
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
```
dormi: No differences

spec/factories/lists.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
```
dormi: No differences

spec/models/board_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
```
dormi: Removes these lines:
```
it { should belong_to(:project).touch(true) }
it { should validate_presence_of(:project) }
```

spec/models/card_spec.rb
```diff
@@ -0,0 +1,94 @@
+require 'rails_helper'
```
dormi: No differences

spec/models/concerns/acts_as_linked_list_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
```
dormi: No differences

spec/models/list_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
```
dormi: No differences

spec/models/node_spec.rb
```diff
@@ -106,7 +107,7 @@
     end
   end
 
-  it 'uses a default type ID if none provieded' do
+  it 'uses a default type ID if none provided' do
```
dormi: A typo that is already fixed in Pro

spec/support/linked_list_shared_examples.rb
```diff
@@ -0,0 +1,37 @@
+# This shared_example tests the move method for linked list items.
```
dormi: No differences

app/models/board.rb
```diff
@@ -0,0 +1,54 @@
...
+      list.previous_id = self.last_list.try(:id)
+      list.save
+    end
+
+    list
+  end
+
+  private
+
+  def validate_node_has_no_other_board
```
brianp: Is keeping this just to keep the CE code close to the Pro code?

dormi: I missed this one!
What to do? 🤔

dormi: I say let's keep it.

spec/models/board_spec.rb
etdsoft: can you please add a reminder Task to ensure that after ce-sync, we don't loose these two in Pro?

PR #495, opened 2019-09-11
Port boards: show board

dormi:
### Spec
In this url: /projects/1/boards/[board_id] we should be able to see a board.
If we:
- create it using the console like this:
`b = Board.create!(name: "Test board", node: Project.new.methodology_library)`
- note down the [board_id] 
- visit _/projects/1/boards/[board_id]_

we should be able to see the board, even if managing lists and cards inside it still does not work.

### How to test
Create a board using Rails console:
```
b = Board.create!(name: "Test board", node: Project.new.methodology_library)
l = List.create!(name: "Test list", board: b)
c = Card.create!(name: "Test card", list: l)
```
Visit the board in _/projects/1/boards/[board_id]_ and assert that the board displays properly.

### Copyright assignment

### Check List

### DIFF
--- a/app/controllers/boards_controller.rb
+++ b/app/controllers/boards_controller.rb
@@ -1,0 +1,123 @@
+class BoardsController < AuthenticatedController
+  include ActivityTracking
+  include ProjectScoped
+
+  before_action :set_board, only: [:show, :edit, :update, :destroy]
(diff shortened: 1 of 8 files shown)


dormi:
closing as this has been included in https://github.com/dradis/dradis-ce/pull/496

PR #496, opened 2019-09-11
Port Boards: Boards

dormi:
### Spec
We should be able, when clicking on Methodologies link on the sidebar, to see an index with our boards (or empty list if none). From there we should be to create a board, rename it, delete it.

**Proposed solution**
Port from Pro the desired functionality

### How to test

- Log into Dradis.
- Assert you can create a board.
- Assert you can rename it
- Assert you can delete it 

### Copyright assignment

### Check List


app/models/node.rb
```diff
@@ -24,6 +24,10 @@ module Types
   has_many :issues, -> { distinct }, through: :evidence
   has_many :notes, dependent: :destroy
 
+  def boards
```
dormi: we have `has_many` in line 22. We don't need this method?

app/models/project.rb
```diff
@@ -47,6 +47,10 @@ def tags
     Tag.all
   end
 
+  def boards
```
dormi: Alpha order these public instance methods

app/services/methodology_migration_job.rb
```diff
@@ -0,0 +1,38 @@
+class MethodologyMigrationJob < ApplicationJob
```
dormi: hey! this should be under `app/jobs` ?

app/services/methodology_migration_service.rb
```diff
@@ -0,0 +1,102 @@
...
+    @project = Project.find(project_id)
+  end
+
+  def migrate(methodology, board_name: nil, node: nil)
+    ActiveRecord::Base.transaction do
+      # create a board for each methodology
+      board = Board.new(
+        name: methodology.name,
+        node: node || project.methodology_library
+      )
```
dormi: I am not 100% sure we must remove the `project_id: project_id` line here.
If we replace it with `project: @project`, the line will be the same in both codebases.

app/services/methodology_migration_service.rb
```diff
@@ -0,0 +1,102 @@
...
+  # Returns 2 arrays, one of "pending" cards (not checked)
+  # and one of "done" cards (checked).
+  def extract_cards(methodology)
+    pending = []
+    done    = []
+    methodology.sections.each do |section|
+      section.tasks.each do |task|
+        # create a card for each task on the methodology
+        card = Card.new
+        card.name = "[#{section.name}] #{task.name}".truncate(Card.columns_hash['name'].limit || 256)
```
dormi: This `256` is new.
Is it a sqlite requirement yes? Because `Card.columns_hash['name'].limit` is `nil` in CE ?

config/routes.rb
```diff
@@ -53,6 +53,15 @@
       end
     end
 
+    resources :boards do
```
dormi: place this in alpha order, it should go before `comments`
It would be in the same location as in Pro

spec/features/board_pages/board_pages_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'rails_helper'
```
dormi: This file seems quite different form Pro.
And maybe it should be more similar or ideally the same file.

Example: this code also works in CE: `create(:board, node: current_project.methodology_library, project: current_project)`

app/views/activities/_poll_methodologies.js.erb
```diff
@@ -0,0 +1,51 @@
+<%# Card activities %>
```
dormi: We are porting this, but never calling this partial. 
It should be from `activities/poll.js.erb`
But I think this belongs to another PR.

app/models/methodology.rb
```diff
@@ -39,7 +43,7 @@ class Methodology
   # 'admin:paths:templates:methodologies' setting
   def self.pwd
     @pwd ||= begin
-      conf = Configuration.create_with(value: Rails.root.join('templates/methodologies/').to_s).
+      conf = Configuration.create_with(value: Rails.root.join('../../shared/templates/methodologies/').to_s).
```
etdsoft: @brianp @dormi (!)

```suggestion
      conf = Configuration.create_with(value: Rails.root.join('templates/methodologies/').to_s).
```

spec/features/board_pages/board_pages_spec.rb
etdsoft: how is this being tracked so it doesn't slip if the PR is merged?

app/models/methodology.rb
brianp: It's weird that it works without this change 🤔

spec/features/board_pages/board_pages_spec.rb
brianp: > how is this being tracked so it doesn't slip if the PR is merged?

Notes/tasks to make sure in the next scope where they become appropriate they get re-added. It's an imperfect solution.

In this particular case the next scope already existed and we made sure the port existed in the next PR.

app/models/methodology.rb
brianp: I tested it and with the change everything still (continues?) to work as expected.

PR #497, opened 2019-09-11
Port Boards: Lists

brianp:
**Spec**
We should be able, when viewing a board see a list of our lists, with the ability to create, rename, remove, and move the lists.

**Proposed solution**
Port from Pro the desired functionality

**How to test**

    Log into Dradis.
    Assert when you view a board you can see it's lists
    Assert you can rename a list
    Assert you can delete a list
    Assert you can move a list, and after a refresh the list retains it's position

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -12,6 +12,10 @@ class Project
   end
 
   # -- Instance Methods -----------------------------------------------------
+  def authors
+    User.all
+  end
+
   def initialize(id: 1, name: 'Dradis CE', **_attrs)
     @id   = id
     @name = name
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,4 +1,5 @@
 class User < ApplicationRecord
+  alias_attribute :name, :email
   # -- Relationships --------------------------------------------------------
   has_many :activities
   has_many :comments, dependent: :nullify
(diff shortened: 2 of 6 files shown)


app/models/project.rb
```diff
@@ -12,6 +12,10 @@ def self.find(id)
   end
 
   # -- Instance Methods -----------------------------------------------------
+  def authors
```
dormi: also present in https://github.com/dradis/dradis-ce/pull/498#pullrequestreview-287401164

app/models/user.rb
```diff
@@ -1,4 +1,5 @@
 class User < ApplicationRecord
+  alias_attribute :name, :email
```
dormi: also present in https://github.com/dradis/dradis-ce/pull/498#pullrequestreview-287401164 (whit extra white line)

etdsoft: I'm not convinced about this @brianp @dormi 

We've a problem w/ :name vs :email discrepancy that we need to address or keep making worse (like with this change).

Before we had issues with Comments and using the "login" in CE, but the "email" in Pro for @- mentions.

I think it's worth deciding here instead of just adding to the spaghetti ball.

brianp: It was "simple" enough to manage without the alias. A couple tests need to change, but I think functionally everything else I've seen falls into place.
I'll do another pass to see if anything goes wrong using email instead of name.

PR #504, opened 2019-09-13
Port Boards: Activity polling

dormi:
### Spec
When using boards/lists/cards and another user modifies them, we should see those changes.

### How to test
Log in with user A in one browser. Log in with user B in another browser.

**In the boards index with both users (/projects/1/boards):**
Add a board with one user, the other must see it.
Rename the board with one user, the other must see it.
Delete the board with one user, the other must see it.

**Inside the same board with both users (/projects/1/boards/X):**
Add a list with one user, the other must see it.
Rename the list with one user, the other must see it.
Delete the list with one user, the other must see it.
Add a task with one user, the other must see it.
Rename the task with one user, the other must see it.
Delete the task with one user, the other must see it.
Delete the board with 1 user, the other must see a warning

**Inside the same card with both users (/projects/1/boards/X/lists/Y/cards/Z):**
Edit the card with one user, the other must see it.
Delete the task with one user, the other must see a warning.

### Check List

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -4,6 +4,8 @@ class Activity < ApplicationRecord
 
   belongs_to :trackable, polymorphic: true, required: false
 
+  def project=(new_project); end
+
   # NOTE: when the project importer creates activities, it will try to match
   # them to existing users based on the 'user email' field in the XML. If it
   # can't find any users with the given address, it will save user_id as '-1'.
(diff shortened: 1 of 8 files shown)


app/models/activity.rb
```diff
@@ -4,6 +4,8 @@ class Activity < ApplicationRecord
 
   belongs_to :trackable, polymorphic: true, required: false
 
+  def project=(new_project); end
```
dormi: We may be using it in nodes, boards, activities in CE. Sounds like a possible model concern. Added a [NICE TO HAVE] task for it.

PR #508, opened 2019-09-16
change jquery selector

MattBudz:
**Spec**
Currently, if there is a long list of issues and the collapse chevron is clicked, the page jumps down.

This PR fixes the jump which was caused by focusing on the import box element every time any chevron is clicked.

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/assets/javascripts/snowcrash/behaviors.js.coffee
+++ b/app/assets/javascripts/snowcrash/behaviors.js.coffee
@@ -179,5 +179,5 @@ document.addEventListener "turbolinks:load", ->
     $('[data-behavior~=collapse-collection]').click ->
       $this = $(this)
       $this.find('[data-behavior~=toggle-chevron]').toggleClass('fa-chevron-down fa-chevron-up')
-      if $('[data-behavior~=import-box]').length
+      if $this.hasClass('import-toggle')
         $('[data-behavior~=import-box]').find("input[type='text']:first").focus()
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/snowcrash/behaviors.js.coffee
```diff
@@ -179,5 +179,5 @@ document.addEventListener "turbolinks:load", ->
     $('[data-behavior~=collapse-collection]').click ->
       $this = $(this)
       $this.find('[data-behavior~=toggle-chevron]').toggleClass('fa-chevron-down fa-chevron-up')
-      if $('[data-behavior~=import-box]').length
+      if $this.hasClass('import-toggle')
```
etdsoft: Why is this working?

I don't love going from `data-behavior` to CSS classes for this. Is it because the 'import-box' element is hidden? Then maybe we can add another data-behavior to the toggler, the reason we're using the `~=` comparison in `data-behavior` is because we can stack several into one element and compare them piece meal when needed.

MattBudz: @etdsoft  

`$('[data-behavior~=import-box]').length` is always true because it exists on the page but is hidden, so when any chevron is clicked it always jumps focus to the import box's first text box even if it is not shown. This change checks to see if `$this` has a class `import-toggle` which is unique to the import-box chevron. We have to check if `$this` is related to the import box. Alternatively, we can check if `$this` which was clicked, has a data-target attribute with the value of `#import-box`. I used `if $this.data('target') == '#importbox'` as well as `if $this.attr('data-target') == '#importbox'` but neither of those use the `~=` comparison so I used `if $this.hasClass('import-toggle')` because it was a shorter line of code. 

Maybe there is a different approach here that I am missing?

Simply selecting an element with the `data-behavior` attribute set to `import-box` or anything else won't work because that always exists. The key is to check if `$this` has anything unique to the `import-box` so it only focuses when that specific chevron is clicked.

PR #510, opened 2019-09-16
Default opened tab

aapomm:
### Specs
This is specially true when destroying. Users sometimes want keep selectively deleting evidence, and after each deletion they have to go click again on the evidence tab.

**Proposed solution** 
Add a way in JS to open a default tab when a URL has #<asdf>-tab much like how anchors work.

### How to test:

1. Create an issue and a node
2. Visit the issue and open the evidence tab
3. Create multiple evidences and confirm that you are routed back with the evidence tab opened.
4. Update an evidence and confirm that you are routed back with the evidence tab opened.
5. Delete the evidence and confirm that you are routed back with the evidence tab opened.

### Check List

- [x] Added a CHANGELOG entry


app/views/issues/show.html.erb
```diff
@@ -15,18 +15,26 @@
 </ul>
 
 <ul class="nav nav-tabs main-tabs">
-  <li class="active">
+  <li <%= "class=active" if !params[:tab] || params[:tab] == 'info-tab' %>>
```
dormi:
```suggestion
  <li<%= " class=active" if !params[:tab] || params[:tab] == 'info-tab' %>>
```

dormi: ^ silly suggestion to avoid extra white space in out generated html:
Captura de pantalla 2019-09-27 a las 11 23 55

app/views/issues/show.html.erb
```diff
@@ -15,18 +15,26 @@
...
     <a href="#info-tab" data-toggle="tab">
       <%= colored_icon_for_model(@issue, 'fa-bug') %> Information
     </a>
   </li>
-  <li><a href="#evidence-tab" data-toggle="tab"><i class="fa fa-laptop"></i> Evidence <span class="badge"><%= @issue.evidence.count %></span></a></li>
-  <li><a href="#activity-tab" data-toggle="tab"><i class="fa fa-refresh"></i> Recent activity</a></li>
+  <li <%= "class=active" if params[:tab] == 'evidence-tab' %>>
```
dormi:
```suggestion
  <li<%= " class=active" if params[:tab] == 'evidence-tab' %>>
```

app/views/issues/show.html.erb
```diff
@@ -15,18 +15,26 @@
...
+    <a href="#evidence-tab" data-toggle="tab">
+      <i class="fa fa-laptop"></i> Evidence <span class="badge"><%= @issue.evidence.count %></span>
+    </a>
+  </li>
+  <li <%= "class=active" if params[:tab] == 'activity-tab' %>>
```
dormi:
```suggestion
  <li<%= " class=active" if params[:tab] == 'activity-tab' %>>
```

dormi:
Wait, I huso remembered this: https://github.com/secroots/dradis-pro/pull/707#discussion_r304795067

app/controllers/evidence_controller.rb
```diff
@@ -122,11 +122,14 @@ def destroy
           # 2. from the node evidence tab
           # 3. from the evidence show page itself (under node)
           # When using redirect_back in case 3, we find an evidence not found error,
-          # since the evidence does not exist anymore. That's why we check the 'Referer' here:
-          if request.headers['Referer'] == project_node_evidence_url(current_project, @node, @evidence)
+          # since the evidence does not exist anymore. That's why we check the referrer here.
```
dormi: Aaron, I understand this code is to keep showing the evidence tab when destroying evidence.

When destroying from nodes evidence tab, seems like we were not doing it, even Iif we redirected when creating/updating.

Maybe that decision was taken to avoid complicating this too much?
If we cannot find a simper solution, I would just leave this as it was and avoid returning to evidence tab on destroy

I think this is a case of https://stackoverflow.com/questions/51054260/rails-5-redirect-back-with-url-query-parameter-option

spec/features/issues_spec.rb
```diff
@@ -306,14 +306,25 @@
...
 
-          it 'creates an evidence with the selected template for selected node' do
-            find('.js-add-evidence').click
-            check('192.168.0.1')
-            select('Simple Note', from: 'evidence_content')
-            expect { click_button('Save Evidence') }.to change { Evidence.count }.by(1)
-            evidence = Evidence.last
-            expect(evidence.content).to eq(NoteTemplate.find('simple_note').content.gsub("\n", "\r\n"))
-            expect(evidence.node.label).to eq '192.168.0.1'
+          context 'when adding an evidence through the evidence tab' do
```
dormi: should we add a spec for the update and destroy cases? 🤔

app/controllers/evidence_controller.rb
```diff
@@ -122,11 +122,14 @@ def destroy
...
+          referrer_path = URI(request.referrer).tap { |u| u.query = nil }.to_s
```
dormi: can you explain this code?

app/controllers/evidence_controller.rb
```diff
@@ -122,11 +122,14 @@ def destroy
...
+          if !referrer_path || referrer_path == project_node_evidence_url(current_project, @node, @evidence)
             redirect_to project_node_path(current_project, @node), notice: notice
+          elsif referrer_path == project_node_path(current_project, @node)
+            redirect_to referrer_path, notice: notice
           else
-            redirect_back fallback_location: project_node_path(current_project, @node), notice: notice
+            redirect_to "#{referrer_path}?tab=evidence-tab", notice: notice
```
dormi: if we need to redirect_to, let's try not no build the url ourselves. Can we use `url_for` ?

app/controllers/evidence_controller.rb
aapomm: This deletes the query part of the URL. This is one line version of:
```
uri = URI(request.referrer)
uri.query = nil
uri.to_s
```

app/controllers/evidence_controller.rb
aapomm: I agree, but the `referrer_path` that we get from the request is a string. I'm open to ideas!

app/controllers/evidence_controller.rb
aapomm: > When destroying from nodes evidence tab, seems like we were not doing it, even Iif we redirected when creating/updating.

Fixed this. I simply removed an `elsif` condition since it pretty much does the same as the `else` case.

> I think this is a case of https://stackoverflow.com/questions/51054260/rails-5-redirect-back-with-url-query-parameter-option

This is exactly it. `redirect_back` would have been a good choice but we can't use it here as we need to attach the tab param.

app/controllers/evidence_controller.rb
dormi: I've read we can use: `URI(request.headers['Referer']).path`

dormi: And if request.referrer is not present we may got an error here ?

app/controllers/evidence_controller.rb
dormi: Also noticed that if referer had other query params, we are losing them here.

app/controllers/evidence_controller.rb
```diff
@@ -122,11 +122,12 @@ def destroy
...
-          # since the evidence does not exist anymore. That's why we check the 'Referer' here:
-          if request.headers['Referer'] == project_node_evidence_url(current_project, @node, @evidence)
-            redirect_to project_node_path(current_project, @node), notice: notice
+          # since the evidence does not exist anymore. That's why we check the referrer here.
+          referrer_path = URI(request.referrer).tap { |u| u.query = nil }.to_s
+          if !referrer_path || referrer_path == project_node_evidence_url(current_project, @node, @evidence)
+            redirect_to project_node_path(current_project, @node, tab: 'evidence-tab'), notice: notice
           else
-            redirect_back fallback_location: project_node_path(current_project, @node), notice: notice
+            redirect_to "#{referrer_path}?tab=evidence-tab", notice: notice
```
dormi: I think I can summarize all my comments in this code proposition:
```
tab = { tab: 'evidence-tab' }
if !request.referrer || URI(request.referrer).path == project_node_evidence_path(current_project, @node, @evidence)
  redirect_to project_node_path(current_project, @node, tab), notice: notice
else
  query = (URI(request.referrer).query || '').chomp(tab.to_query) << tab.to_query
  redirect_to "#{URI(request.referrer).path}?#{query}", notice: notice
end
```

Ideas to improve it welcome to

dormi: if we end up removing `redirect_back`, it would be nice to add a comment like
```
we would like to just redirect_back here
(https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_back)
but that does not allow us to include the 'tab' param
```

dormi: And one last idea: if we fine tune the UX so when we click the evidence tab on an issue, the url of the page really contain `?tab=evidence' , we could just leave the backend code almost as it was?
I've not explored that option.

app/controllers/evidence_controller.rb
aapomm: We are redirecting anyway so I don't think we should preserver that?

app/controllers/evidence_controller.rb
aapomm: I can confirm this throws an error when the referer is missing.

app/controllers/evidence_controller.rb
aapomm: I think something like this should work:
```
referrer = URI(request.referrer || '')
tab = { tab: 'evidence-tab' }
if !request.referrer || referrer.path == project_node_evidence_path(current_project, @node, @evidence)
  redirect_to project_node_path(current_project, @node, tab), notice: notice
else
  referrer.query = tab.to_query
  redirect_to referrer, notice: notice
end
```
thoughts?

brianp:

PR #518, opened 2019-09-26
Copy node label in creation forms

brianp:
### Summary

Multiple times I've clicked to create new nodes, begun typing in the add one field, and mid-label realized I wanted multiple nodes. I then select Add Multiple but have to retype the label I already typed (or copy paste). Instead, it would be a nice Quality of Life feature to copy the contents of the Add one box over to the Add multiple magically. So I may be delighted that it is already present.

### Proposed solution
On the click of the Add Multiple select. If the multiple box currently have no value, copy the contents (if any) of the Add one box into the multiple box. Have the cursor positioned in the second line ready for typing.

### How to test
Conditions for delight:

One to Multiple

- Click the new nodes + sign
- Leave the select as Add one
- Add a label name
- Then select Add multiple
- See that the label you type in Add one is now present in the textarea
- See the cursor is positioned on the second line.

Multiple to One

- Click the new nodes + sign
- Select as Add multiple
- Add a two label names
- Select the icon
- Reselect Add one
- See that the first label from the textarea is now the only label
- See the icon has been selected

Ensuring no frustration:

- Click the new nodes + sign
- Select Add multiple
- Add a label name or two
- Then select Add one
- Type a single label name
- Reselect Add multiple
- See that the contents you originally typed are unchanged

### Note
This should apply to the sub-nodes form as well.

### Examples

One to multiple:
nodes-one-to-multiple-sm

Multiple to one:
nodes-multiple-to-one-sm

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/assets/javascripts/snowcrash/pages/nodes/new_form.js
+++ b/app/assets/javascripts/snowcrash/pages/nodes/new_form.js
@@ -1,8 +1,8 @@
 (function() {
-  var copyOver = function($to, from, typeTo, typeFrom) {
+  var copyOver = function($to, from, $typeTo, $typeFrom) {
     if ($to.val() === '' && from !== '') {
       $to.val(from.split('\n')[0] + '\n');
-      $(typeTo).val($(typeFrom).val());
+      $typeTo.val($typeFrom.val());
 
       // The click function we're in actually takes focus when clicked. If
       // we called focus within the function it gets cancelled out. We give
@@ -17,16 +17,19 @@
   document.addEventListener("turbolinks:load", function() {
     if ($('[data-behavior~=copy-node-label]').length) {
       $('[data-behavior~=copy-node-label]').click(function(eventData) {
-        var $multi = $('#nodes_list');
-        var $label = $('#node_label');
+        var $nodeType = $('#node_type_id'),
+            $nodesType = $('#nodes_type_id'),
 
-        var multiVal = $multi.val();
-        var labelVal = $label.val();
+            $multi = $('#nodes_list'),
+            $label = $('#node_label'),
+
+            multiVal = $multi.val(),
+            labelVal = $label.val();
 
         if ($(this).find('input').val() === 'one') {
-          copyOver($label, multiVal, '#node_type_id', '#nodes_type_id');
+          copyOver($label, multiVal, $nodeType, $nodesType);
         } else {
-          copyOver($multi, labelVal, '#nodes_type_id', '#node_type_id');
+          copyOver($multi, labelVal, $nodesType, $nodeType);
         }
       });
     }
--- a/app/assets/javascripts/snowcrash/pages/nodes/new_form.js
+++ b/app/assets/javascripts/snowcrash/pages/nodes/new_form.js
@@ -14,7 +14,7 @@
     }
   }
 
-  document.addEventListener("turbolinks:load", function() {
+  document.addEventListener('turbolinks:load', function() {
     if ($('[data-behavior~=copy-node-label]').length) {
       $('[data-behavior~=copy-node-label]').click(function(eventData) {
         var $modal = $(this).parents('.modal'),
--- a/app/assets/javascripts/snowcrash/pages/nodes/new_form.js
+++ b/app/assets/javascripts/snowcrash/pages/nodes/new_form.js
@@ -1,43 +1,32 @@
 (function() {
-  document.addEventListener("turbolinks:load", function() {
-    if ($('[data-behavior~=multi-node]').length) {
-      $('[data-behavior~=multi-node]').click(function() {
-        var $multi = $('#nodes_list');
-        var labelVal = $('#node_label').val();
+  var copyOver = function($to, from, typeTo, typeFrom) {
+    if ($to.val() === '' && from !== '') {
+      $to.val(from.split('\n')[0] + '\n');
+      $(typeTo).val($(typeFrom).val());
 
-        if ($multi.val() === '' && labelVal !== '') {
-          $multi.val(labelVal + '\n');
-
-          $('#nodes_type_id').val($('#node_type_id').val());
-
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/snowcrash/pages/nodes/new_form.js
```diff
@@ -0,0 +1,38 @@
...
+      // we called focus within the function it gets cancelled out. We give
+      // it a brief timout to allow it to occur after this function ends
+      // but seemingly at the same time.
+      setTimeout(function() {
+        $to.focus();
+      }, 0);
+    }
+  }
+
+  document.addEventListener("turbolinks:load", function() {
```
aapomm:
```suggestion
  document.addEventListener('turbolinks:load', function() {
```

app/assets/javascripts/snowcrash/pages/nodes/new_form.js
```diff
@@ -0,0 +1,38 @@
...
+
+        if ($(this).find('input').val() === 'one') {
+          copyOver($label, multiVal, $nodeType, $nodesType);
+        } else {
+          copyOver($multi, labelVal, $nodesType, $nodeType);
+        }
+      });
+    }
+  });
+}).call(this);
```
aapomm: this should work too right?
```suggestion
})();
```

app/assets/javascripts/snowcrash/pages/nodes/new_form.js
```diff
@@ -0,0 +1,38 @@
...
+      // but seemingly at the same time.
+      setTimeout(function() {
+        $to.focus();
+      }, 0);
+    }
+  }
+
+  document.addEventListener("turbolinks:load", function() {
+    if ($('[data-behavior~=copy-node-label]').length) {
+      $('[data-behavior~=copy-node-label]').click(function(eventData) {
```
aapomm: `eventData` is not getting used anyway
```suggestion
      $('[data-behavior~=copy-node-label]').click(function() {
```

app/assets/javascripts/snowcrash/pages/nodes/new_form.js
```diff
@@ -0,0 +1,38 @@
+(function() {
+  var copyOver = function($to, from, $typeTo, $typeFrom) {
```
aapomm: We're passing the `$to` element but not the `from` element. I'm assuming this is because we don't need the `from` element, only the value. But why are we passing the `$typeFrom` element and not the value?

app/views/nodes/modals/_add_node.html.erb
```diff
@@ -21,7 +21,7 @@
     <%# we need to scope the names of the radio buttons to this specific modal
       # so that clicking a radio in the 'add child' modal won't deselect the
       # radios in the 'add branch' modal %>
-    <div class="radio">
+    <div class="radio" data-behavior="copy-node-label">
```
aapomm:
```suggestion
    <div class='radio' data-behavior='copy-node-label'>
```

app/views/nodes/modals/_add_node.html.erb
```diff
@@ -32,7 +32,7 @@
       </label>
     </div>
 
-    <div class="radio">
+    <div class="radio" data-behavior="copy-node-label">
```
aapomm:
```suggestion
    <div class='radio' data-behavior='copy-node-label'>
```

app/assets/javascripts/snowcrash/pages/nodes/new_form.js
```diff
@@ -0,0 +1,38 @@
...
+    if ($to.val() === '' && from !== '') {
+      $to.val(from.split('\n')[0] + '\n');
```
aapomm: Minor suggestion. If we add `trim()` here:
```suggestion
      $to.val(from.trim().split('\n')[0] + '\n');
```

We can catch the first element in a text area that looks like this:
```


10.0.0.1
```
which would replace the text input with `10.0.0.1`

app/assets/javascripts/snowcrash/pages/nodes/new_form.js
brianp: That's a good inconsistency catch. I'm going to make this change for `$typeFrom` to just be a passed value.

app/assets/javascripts/snowcrash/pages/nodes/new_form.js
```diff
@@ -0,0 +1,38 @@
...
+      $typeTo.val($typeFrom.val());
+
+      // The click function we're in actually takes focus when clicked. If
+      // we called focus within the function it gets cancelled out. We give
+      // it a brief timout to allow it to occur after this function ends
+      // but seemingly at the same time.
+      setTimeout(function() {
```
aapomm: @brianp This hack gets a pass for me for its brevity. However, I think we can use JS promise or Jquery deferred objects for this purpose by ensuring the `focus` call. I personally haven't used those outside of ajax calls, so if it gets too complicated, I say we go with the current approach.

brianp: I like this idea, but as far as actual Promises, anything that is defined within the `click` jquery function, will still finish within the function so
- the promise will finish
- the focus callback will finish
- then the click will finish and take focus back

Jquery deferred looks like a better candidate, but we require the modal scope, which is found within the click event, again bounding us to the inner scope of the click event which will always be the last scope to finish, and taking focus.

I like the idea overall but yeah, to complicated for the gain.

brianp: I actually decided to try this, and for no reason I could comprehend once I added the promise, double click events started firing. And due to the situation outlined above, focus was still never called at the appropriate time because `click` always finishes last.

PR #526, opened 2019-10-16
Notification Mailer

brianp:
## Spec
I want to receive notifications via email, so I can be made aware of issue progress and comments as it's happening.

Currently notifications are only made available in app. When comments are made they could be easily missed. Users would like to be able to receive notifications in their email when comments are made.

## Proposed solution
We implement a mailer to send digests (daily) and instant digests (as they come, grouped in 10minute increments). We'll reduce the amount of mail by not sending more then a single email within a 10 minute window. Multiple notifications within the 10 minute window will be grouped into a single instant digest.

We will execute both the digest and instant digests for all users (until preferences are implemented).

CE is a no-op for all digests. Users will be required to run additional setup commands applying crontabs, and editing notification settings via the console.

### Copyright assignment


app/mailers/notification_mailer.rb
```diff
@@ -0,0 +1,39 @@
...
+  def instant(user:)
+    @user = user
+    @notifications = instant_notifications
+
+    mail to: user.email, subject: 'Notifications in the last 10 minutes.'
+  end
+
+  private
+
+  def digest_notifications
```
brianp: @aapomm I've got a proposal.

We could define a class `Digest` that the thor tasks can call `Digest.send` or `Digest.send_instant`.

This class can manage finding, creating notification groups, and calling the mailer with the collection.

This can make the mailer "dumb" about the collection is receives. And moves the collection generation to an easier to test class, that is more accessible to our thor tasks.

brianp: Here's an example of what I mean (not final), and can likely my optimized/smaller, opinions welcome.

```ruby
class Digest
  INSTANT_DIGEST_INTERVAL = 10.minutes

  def self.send
    User.all.each do |user|
      Digest.new(user).send
    end
  end

  def self.send_instant
    User.all.each do |user|
      Digest.new(user).send_instant
    end
  end

  def initialize(user)
    @user = user
  end

  attr_accessor :user

  def send
    notifications = user.notifications.
      where('created_at >= ?', user.last_digest_email).
      unread.
      newest

    NotificationMailer.with(user: user, notifications: notifications).digest
  end

  def send_isntant
    notifications = user.notifications.
      where('created_at >= ?', Time.now - INSTANT_DIGEST_INTERVAL).
      unread.
      newest

    NotificationMailer.with(user: user, notifications: notifications).instant
  end
end
```

aapomm: Implemented this wrapper and went with `DigestMailer` since `Digest` is already being used. Also, thor will call the class methods `DigestMailer.send_digests` or `DigestMailer.send_instants`. I don't think thor should concern itself with the logic of which users to send it to.

app/views/layouts/mailer.html.erb
```diff
@@ -3,11 +3,175 @@
...
+        </div>
+      </div>
+
+      <!-- Dradis Pro Version
+      <div class="header">
+        <div class="logo">
+          <%#= image_tag 'DradisPro_full_small.png' %>
+        </div>
+        <div class="gravatar">
+          <%= image_tag 'profile.jpg' %>
```
brianp: @MattBudz Who's the gravatar supposed to represent here?

app/views/layouts/mailer.html.erb
```diff
@@ -3,11 +3,175 @@
...
+      </div> -->
+
+      <div class="main">
+        <%= yield %>
+      </div>
+    </div>
+
+    <div class="mail-footer">
+      <div class="row">
+        <a href="javascript:void(0)">Manage your Dradis email preferences</a>
```
brianp: This being in the default mailer will effect the emails send out by gateway I think. Who don't have preferences or the option to change them. We may need to see what's happening in pro and adjust the gateway layout as well.

app/views/layouts/mailer.html.erb
MattBudz: @brianp This would be the user that is receiving the email. Thinking about it now, I want to add some logic to only show the gravatar if there is a gravatar set or custom user profile pic uploaded. If it's just going to show the placeholder image, I want it to render just the logo. Ee can make this the same for both CE and Pro since CE now will have working gravatars. The only difference would be the CE vs Pro logo.

app/models/notification.rb
```diff
@@ -17,13 +17,22 @@ class Notification < ApplicationRecord
   scope :read,    -> { where.not(read_at: nil) }
   scope :unread,  -> { where(read_at: nil) }
 
+  # All unread notifications within a given span of time
+  scope :current, -> (time_ago=1.day) {
```
brianp:
```suggestion
  scope :current, -> (time_ago = 1.day) {
```

brianp:
```suggestion
  scope :since, -> (time_ago=1.day) {
```
`current` doesn't help me understand what current notifications would be for. Since, with a time makes more sense to read out loud. `Notification.since 1.day`

app/models/notification_group.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def each(&block)
+    @notifications_hash.each(&block)
+  end
+
+  def raw_hash
+    @notifications_hash
+  end
+
```
brianp:
```suggestion
```

app/models/notification_group.rb
```diff
@@ -0,0 +1,44 @@
...
+  def initialize(notifications)
+    @notifications_hash = build_notifications_hash(notifications)
+    @count = notifications.count
+  end
+
+  def each(&block)
+    @notifications_hash.each(&block)
+  end
+
+  def raw_hash
```
brianp:
```suggestion
  def to_h
```

app/models/notification_group.rb
```diff
@@ -0,0 +1,44 @@
+class NotificationGroup
+  attr_reader :count, :notifications_hash
```
brianp:
```suggestion
  include Enumerable
  attr_reader :notifications_hash
```

We should then get all collections based methods because you defined `each`. We'll get `count`, `map`, `inject` etc for free if we want/need them. It's only a benefit if we want to treat the group like a true collection. Not needed if count, and each is completely enough.

app/presenters/digest_presenter.rb
```diff
@@ -0,0 +1,71 @@
...
+  def avatar_image(size)
+    if notification.actor
+      h.image_tag(
+        attachments['profile'].url,
+        alt: notification.actor.email,
+        class: 'gravatar',
+        title: notification.actor.email,
+        width: size,
+        # Fallback image
+        onerror: "this.src = '#{image_path('logo_small')}';"
```
brianp: 🤯

app/presenters/digest_presenter.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+    else
+      h.content_tag :span, "#{notification.actor.email} and #{pluralize(actor_count - 1, 'other')}", class: 'user-name'
+    end
+  end
+
+  def notification
+    @notification ||=
+      if notifications.count > 1
+        # Get the first notification with an existing actor
```
brianp: Can you explain this? I don't get why there's a difference of which first notification we get.

app/models/user.rb
```diff
@@ -17,6 +17,10 @@ class User < ApplicationRecord
   # -- Callbacks ------------------------------------------------------------
   # -- Validations ----------------------------------------------------------
   # -- Scopes ---------------------------------------------------------------
+  # Preferred notification frequencies
+  scope :dailies, -> { where('preferences LIKE "%digest_frequency: daily%"') }
```
brianp:
```suggestion
  scope :digests_daily, -> { where('preferences LIKE "%digest_frequency: daily%"') }
```

app/models/user.rb
```diff
@@ -17,6 +17,10 @@ class User < ApplicationRecord
...
+  scope :instants, -> { where('preferences LIKE "%digest_frequency: instant%"') }
```
brianp:
```suggestion
  scope :digests_instant, -> { where('preferences LIKE "%digest_frequency: instant%"') }
```

Just using dailies, and instant makes me think we're looking for daily or instant Users, as they are scopes on Users. Being slightly more specific is a bit nicer.

app/services/digest_mailer.rb
```diff
@@ -0,0 +1,44 @@
...
+  def send
+    notifications = user.notifications.for_digest(interval)
+    NotificationMailer.with(user: user, notifications: notifications, type: type).
+      digest.
+      deliver_now
+  end
+
+  private
+
+  def interval
```
brianp: I don't think this is being used.

app/views/layouts/mailer.html.erb
```diff
@@ -3,11 +3,188 @@
...
+        line-height: 1.2rem;
+      }
     </style>
   </head>
 
   <body>
-    <%= yield %>
+    <div class="mail-wrapper">
+
+      <div class="header"> <!-- add class "center" if there is no gravatar or profile picture available for this user -->
```
brianp: These are html comments, and they will show up in the email if anybody looks close enough. Let's make the erb/ruby comments.

app/views/layouts/mailer.html.erb
```diff
@@ -3,11 +3,188 @@
...
+        <div class="logo">
+          <%= image_tag attachments['DradisCE_full_small'].url %>
+        </div>
+        <div class="gravatar"> <!-- Gravatar of the user recieving this email. Only render this div if a gravatar image is available or if the user uploaded a custom profile pic in pro. Do not render this div if the fallback/placeholder image will be rendered. If the is no gravatar that will be rendered, add class "center" to div.header above. -->
```
brianp: Also html comments

spec/models/notification_group_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+describe NotificationGroup do
+  before do
+    @user = create(:user)
+    @issue = create(:issue)
+    @comment = create(:comment, commentable: @issue)
+    @notification = create(:notification, notifiable: @comment, recipient: @user)
+    @user_notifications = @user.notifications.current
+  end
+
```
brianp:
```suggestion
```

app/models/notification_group.rb
```diff
@@ -0,0 +1,44 @@
...
+    @notifications_hash
+  end
+
+
+  private
+
+  # The resulting notifications hash has the format:
+  # {
+  #   project1 => [
+  #     [item1, [notifications]],
```
brianp: How come you decided to go with nested arrays instead of hashes?
Not requesting a change, just wondering what the thought process was.

```
{
  project1 => [
    { item1 => [notifications] },
    { item2 => [...] }
  ],
  project2 => ...
}
```

app/views/layouts/mailer.html.erb
```diff
@@ -3,11 +3,188 @@
...
+          <%= image_tag @avatar_url, alt: @user.email %>
```
brianp: `avatar_url` is always set to the attached profile image and won't actually attempt to load the gravatar.

app/mailers/notification_mailer.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def set_inline_attachments
+    attachments.inline['profile'] = File.read(find_asset('profile.jpg'))
+    attachments.inline['logo_small'] = File.read(find_asset('logo_small.png'))
+    attachments.inline['DradisCE_full_small'] = File.read(find_asset('DradisCE_full_small.png'))
+  end
+
+  def set_user_avatar
+    # The arguments are a noop here for CE-Pro parity.
+    @avatar_url = attachments['profile'].url
```
brianp: `avatar_url` is always set to the attached profile image and won't actually attempt to load the gravatar. CE users should also receive the gravatar if it's available.

app/services/digest_mailer.rb
aapomm: It's used in Line 27 for querying the notifications.

app/models/notification_group.rb
aapomm: Good suggestion, but `count` and `each` are the only ones we need at the moment. I'm happy to revisit this should we want to start treating this class as an Enumerable beyond the `each` method.

app/models/notification_group.rb
aapomm: This is a side-effect of the group_by method:
```
> h = {1 => 'one', 2 => 'two', 3 => 'three'}
=> {1=>"one", 2=>"two", 3=>"three"}
> h.group_by { |k,v| k % 2 }
=> {1=>[[1, "one"], [3, "three"]], 0=>[[2, "two"]]}
```
note that the values are returned as arrays.

I do agree that the hash is more readable/semantic, but with the way that we iterate through this hash (see digest.html.erb) I don't think the format matters. I'd be happy to do some preprocessing of the notifications hash should the need arise in the future.

app/presenters/digest_presenter.rb
aapomm: Added a comment for this hack

app/presenters/digest_presenter.rb
aapomm: (Updated the following line's method from `first` to `find`)

For context, we use the `notification` variable for displaying the first user in a group of notifications:
> <first_user> and 1 other users....

`notifications.first` should work, but as a precaution, we only find the first notification with an existing actor. We have multiple calls to `actor.email` around the presenter and I wanted to make sure that `actor` exists. Note that's `nil` for deleted users.

app/mailers/notification_mailer.rb
aapomm: I didn't implement that logic because we don't have working gravatars in CE yet (see: [application_helper.rb](https://github.com/dradis/dradis-ce/blob/notifications-mailer/app/helpers/application_helper.rb#L2)). What you're pointing out should be the correct behavior, though.

brianp: It's been updated in master:
https://github.com/dradis/dradis-ce/blob/master/app/helpers/avatar_helper.rb

aapomm: Updated!

app/models/notification.rb
```diff
@@ -17,13 +17,22 @@ class Notification < ApplicationRecord
   scope :read,    -> { where.not(read_at: nil) }
   scope :unread,  -> { where(read_at: nil) }
 
+  # All unread notifications within a given span of time
+  scope :since, -> (time_ago = 1.day) {
```
dormi: are we using this a lot in non spec code? I see it just once. 
I like the scope but a private method could be enough.

app/models/notification_group.rb
dormi: I like this suggestion

app/models/user.rb
```diff
@@ -17,6 +17,10 @@ class User < ApplicationRecord
   # -- Callbacks ------------------------------------------------------------
   # -- Validations ----------------------------------------------------------
   # -- Scopes ---------------------------------------------------------------
+  # Preferred notification frequencies
+  scope :digests_daily, -> { where('preferences LIKE "%digest_frequency: daily%"') }
```
dormi: These scopes could be placed on the `digest_mailer` itself?

app/services/digest_mailer.rb
```diff
@@ -0,0 +1,44 @@
+class DigestMailer
```
dormi: we are calling this class `Mailer`, but it is not a real `Mailer`

app/presenters/digest_presenter.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def created_at_ago
+    # We can't use the local_time gem here because there's no JS
+    "#{time_ago_in_words(notification.created_at)} ago"
+  end
+
+  private
+
+  def avatar_image(size)
```
dormi: here, we are overwriting the method we just included with `AvatarHelper` yes?

app/models/notification_group.rb
aapomm: When I tested this turns out we still need to have `attr_reader :count`. Otherwise, it will always return `1`, so I'm not sure if we're winning a lot by including `Enumerable` at the moment.

app/presenters/digest_presenter.rb
aapomm: yes, we can't use the original method because we 1) need to handle a non-existent user and 2) add a fallback image.

dormi: can that be done in the original method and only have 1 `avatar_image` method ?

app/models/user.rb
aapomm: Makes sense since I think we'll only be using this in the context of digests.

app/models/notification.rb
aapomm: I'm hesitant with moving this to a private method just because we lose the advantage of scopes always returning an ActiveRecord::Relation. In a scope, we can be sure that we can always chain the result.

brianp: It can still be a private AR scope if that works. Move the line under `private` and it applies the function like any other.

app/presenters/digest_presenter.rb
aapomm: I tried it out and it seems possible. However, there are gonna be some substantial changes that I think it belongs to another PR. For starters, the `avatar_image` in the helper wraps the image around a `span` tag while this presenter doesn't. The helper also forces a `style` attribute which we don't need in this presenter. Those and the fact that this refactor benefits the other presenters too makes me think it should be done in a different PR.

app/models/notification.rb
aapomm: ^ doesnt' seem to work. Adding this line does though: `private_class_method :since`. But I'm not convinced we should make it private in the first place?

brianp: > doesnt' seem to work

huh. Today I learned.

I feel it's fine to leave it as is then.

app/presenters/digest_presenter.rb
brianp: I can get onboard with doing this as a different PR. It would be good to get all the presenters/views using the same helper, but it doesn't have to be done here.

app/services/digest_sender.rb
```diff
@@ -0,0 +1,48 @@
...
+  def self.digest_users(type:)
+    User.includes(:notifications).where("preferences LIKE '%digest_frequency: #{type}%'")
+  end
+
+  def initialize(type:, user:)
+    @type = type
+    @user = user
+  end
+
+  attr_accessor :type, :user
```
etdsoft: Isn't the convention to have this right after the `class` line?

We've also used some other conventions to structure classes that I think it's worth keeping, including separating blocks of methods:

```ruby
  # -- Class Methods -----------------------------------------------------
```

and of course :alpha order.

app/presenters/digest_presenter.rb
```diff
@@ -0,0 +1,75 @@
...
+      h.image_tag(
+        avatar_url(notification.actor, size: size),
+        alt: notification.actor.email,
+        class: 'gravatar',
+        title: notification.actor.email,
+        width: size,
+        # HACK: we can't use data-fallback-image for setting the fallback image
+        # because there's no JS in the mail. Instead, we're relying on onerror
+        # callback to re-set the image tag's src attribute.
+        onerror: "this.src = '#{image_path('logo_small')}';"
```
etdsoft: Shouldn't we default to `profile`?

```suggestion
        onerror: "this.src = '#{image_path('profile')}';"
```

However see my note in the mailer re: using proper file names for the attachments.

lib/tasks/thor/digest.rb
```diff
@@ -0,0 +1,17 @@
+class DradisTasks < Thor
+  class Digest < Thor
+    namespace 'dradis:digests'
+
+    desc 'send_digests', 'Send a daily digest to all users'
+    def send_digests
```
etdsoft: see my DigestSender comment, we may want to include this in the renaming effort.

app/mailers/notification_mailer.rb
```diff
@@ -0,0 +1,31 @@
...
+  include AvatarHelper
+
+  before_action :set_inline_attachments
+
+  def digest
+    @user = params[:user]
+    @notifications = params[:notifications]
+    @type = params[:type]
+
+    set_user_avatar
```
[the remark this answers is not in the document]
aapomm: This is for parity with the Pro version. In Pro we're checking for the existence of an uploaded avatar, but it's a no-op here.

app/models/notification_group.rb
```diff
@@ -0,0 +1,43 @@
...
+    hash = notifications.
+      # FIXME: This only applies to notifications coming from a comment
+      group_by { |n| n.notifiable.commentable }
+
+    project = Project.new
+    hash.group_by do |item, _|
+      if defined?(Dradis::Pro)
+        item.project
+      else
+        # We use a single instance of Project since the project instance in each
```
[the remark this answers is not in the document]
aapomm: Good suggestion! I didn't think to approach it this way. I've overridden the methods `eql?` and `hash` to the `Project` class so the `group_by` method can see the instances as equal.

app/mailers/notification_mailer.rb
```diff
@@ -0,0 +1,31 @@
...
+  private
+
+  def find_asset(name)
+    Rails.application.assets.find_asset(name).pathname
+  end
+
+  def set_inline_attachments
+    attachments.inline['profile'] = File.read(find_asset('profile.jpg'))
+    attachments.inline['logo_small'] = File.read(find_asset('logo_small.png'))
+    attachments.inline['DradisCE_full_small'] = File.read(find_asset('DradisCE_full_small.png'))
```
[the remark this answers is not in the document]
aapomm: Symbols don't work, unfortunately.
Renamed the attachment to `dradis_logo` and added a check for `Dradis::Pro` to make it easier for syncing.

app/models/notification_group.rb
```diff
@@ -0,0 +1,43 @@
+class NotificationGroup
```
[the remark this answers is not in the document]
aapomm: Added a comment on the class

app/services/digest_sender.rb
```diff
@@ -0,0 +1,48 @@
+class DigestSender
+  DIGEST_INTERVAL   = 1.day
+  INSTANT_INTERVAL  = 10.minutes
+
+  def self.send_digests
+    digest_users(type: :daily).each do |user|
```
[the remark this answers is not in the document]
aapomm: Renamed `digest` to `daily`

app/models/notification.rb
```diff
@@ -17,13 +17,22 @@ class Notification < ApplicationRecord
...
+    where('created_at >= ?', Time.now - time_ago).unread.newest
```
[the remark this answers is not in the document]
aapomm: Updated the scope to use `.ago`.

app/models/notification_group.rb
```diff
@@ -0,0 +1,43 @@
...
+  # {
+  #   project1 => [
+  #     [item1, [notifications]],
+  #     [item2, ...]
+  #   ],
+  #   project2 => ...
+  # }
+  def build_notifications_hash(notifications)
+    hash = notifications.
+      # FIXME: This only applies to notifications coming from a comment
```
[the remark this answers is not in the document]
aapomm: Fixed. Added a type check for the notifiable.

app/views/notification_mailer/digest.text.erb
```diff
@@ -0,0 +1,12 @@
+<% if @type == :digest %>
+  Your Notifications From The Past 24h
+<% elsif @type == :instant %>
+  Your Recent Notifications
+<% end %>
+
+<% @notifications.each do |project, project_notifications| %>
+  <% project_notifications.each do |_, notifications| %>
+    <% presenter = DigestPresenter.new(notifications, self) %>
+    <%= render partial: 'notification_mailer/notification', locals: { presenter: presenter } %>
```
brianp: This isn't a plain text email. The partial loaded here has html formatting. I think we'll need a secondary partial for the plain text version.

brianp: My mistake. I completely glossed over the  `app/views/notification_mailer/_notification.text.erb` file and thought we missed it. Sorry about that.

aapomm: Added one for the notification partial: https://github.com/dradis/dradis-ce/pull/526/files#diff-18372eaa9ef09e00d3ac3ab4bef3558d

and another for the comment partial: https://github.com/dradis/dradis-ce/pull/526/files#diff-7318e2e328e3ece858e4b0585ec04989

app/mailers/notification_mailer.rb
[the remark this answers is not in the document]
etdsoft: is this still the case under the new plan of making User and Project in CE quack like in Pro?

app/mailers/notification_mailer.rb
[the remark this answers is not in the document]
etdsoft: have you confirmed these are not the attachment names that the user will get? If they are, they need extensions.

app/models/notification_group.rb
```diff
@@ -0,0 +1,43 @@
+class NotificationGroup
+  attr_reader :count, :notifications_hash
+
+  def initialize(notifications)
+    @notifications_hash = build_notifications_hash(notifications)
```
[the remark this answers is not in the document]
etdsoft: why did you close down this without addressing it?

spec/mailers/previews/notification_mailer_preview.rb
```diff
@@ -0,0 +1,9 @@
+class NotificationMailerPreview < ActionMailer::Preview
+  def digest
+    @user = User.first
+    @notifications = @user.notifications.for_digest(1.day.ago)
+    @type = :digest
+    ActionMailer::Base.default_url_options[:host] = 'dradis-framework.dev'
```
etdsoft: why is this needed? Isn't it set by smtp.yml + setup_mail initializer?
```suggestion
    ActionMailer::Base.default_url_options[:host] = 'dradisframework.dev'
```

config/schedules/digests.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+every 10.minutes do
+  thor 'dradis:digests:send_instants'
+end
+
+every 1.day, at: '9:00' do
+  thor 'dradis:digests:send_digests'
```
etdsoft: Is this the right name now?
```suggestion
  thor 'dradis:digests:send_dailies'
```

app/services/digest_sender.rb
```diff
@@ -0,0 +1,50 @@
...
+  attr_accessor :type, :user
+
+  # -- Class Methods --------------------------------------------------------
+  def self.digest_users(type:)
+    User.includes(:notifications).where("preferences LIKE '%digest_frequency: #{type}%'")
+  end
+
+  def self.send_dailies
+    digest_users(type: :daily).each do |user|
+      DigestSender.new(user: user, type: :digest).send
```
etdsoft: Shouldn't this be `type: :daily` now?

app/services/digest_sender.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    NotificationMailer.with(user: user, notifications: notifications, type: type).
+      digest.
+      deliver_now
+  end
+
+  private
+
+  def interval
+    if type == :digest
```
etdsoft: same here, if we're moving to :daily

app/views/notification_mailer/digest.html.erb
```diff
@@ -0,0 +1,29 @@
+<h2 class='digest-header'>
+  <% if @type == :digest %>
```
etdsoft: digest vs daily

app/views/notification_mailer/digest.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    Your Notifications From The Past 24h
+  <% elsif @type == :instant %>
+    Your Recent Notifications
+  <% end %>
+</h2>
+
+<% @notifications.each do |project, project_notifications| %>
+  <div class='project-wrapper'>
+    <div class='project-name-wrapper'>
+      <% if defined?(Dradis::Pro) %>
```
etdsoft: This could be removed if CE's `Project#name` returned something? Even an empty string

app/mailers/notification_mailer.rb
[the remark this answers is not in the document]
brianp: Yes. To a point. There has been changes but the changes should maintain CE and Pro parity.

app/mailers/notification_mailer.rb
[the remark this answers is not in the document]
brianp: My understanding is that you are correct, they should require extensions. mostly for aesthetics and proper mime types when downloading etc. I'll add it in.

spec/mailers/previews/notification_mailer_preview.rb
brianp: local testing in the event the smtp file doesn't exist.

config/schedules/digests.rb
brianp: Good catch.

app/services/digest_sender.rb
```diff
@@ -0,0 +1,50 @@
+class DigestSender
+  DAILY_INTERVAL    = 1.day.ago
+  INSTANT_INTERVAL  = 10.minutes.ago
```
brianp: I just noticed this change.
Setting "a time ago" in a constant is a tricky bug to catch later on. This means at the time the class is loaded `DAILY_INTERVAL` will get set to `1.day.ago`. And never change. It's a constant, and the `.ago` method is only applied when the constant is set.

This means if 1 day ago at load time is 17/11/2019. Then two weeks from today. DAILY_INTERVAL will still be 17/11/2019. Which is no longer a day ago, it's two weeks ago.

Anytime `ago` methods are used, they need to be dynamic to the time now (usually). This means we want to store the interval here which is 1.day. But never call `ago` until we want the differential.

app/services/digest_sender.rb
brianp: Yup, changed.

app/services/digest_sender.rb
brianp: Updated.

app/models/notification_group.rb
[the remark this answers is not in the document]
brianp: Variables we're previously renamed. Maybe it was showing an outdated diff. Also added additional comment.

app/views/notification_mailer/digest.html.erb
brianp: Right now it returns "Dradis CE" by default. Which we opted not to have as the project title. Happy to return it in for parity though. The amount of people using mail with CE as mentioned is likely low.

This discussion can always come back up later.

PR #527, opened 2019-10-18
When creating a sub-node, validate the parent node belongs to the same project

dormi:
### Specs

When creating sub-nodes, we do not check that the parent node id belongs to the same project.
**This is not a problem in CE, because we don't have multiple projects.**
But for our code base consistency here it goes this PR to be also (or only) merged in CE.

### How to test
This cannot be really tested in CE, but specs should keep green.

### Copyright assignment

### Check List

- [x] Added a CHANGELOG entry


spec/models/node_spec.rb
```diff
@@ -108,7 +115,8 @@
   end
 
   it 'uses a default type ID if none provided' do
-    node = Node.create(label: 'Foo')
+    project = create(:project)
```
dormi: one less difference with Pro

app/models/node.rb
```diff
@@ -102,7 +102,7 @@ def destroy_attachments
   end
 
   def parent_node
-    if self.parent.nil?
+    if self.parent.nil? || self.parent.project.id != self.project.id
```
etdsoft: let's avoid unnecessary SQL...

```suggestion
    if self.parent.nil? || self.parent.project_id != self.project_id
```

dormi: this was like this because `node.project_id` does not exist in CE

PR #528, opened 2019-10-21
Notifications email preferences

brianp:
# Spec
I want to manage my notification settings, so I can choose between getting notifications in instant intervals, once a day, or not at all.

Once notifications have been implemented users would receive both instant digests and daily digests.

# Proposed solution
We will create notification settings to allow users to decide between having no notifications, instant digests, or digests. These settings will be available through the console only.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,6 +1,9 @@
 class User < ApplicationRecord
   alias_attribute :name, :email
 
+  serialize :preferences, UserPreferences
+  validates_associated :preferences
+
   # -- Relationships --------------------------------------------------------
   has_many :activities
   has_many :comments, dependent: :nullify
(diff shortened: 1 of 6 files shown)


db/migrate/20191024132834_set_default_notification_settings.rb
```diff
@@ -0,0 +1,13 @@
+class SetDefaultNotificationSettings < ActiveRecord::Migration[5.1]
+  def up
+    User.all.each do |user|
+      user.preferences.digest_frequency = 'instant'
```
etdsoft: we may need a `#save` here, or maybe we can remove the migration entirely (what are the chances of someone upgrading Dradis mid-project?)

PR #529, opened 2019-10-22
Notifications config

aapomm:
# Spec
I want to be able to manage the SMTP configuration through a YAML file that will persist through upgrades.

# Proposed solution
Add an initializer that reads a YAML config file and sets the options for actionmailer.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -35,6 +35,10 @@ chdir APP_ROOT do
     cp 'config/secrets.yml.template', 'config/secrets.yml'
   end
 
+  unless File.exist?('config/smtp.yml')
+    cp 'config/smtp.yml.template', 'config/smtp.yml'
+  end
+
   puts "\n== Preparing database =="
   system! 'bin/rails db:setup'
 
--- a/config/initializers/z_03_setup_mail.rb
+++ b/config/initializers/z_03_setup_mail.rb
@@ -1,9 +1,12 @@
+# Defaults
+ActionMailer::Base.default_url_options[:host] = 'dradis-pro.dev'
+ActionMailer::Base.default_url_options[:script_name] = '' if Rails.env.production?
+
 config_path = Rails.root.join('config/smtp.yml')
 
 if File.exist?(config_path)
   config = YAML.load_file(config_path)[Rails.env]
 
   ActionMailer::Base.smtp_settings = config['smtp_settings']
-  ActionMailer::Base.default_url_options = config['default_url_options']
-  ActionMailer::Base.default_url_options[:script_name] = '' if Rails.env.production?
+  ActionMailer::Base.default_url_options = config['default_url_options'].symbolize_keys
 end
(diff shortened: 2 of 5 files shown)


config/initializers/z_03_setup_mail.rb
```diff
@@ -0,0 +1,12 @@
+# Defaults
+ActionMailer::Base.default_url_options[:host] = 'dradis-pro.dev'
```
brianp:
```suggestion
ActionMailer::Base.default_url_options[:host] = 'dradis-framework.dev'
```

config/smtp.yml.template
```diff
@@ -0,0 +1,20 @@
+default: &default
+  default_url_options:
+    host: dradis-pro.dev
```
brianp:
```suggestion
    host: dradis-framework.dev
```

config/smtp.yml.template
```diff
@@ -0,0 +1,20 @@
...
+  smtp_settings:
+    address:               smtp.gmail.com
```
brianp: I don't think the default should actually be setup with an address.
If someone runs bin/setup this setup will be used. If they create issues gmail will start receiving un-authenticated mail attempts. Which would leak comments/issue information into their logs even if the mail is rejected.

bin/setup
```diff
@@ -35,6 +35,10 @@ chdir APP_ROOT do
     cp 'config/secrets.yml.template', 'config/secrets.yml'
   end
 
+  unless File.exist?('config/smtp.yml')
```
brianp: Should we be adding this config file during setup? If the config exists mail will attempt to send even if they haven't populated any details.

I think it's safer and fine to not automatically apply the smtp config on setup. We have the sample file, and the documentation regarding it's application.

brianp: Actually- The user still has to setup cronjobs for emails to be triggered to send. So even if the SMTP config is blank in CE it won't change anything until the use specifically goes to setup emails, in which case they'll add settings. So I think this will be fine.

config/smtp.yml.template
```diff
@@ -0,0 +1,21 @@
+default: &default
```
etdsoft: I'd include some comments as a header, and a pointer to:
https://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-configuration

I'd also include the entire list of settings, a notable absent is `:openssl_verify_mode` as some people may be tripped by the use of self-signed SSL certs in their organisation.

Also, are we better of providing blank defaults, or would it be better to provide some clearly bogous details: "mail.starfleet.com", port 123456, etc. I suspect some errors may slip through the cracks with blank values but may become more evident if some crazy defaults are provided.

etdsoft:
I don't know how we're planning to handle this, but the ActionMailer config has an interesting-looking `default_options` key that can be used to set default :from and :reply_to addresses, that could be one good way to ensure that people can reply to the emails and at least they'd be routed to a human?

I can see how in the future we can implement some sort of "reply to comment" feature, but that's definitely out of scope for this cycle.

config/smtp.yml.template
aapomm: ✅

PR #532, opened 2019-10-23
When updating a record with ajax, do not show conflict resolver message on the next record we visit

dormi:
### Summary
The [conflict resolver](https://github.com/dradis/dradis-ce/blob/master/app/controllers/concerns/conflict_resolver.rb) uses a param (`original_updated_at`) to know if the record have been saved after we started editing it.
When this param is not set, we detect it as value 0, and the conflict resolver always detects a revision conflict (wrong). 
This happens for instance when updating a record with ajax (tagging an issue...), because we don't send the full list of form params, including `original_updated_at`.
A quick and easy fix is to just skip the conflict resolver check when that param is not set. It makes sense because conflict detection does not make a lot od sense for this kind of fast interactions (tagging an issue with ajax).

### Copyright assignment

### Check List

- [x] Added a CHANGELOG entry


spec/support/edit_conflicts_shared_examples.rb
```diff
@@ -29,8 +29,16 @@ def conflict_warning
     end
   end
 
-  context "when another user has updated the record in the meantime" do
-    let(:email_1) { "someone@example.com" }
+  specify "an update wo/ original_updated_at doesn't say anything about edit conflicts", js: true do
```
dormi: this added spec is the only real change in this file

aapomm:
```suggestion
  specify 'an update wo/ original_updated_at doesn't say anything about edit conflicts', js: true do
```

dormi: I am not sure about committing this suggestion, the string already contains a `'` in `doesn't`

aapomm: oh you're right. use `doesn\'t` instead?

app/controllers/concerns/conflict_resolver.rb
```diff
@@ -18,7 +18,7 @@ module ConflictResolver
 
   def check_for_edit_conflicts(record, updated_at_before_save)
     name = record.model_name.name.downcase
-    if params[name][:original_updated_at].to_i < updated_at_before_save
+    if params[name][:original_updated_at] && params[name][:original_updated_at].to_i < updated_at_before_save
```
etdsoft: there is too much going on in this conditional, what about some cleanup?

```ruby
def check_for_edit_conflicts(record, updated_at_before_save)
  return unless params[name][:original_updated_at]

  name = record.model_name.name.downcase
  if params[name][:original_updated_at].to_i < updated_at_before_save
    # ...
```

PR #536, opened 2019-10-28
NotificationsMailerPresenter

aapomm:


app/services/digest_mailer.rb
```diff
@@ -24,11 +24,15 @@ def initialize(type:, user:)
   attr_accessor :type, :user
 
   def send
+    # FIXME: This only applies to notifications coming from a comment
     notifications = user.notifications.
```
brianp: We could declare scope(s) for this.
`notifications.for_digest(interval)`

PR #540, opened 2019-10-29
Remove unexpected flash keys

dormi:
### Summary

We had an "improper" use of the flash object that was causing some errors:

We stored unexpected keys (`update_conflicts_since`) in the flash object. Since we use a helper to display the flash messages, the helper raised an error on unexpected keys.
To fix this, we moved away from using the flash object in those cases.

When we want to store a temporal value in the session , we use the session.

### Copyright assignment

### Check List

- [x] Added a CHANGELOG entry


app/controllers/concerns/conflict_resolver.rb
```diff
@@ -28,15 +28,16 @@ def check_for_edit_conflicts(record, updated_at_before_save)
...
   end
 
   def load_conflicting_revisions(record)
-    if flash[:update_conflicts_since]
+    if session[:update_conflicts_since]
       @conflicting_revisions = record.versions\
-        .order("created_at ASC")\
+        .order('created_at ASC')\
         .where("created_at > '#{flash[:update_conflicts_since]}'")
+      session[:update_conflicts_since] = nil
```
etdsoft:
```suggestion
      session.delete(:update_conflicts_since)
```

PR #545, opened 2019-11-12
Notifications per project

aapomm:
**Note**: This is not an issue in CE but is done for CE-Pro code parity.

### Spec
When clicking on the notification dropdown, the NotificationPresenter generates a link to the commented item using the current_project. This doesn't work since the notification can come from a different project. In that case, the link shows a 404 instead.

Relevant code: https://github.com/dradis/dradis-ce/blob/master/app/presenters/notification_presenter.rb#L16

**Proposed solution**
Stop showing notifications from different projects by scoping the notifications under the current project.

### How to test

1. Login to Account#1
2. Open an incognito window and login to Account#2
3. Access one project using the 2 accounts and subscribe to the same issue.
4. Visit a different project using Account#1
5. Using Account#2, comment on an issue.
6. Using Account#1, click on the new notification.
7. Confirm that the link was visited correctly and no error was thrown.


app/models/notification.rb
```diff
@@ -25,6 +25,15 @@ def self.mark_all_as_read!
   end
 
   # -- Instance Methods -----------------------------------------------------
+  def project
+    @project ||= Project.new
+  end
+
+  def project=(new_project)
```
dormi: Do we need to call `save` ?
In the `Node` model we have:
`def project=(new_project); end`

PR #548, opened 2019-11-19
Mention & Avatar Remix

brianp:
### Summary
A refactoring to help with a handful of inconsistencies.

- [x] Gravatars not rendering outside of comments. (Notifications page, notifications dropdown)
- [x] Inconsistent use of the default avatar image. In some places it's the dradis logo in other places it's a person silhouette
- [ ] ~Broken default avatar image for emails from air gapped instances~
- [x] Some gravatars are rendering with extremely low quality
- [x] Mentioning a user with a non email login
- [x] General mentions refactoring of code
- [x] avatar rendering unification
- [ ] ~Rendering of avatar/name for users who were removed is inconsistent. In some places it will continue to render their avatar and show a name (activity feed) while other places will show default image, and "this user has been deleted"~

### Check List

- [ ] Added a CHANGELOG entry


app/helpers/avatar_helper.rb
```diff
@@ -1,44 +1,43 @@
...
-      image_path('profile.png')
-    else
-      gravatar_id = Digest::MD5.hexdigest(user.email.downcase)
-      size = options.fetch(:size, 64).to_i
-      "https://secure.gravatar.com/avatar/#{gravatar_id}.png?r=PG&s=#{size}"
-    end
+    return DEFAULT_PROFILE_IMAGE if user.nil? || !user.email.include?('@')
+
+    gravatar_id = Digest::MD5.hexdigest(user.email.downcase)
+    size = options.fetch(:size, DEFAULT_PROFILE_IMAGE_SIZE).to_i * 2 # Retna displays mean dot density can be higher.
```
dormi:
```suggestion
    size = options.fetch(:size, DEFAULT_PROFILE_IMAGE_SIZE).to_i * 2 # Retina displays mean dot density can be higher.
```

app/helpers/avatar_helper.rb
```diff
@@ -1,44 +1,43 @@
...
+    "https://secure.gravatar.com/avatar/#{gravatar_id}.png?r=PG&s=#{size}&d=forceErrorOnDefault"
```
dormi: can you explain this option? I'm looking for it in https://en.gravatar.com/site/implement/images/ but can't find it

brianp: If no default is provided gravatar will provide *their* default image, the gravatar logo. We don't want that. Our code specifies that if a network error occurs while loading (air-gapped instance) we use our fallback images. But if no error occurs, and the user doesn't have gravatar we want to ensure our fallback image is still applied. This triggers our fallback usage by purposefully supplying a url that is inaccessible.

Based on the conversation today about air-gapped instances. We could also provide a URL to the default asset on the VM, and if it works it works, if it doesn't work we *might* still fallback properly, otherwise it won't work and it sounds like that's okay.

app/helpers/mentions_helper.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module MentionsHelper
+  def mention_matcher
```
dormi: private method ?

brianp: Modules don't have private methods. Which makes me want to check how that works again 🤔
But I verified, by adding `private` it does not prevent the view context from accessing the method. I think everything in a helper is fair game to the view.

app/models/evidence.rb
```diff
@@ -11,6 +11,7 @@ class Evidence < ApplicationRecord
   belongs_to :node, touch: true
   has_many :activities, as: :trackable
 
+  delegate :project, to: :node
```
etdsoft: why do I dislike this so much?!

app/helpers/comments_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module CommentsHelper
+  def comment_formatter(comment)
```
etdsoft: I can see how other "text like places" may need a "mentions_formatter" to format their content, but it seems that this helper has now turned into a _very_ thin shell? Do we still need it?

app/helpers/mentions_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+    @mention_matcher ||= begin
+      users = current_project.authors.all.select(:email, :name)
```
etdsoft: Do we need this extra call?

`current_project.authors` is already cached as `@mentionable_users` by `app/controllers//concerns/mentioned.rb` and also included as a `<meta>` tag in the layout's header via `app/views//comments/_mentionable_users.html.erb`.

app/models/comment.rb
```diff
@@ -49,14 +49,13 @@ def notify(action)
...
-          mentioned_users << mentioned_user
-        end
+      emails = []
+      HTML::Pipeline::MentionFilter.mentioned_logins_in(content, /[a-z0-9][a-z0-9\-@\.]*/) do |_, login, _|
+        emails << login
       end
 
-      mentioned_users
+      project = commentable.project
+      project.authors.where(email: emails.uniq)
```
etdsoft: Nice refactor 🙌

Although the `Evidence` end of it gives me an allergic reaction. Is it just me?

app/helpers/comments_helper.rb
brianp: It's an abstraction that lives in another place which I think is the biggest pitfall.
I quite like the separation of the wrapper though, because this, and the mention formatter perform different jobs. Not even because we *might* re-use it, but because it doesn't muddle the concept of what's happening. It would be simple to do it all at once, but it becomes less clear if we begin santizing html at the same time as we're enhancing mentions.

Also- with the sanitization that happens there's a required order of operations here.

I'm happy to combine them though if it's preferred.

app/helpers/mentions_helper.rb
brianp: I had higher hopes for the `meta` tag. But because we're doing the parsing server side it doesn't help unless we opt to move the mention rendering to the client...

Having the `@mentionable_users` already cached could be utilized, but we would need to expand it to be included in `app/controllers/comments_controller.rb` and in the create, update actions, not just show.

Because the comments get rendered 1 by 1 when being created or edited we need the list accessible at that point.

It makes it explicit when and where mentions can be rendered. Where right now any call to the mentions formatter will prep and cache for the request. I'll expand Mentioned to support other actions (when requested) add a commit and see what you think.

app/models/comment.rb
brianp: The `Evidence` delegate brings it together some. I think it was the only `item` without it so now they're more consistent at least.

app/helpers/mentions_helper.rb
brianp: I knew there was an outlier. `activites_controller` also renders comments, which means we would need it there too which would cache the query on every poll interval, even if it doesn't have comments, or need mentions.

Maybe there's an inbetween- although it can be dangerous for performance if we miss it later.

`users = @mentionable_users ? @mentionable_users : current_project.authors.all.select(:email, :name)`

This will use the cached collection if available, but also build a new one only if needed.

app/helpers/comments_helper.rb
etdsoft: it seems this is the job the Html::Pipeline was designed to handle? We can have a Comments pipeline that does whatever is needed, and build another pipeline for other types of formatters?

brianp: Oh I think you might like where this comment lead me.

app/helpers/mentions_helper.rb
```diff
@@ -3,7 +3,7 @@
 module MentionsHelper
   def mention_matcher
     @mention_matcher ||= begin
-      users = current_project.authors.all.select(:email, :name)
+      users = @mentionable_users || current_project.authors
```
dormi: we already did the `|| current_project.authors` in `find_mentionable_users` ?

app/helpers/comments_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+    @comment_pipeline ||= HTML::Pipeline.new [
+      HTML::Pipeline::DradisCommentFilter,
+      HTML::Pipeline::DradisMentionsFilter,
+    ], {
+      mentionable_users: @mentionable_users || current_project.authors,
```
dormi: we already did `|| current_project.authors` in `find_mentionable_users` so we can skip it here?

brianp: See this comment: https://github.com/dradis/dradis-ce/pull/548#discussion_r348703802

lib/dradis/ce.rb
```diff
@@ -9,5 +9,7 @@ def self.version
 end
 
 require 'html/pipeline/dradis_code_highlight_filter'
+require 'html/pipeline/dradis_comment_filter'
```
etdsoft: It's :alpha 😭😍

PR #550, opened 2019-11-21
Rework html email template to be compatible with wider range of email clients

MattBudz:
### Summary

This PR changes the html email structure to position everything with html tables and inline css.

### Check List

- [ ] Added a CHANGELOG entry


app/views/notification_mailer/digest.html.erb
```diff
@@ -1,25 +1,98 @@
...
+    <table cellspacing="0" cellpadding="0">
+      <% @notifications.each do |project, project_notifications| %>
+        <tr>
+          <td colspan="12">
+            <table cellspacing="0" cellpadding="0" align="center">
+              <tr>
+                <td height="20px"></td>
+              </tr>
+              <tr>
+                <% if defined?(Dradis::Pro) %>
```
brianp: We decided we're fine with the Dradis CE project title here. It make the Pro / CE sync easier to manage.

app/views/notification_mailer/digest.html.erb
```diff
@@ -1,25 +1,98 @@
...
+                <td width="590px">
+                  <hr style="border-top: 0.5px solid #ddd; border-bottom: 0.5px solid #ddd; border-right: 0.5px solid transparent; border-left: 0.5px solid transparent;">
+                </td>
+                <td width="20px"></td>
+              </tr>
+            </table>
+          </td>
+        </tr>
+        
+        <% unless defined?(Dradis::Pro) %>
```
brianp: Same here. No problem having a link to the notifications page. It does actually exist in CE and should work fine.

app/views/notification_mailer/digest.html.erb
MattBudz: ✅

app/views/notification_mailer/digest.html.erb
MattBudz: ✅

app/views/layouts/mailer.html.erb
```diff
@@ -15,174 +15,68 @@
...
+          <table cellpadding="20px">
+            <tr>
+              <td width="320" style="padding-bottom: 0;" colspan="6">
+                <div style="text-align: left;">
+                  <%= image_tag attachments['dradis_logo.png'].url, height: '80px' %>
+                </div>
+              </td>
+              <td width="320" style="padding-bottom: 0;" colspan="6">
+                <div style="text-align: right;">
+                  <%= avatar_image @user, alt: @user.email, height: '70px' %>
```
brianp: `height` isn't a valid attribute for `avatar_image` we'll want `size: 70`

app/helpers/avatar_helper.rb
```diff
@@ -28,7 +28,7 @@ def avatar_image(user, options = {})
       alt: alt,
       data: { fallback_image: fallback_image },
       height: size,
-      style: "width: #{size}px; height: #{size}px",
+      style: "width: #{size}px; height: #{size}px; border-radius: 50%;",
```
etdsoft: this doesn't belong here? the spirit of this `:style` was to include sizing information only, the webapp already has the `border-radius` CSS elsewhere.

I understand this is here to make it look nice in the embedded mail views, however we may need to reconsider passing a `:style` key to the options method that gets merged here. That way the mailer view can use it, and we avoid repeating this inline redundant rule all over the app?

app/views/layouts/mailer.html.erb
```diff
@@ -15,174 +15,68 @@
...
-        white-space: nowrap;
-        align-items: center;
-      }
-      .main .project-wrapper .project-name-wrapper .project-name {
-        font-weight: 700;
-        font-size: 1.1rem;
-        margin: 0 0.5rem;
-        line-height: 1.2rem;
+      {
+        .gravatar {}
```
etdsoft: if it's empty, why keep it?

MattBudz: That should've been removed.

app/helpers/avatar_helper.rb
MattBudz: ✅

PR #551, opened 2019-11-22
Fix relative path for thor command

brianp:
### Summary

Thor uses a relative path for template files that was not updated during refactoring and would break welcome kit installation.

This addresses the issue by updating the relative path, as well as updating a function to utilize the path change.

### DIFF
--- a/lib/tasks/thor/setup.rb
+++ b/lib/tasks/thor/setup.rb
@@ -8,7 +8,7 @@ class DradisTasks < Thor
     namespace 'dradis:setup'
 
     def self.source_root
-      File.join(File.dirname(__FILE__), 'templates')
+      File.join(File.dirname(__FILE__), '../templates')
     end
 
     desc 'configure', 'Creates the Dradis configuration files from their templates (see config/*.yml.template)'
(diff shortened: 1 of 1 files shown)


lib/tasks/thor/setup.rb
```diff
@@ -8,7 +8,7 @@ class Setup < Thor
     namespace 'dradis:setup'
 
     def self.source_root
-      File.join(File.dirname(__FILE__), 'templates')
+      File.join(File.dirname(__FILE__), '../templates')
```
dormi:
```suggestion
      File.join(__dir__, '../templates')
```

lib/tasks/thor/setup.rb
```diff
@@ -84,7 +84,7 @@ def welcome
       )
 
       importer = Dradis::Plugins::Projects::Upload::Template::Importer.new(task_options)
-      importer.import(file: File.expand_path('../templates/project.xml', __FILE__))
+      importer.import(file: template('project.xml'))
```
dormi: 👌

PR #555, opened 2019-11-25
Avatar fallback

brianp:
### Summary

Use the image inline onerror fallback for all avatars. This prevents issues with asynchronously loaded views such as comments or activities. Previously the fallback js was applied on pageload which meant async renders would not have the fallback applied. 

### DIFF
--- a/app/views/comments/_mentionable_users.html.erb
+++ b/app/views/comments/_mentionable_users.html.erb
@@ -1,10 +1,8 @@
 <%# this is picked up by javascripts/snowcrash/modules/comments.js.coffee %>
 <% cache(@mentionable_users) do %>
-  <meta name="mentionable-users" content="<%= @mentionable_users.map do |user|
-    {
-      key: h(user.email),
-      value: user.email,
-      avatar_url: avatar_url(user)
-    }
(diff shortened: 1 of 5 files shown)


app/views/comments/_mentionable_users.html.erb
```diff
@@ -6,5 +6,5 @@
       value: user.email,
       avatar_url: avatar_url(user)
     }
-  end.to_json %>">
+  end.to_json %>" data-fallback-image="<%= image_url(AvatarHelper::DEFAULT_PROFILE_IMAGE) %>">
```
etdsoft: we're stretching ERB inlining in this template, maybe time to extract to a `content_tag :meta` call

brianp: I figured there would be an easy way to manage this. And there isn't.

TIL: `<%= content_for :meta, {any: :options} %>`

will produce `<meta>{any: :options}</meta>`

And it doesn't look like rails has *any* included methods for adding meta tags nicely. So as much as I'd like to fix it up more, lets merge as-is and worry about prettying it up later on.

PR #558, opened 2019-12-05
Implement Project#testers

aapomm:
_Note: This is a backport from Pro to make syncing easier._
Add a testers scope for projects to query for all the users that can access the project.

### DIFF
--- a/spec/models/comment_spec.rb
+++ b/spec/models/comment_spec.rb
@@ -22,7 +22,7 @@ describe Comment do
       user2   = create(:user, email: 'bar@dradis.test')
       comment = create(:comment, content: 'Hello @foo@dradis.test and hello @bar@dradis.test')
 
-      expect(comment.mentions).to eq [user1, user2]
+      expect(comment.mentions).to match_array [user1, user2]
     end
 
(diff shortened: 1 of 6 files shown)


spec/models/comment_spec.rb
```diff
@@ -24,6 +24,14 @@
 
       expect(comment.mentions).to eq [user1, user2]
     end
+
+    it 'detects admins in mentions' do
+      user1   = create(:user, :admin, email: 'admin@dradis.test')
+      user2   = create(:user, email: 'foo@dradis.test')
+      comment = create(:comment, content: 'Hello @admin@dradis.test and @foo@dradis.test')
+
+      expect(comment.mentions).to eq [user1, user2]
```
dormi: in `CE` we are switching to `match_array`.
I am not sure why, but we should be consistent.

aapomm: I'm guessing it's due to the default ordering of the database, which does not exist. This means that `eq` will only work sometimes. `match_array` or `contains_exactly` should be the right method to use here since the order does not really matter.

PR #569, opened 2019-12-26
Upgrade Rails to 5.2.4.1

dormi:


config/application.rb
```diff
@@ -4,25 +4,29 @@
 require "active_model/railtie"
 require "active_job/railtie"
 require "active_record/railtie"
+require "active_storage/engine"
```
brianp: Do we actually need to load active_storage if we're not using it?

app/models/activity.rb
```diff
@@ -22,7 +22,7 @@ def project=(new_project); end
   # -- Scopes ---------------------------------------------------------------
 
   scope :latest, -> do
-    includes(:trackable).order("`activities`.`created_at` DESC").limit(10)
+    includes(:trackable).order(Arel.sql("`activities`.`created_at` DESC")).limit(10)
```
brianp: The discussion around this fix was an interesting read but it made me wonder why we needed to fix this instance in the app, but no other instance of `.order("created_at DESC")` was throwing the same deprecation. Seems to be related specifically to the backticks which may be needed under certain circumstances but not in ours. As long as were dealing with column names `order("activities.created_at DESC")` or even `order("created_at DESC")` will do.

app/models/methodology.rb
```diff
@@ -154,7 +154,7 @@ def content
     if name_changed?
       doc.xpath('/*/name/text()')[0].replace(@name)
       @content = doc.to_s
-      @changed_attributes.clear
+      changes_applied
```
brianp: `changes_applied` has the implication of still storing the changes that happened on the record. So it's not quite the same thing as clearing the previous hash. If tests are green seems okay though. Just wanted to point out the difference.

config/application.rb
dormi: Let's load it, so we are one step closer to use it.

app/models/activity.rb
dormi: Thanks. Removed backticks.

app/models/methodology.rb
dormi: I went back in the docs (`4.0`) to see when `changed_attributes.clear` was supposed to be called when using `ActiveModel::Dirty`. It was supposed to be called on `save`: https://github.com/rails/rails/blob/4-0-stable/activemodel/lib/active_model/dirty.rb#L43

In the current version of the docs (`5.2`), on `save` I see the recommended call is `changes_applied`: https://github.com/rails/rails/blob/5-2-stable/activemodel/lib/active_model/dirty.rb#L48

If the sepcs go green, I think we can keep this change.

PR #589, opened 2020-02-13
Fix fileupload prompt

MattBudz:
**Spec**
Currently when copy + pasting an image into dradis it will request to rename the image multiple times based on the number of views visited. This is caused by turbolinks re-binding the event on each page load.

**Solution**
Unbind the event on each page unload

**Extra**
I also noticed that large uploaded images overflow the containing div when displayed within a note, evidence, etc. This is fixed here as well.

### Check List

- [x] Added a CHANGELOG entry


app/assets/javascripts/tylium/behaviors.js.coffee
```diff
@@ -216,3 +219,8 @@ document.addEventListener "turbolinks:load", ->
   $('[data-behavior~=smooth-scroll]').on 'click', ->
     target = $(this).data('target');
     $('[data-id~=' + target + ']')[0].scrollIntoView({behavior: "smooth"});
+
+# Un-bind fileUpload on page unload.
+document.addEventListener "turbolinks:before-cache", ->
```
aapomm:
```suggestion
document.addEventListener 'turbolinks:before-cache', ->
```

app/views/layouts/tylium.html.erb
```diff
@@ -11,7 +11,7 @@
     <% if flash.any? %>
       <% # when flash messages present, disable turbolinks cache to avoid
         # the message appear and disappear in the next page load %>
-      <meta name="turbolinks-cache-control" content="no-cache">
+      <meta name="turbolinks-cache-control" content="reload">
```
etdsoft, line 14: why is this needed?

app/assets/javascripts/tylium/behaviors.js.coffee
```diff
@@ -217,3 +220,9 @@ document.addEventListener "turbolinks:load", ->
     history.pushState({}, '', this.href);
     $(this.hash)[0].scrollIntoView behavior: 'smooth'
     return
+
+# Un-bind fileUpload on page unload.
+document.addEventListener 'turbolinks:before-cache', ->
```
etdsoft, line 225: I think this code needs to be moved closer to the initialiser block:

- It saves the `if ().length` check
- It makes it easier to understand
- It allows us to potentially extract the entire behaviour elsewhere.

PR #596, opened 2020-02-19
Fix cards revisions

aapomm:
### Spec
Currently, the app assumes that every created item in dradis has a revision for "create". Unfortunately, this assumption fails for projects/cards created before the revision history was implemented.

**Proposed Solution**
hide the "Previous Revision" section for such scenario.


brianp:

PR #597, opened 2020-02-19
Better "dots" menu

etdsoft:
We do have a great interface for Node "extra" operations:

Screenshot 2020-02-19 at 11 27 50

And it would be great if we could use it for other "extra" operations, including:
- Save Board as template
- Send To menu of issues, etc.

The current implementation can be massaged a bit to include these "extra" menus as part of the already existing `:breadcrumbs` content block, but some additional tweaks to the CSS are needed (we're currently tying too tightly the CSS to the Node views):

Screenshot 2020-02-19 at 11 28 26

### Check List

- [x] Added a CHANGELOG entry


etdsoft:
I hadn't checked out the Tylium style guide, but it's very sleek!

Screenshot 2020-03-04 at 10 58 15

PR #601, opened 2020-02-28
fix long project names & search bar interference

MattBudz:
Currently, when there is a long project name it causes issues with the search bar expansion. This PR justs the navbar CSS and HTML to fix this bug.

### Check List

- [x] Added a CHANGELOG entry


CHANGELOG
```diff
@@ -1,26 +1,20 @@
...
+       - Long project names interfering with search bar expansion.
   * New integrations:
-    - [new integration #1]
-    - [new integration ...]
+       - 
   * Integration enhancements:
-    - Nexpose: better cipher wrapping coverage
-    - Nikto: support new nested <niktoscan> format
+       - 
   * Reporting enhancements:
```
etdsoft: L7
L9
L12, L13

should not disappear.

Also worth looking into what's going on on L21-23

MattBudz: ✅

PR #611, opened 2020-03-10
Smooth scroll link enhancements

MattBudz:
### Spec
Currently, smooth scrolling is done using javascript only and does not use anchor links. If JS is disabled the links do not function.

### Solution
While working on this I made some significant improvements around anchor links and smooth scrolling.

- The awkward jump to the top of a view while using anchors is due to turbolinks. This is a known turbolinks bug but no fix is available yet. I added JS to check for any links with href attributes that start with "#". If any exist, data-turbolinks="false" is added to the anchor to eliminate this jump.

- We had a couple <a href="#"> elements that are updated to <a href="javascript:void(0)"> to prevent reloading the view and jumping to the top of the page.

- The /styles quick nav now uses anchor links rather than 100% JS.
     - when a quick nav link is clicked the hash is added to the url.
     - if someone navigates to (for example) /styles#tables the page will auto-smooth-scroll to the tables section.

### Check List


app/assets/javascripts/tylium/behaviors.js.coffee
```diff
@@ -212,7 +212,25 @@ document.addEventListener "turbolinks:load", ->
       else
         $(this).find($('[data-behavior~=scroll-wrapper')).removeClass('hidden');
 
-  # Smooth Scrolling
-  $('[data-behavior~=smooth-scroll]').on 'click', ->
-    target = $(this).data('target');
-    $('[data-id~=' + target + ']')[0].scrollIntoView({behavior: "smooth"});
+  # Disable turbolinks for on-page anchor links (prevents page from jumping to top and allows smooth-scrolling)
+  if $('a[href^="#"]').length
```
aapomm: I don't think we need this check. The `each` in the next line will not iterate anyway if there are no elements found.

app/assets/javascripts/tylium/behaviors.js.coffee
```diff
@@ -212,7 +212,25 @@ document.addEventListener "turbolinks:load", ->
...
+    $('a[href^="#"]').each ->
+      if $(this).data('turbolinks')
+        return
+      else
+        $(this).attr 'data-turbolinks', 'false'
+      return
```
aapomm: Would something like this work?
```suggestion
    $('a[href^="#"]').each ->
      if !$(this).data('turbolinks')
        $(this).attr 'data-turbolinks', 'false'
      return
```

MattBudz: works for me!

app/assets/javascripts/tylium/behaviors.js.coffee
```diff
@@ -212,7 +212,23 @@ document.addEventListener "turbolinks:load", ->
...
+  if $('a[href^="#"]').length
+    $('a[href^="#"]').each ->
+      if !$(this).data('turbolinks')
+        $(this).attr 'data-turbolinks', 'false'
+      return
+
+  # Smooth Scrolling - scroll to element on page load if hash present in current browser url
+  if window.location.hash
+    $target = window.location.hash
+    $($target)[0].scrollIntoView behavior: 'smooth'
```
etdsoft:
```suggestion
    target = window.location.hash
    $(target)[0].scrollIntoView behavior: 'smooth'
```
The $var convention is to signify it contains a jQuery object.

PR #614, opened 2020-03-12
Author placement

MattBudz:
### Summary
Issue and ContentBlock (pro-only) seem to be displaying the author in the lower-right corner. Notes, evidence, cards do not display an author.

### Solution
Add author to notes, evidence, cards

### Check List

- [x] Added a CHANGELOG entry


db/schema.rb
```diff
@@ -60,6 +60,7 @@
     t.integer "previous_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string "author"
```
etdsoft: how is this happening? why are we adding a new field to the DB?

db/migrate/20200311133454_add_author_to_cards.rb
```diff
@@ -0,0 +1,5 @@
+class AddAuthorToCards < ActiveRecord::Migration[5.2]
```
etdsoft: @MattBudz as a general rule we don't slip changes to the DB as part of individual tasks w/o previous discussion.

Can you remove this (and the db/schema.rb) changes from here?

Maybe we need to open a separate convo for what to do with the Card's author.

app/views/issues/show.html.erb
```diff
@@ -54,7 +54,7 @@
                 <%= markup(@issue.text) %>
               </div>
               <div class="author-info">
-                <span class="ml-auto">Issue Author: <%= @issue.author %></span>
+                <span class="ml-auto">Author: <%= @issue.author || 'n/a' %></span>
```
etdsoft: why is 'n/a' needed now?

(same for the other files)

MattBudz: I noticed that when some cards have been populated from a template the author is nil. Resulting in the view rendering only `Author:`. I added this to each file in case there were to ever be something that doesn't have an author for any reason. Alternatively, we can check for author and not render anything if there is no author (this will also require a conditional class on the content-container).

aapomm: Seems inconsistent with content blocks since it doesn't use `n/a` there.

MattBudz: > Seems inconsistent with content blocks since it doesn't use n/a there.

https://github.com/secroots/dradis-pro/pull/862

PR #622, opened 2020-03-26
Use POST for preview endpoint

aapomm:
### Spec
The textile preview endpoint has a character limit because we are using GET and we pass the text in the url query.

**Proposed solution**
Use POST in the endpoint instead.

### How to test
1. Add a new issue
2. Add around 8k bytes or more of text in the text area
3. Click Preview (or view the preview on the side)
4. Confirm that the preview is showing correctly.


brianp:
I git fu-barred this PR. It won't recognize the changes anymore, but it does on #625.

PR #626, opened 2020-03-27
Editor Form View

aapomm:
### Spec
Currently, there is no way to format notes/evidence/issues etc without manually entering textile.

**Proposed solution**
Create a standard form view with the following specs:

- In the form view, fields can be added/removed as necessary.
- Switching between the form and write view should keep the content. We build the form and write views from the server to reduce JS complexity.
- When the user creates an item through a template, the form view should be pre-filled with the template info. This means that for the fields that provide options, we generate a dropdown for the users.
- When the form view is submitted by the user, the controller converts the form data to the usual text form that the Dradis items use.
- Editing the form view should reflect in the side-by-side preview.
- Auto-save should work correctly in the form view as well as the write view.

### How to test
Test out creating/editing the form in the following item views:
- Issue
- Notes
- Evidence
- Cards
- Issue merge

In the form view, you should be able to:
- Create key/values
- Remove them
- Edit the values

In every scenario, confirm the following:
- Fields can be added/removed.
- Every change should update the live preview.
- The form view saves the changes when submitted.
- Auto-save should work with changed form values, removed values, and added values.
- Navigating between form view and source view should persist changes between views.


app/views/issues/_form.html.erb
```diff
@@ -6,12 +6,11 @@
...
-          data: {
-            behavior: 'auto-save',
-            autosave_key: "#{current_project.id}-issue-#{ @issue.persisted? ? "edit-#{@issue.id}" : 'new' }",
-            help_url: markup_path,
-            preview_url: preview_path
-          }
+        data: textile_form_urls.merge({
+          autosave_key: "#{current_project.id}-issue-#{ @issue.persisted? ? "edit-#{@issue.id}" : 'new' }",
+          behavior: 'auto-save',
+          content: @issue.text
```
brianp, line 12: I don't think we need to append a `content` data attribute to a textarea. We can always reference the text area value. Managing the value in two places means there's no source of truth. We should trust the value of value :D

app/views/textile/form.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if @form_data %>
+
```
brianp, line 2: Empty line.

app/controllers/cards_controller.rb
```diff
@@ -31,6 +32,8 @@ def create
     # Set the new card as the last card of the list
     @card.previous_id = @list.last_card.try(:id)
 
+    @card.description = dradify_form if params[:item_form]
```
brianp, line 35: I don't like how multiple controllers are peppered for this addition. Is it something we could add into the `FormDradifier` module? Something like:

disclaimer: this is untested
```ruby
before_filter :set_content

def set_content
  return unless params[:item_form]

  resource = @controller.instance_variable_get("@#{instance_name}")

  if resource&.description
    'description'
  elsif resource&.content
    'content'
  elsif resource&.text
    'text'
  end

  resource.send("#{method}=", dradify_form)
end

Comment (aapomm) [reply app/controllers/cards_controller.rb:35]: @brianp  I tested this idea and noticed that it'll only work if we declare the `include` after we declare the `before_action` callbacks in the controller:
```
before_action ....
include FormDradifier
```

This is because we use the item instance (`@issue`, `@note`, etc.) in`set_content` but they only get defined in the other `before_action` callback. Moving the `FormDradifier` inclusion after the callbacks also moves the `set_content` call, and only then we'll have the item instance available for us.

Note that the same thing happens if we use `around_action` because rails pushes the callbacks sequentially: [callbacks.rb](https://github.com/rails/rails/blob/157920aead96865e3135f496c09ace607d5620dc/actionpack/lib/abstract_controller/callbacks.rb#L91)

I do like the improvement of moving this code to the module. It looks a bit cleaner but it's also inconsistent with what we have. Thoughts?

Comment (brianp) [reply app/controllers/cards_controller.rb:35]: We should add a class method into `FormDradifier` that allows us to include it into controllers at the top as normal and then we can call something like `dradify_form only: %w[create update]`

The method `dradify_form` could do one of two things

- Define the before filter for the method behind the scenes.
- Call the `set_content` method directly assuming it has been placed under other before filters.

Alternatively (and simplest): Leave the logic in the module, and move the `before_filter` from the module to the controller. Allowing the controller to decide the flow. It still requires an addition in the controllers, but I feel it's cleaner than additions per action.

Comment (aapomm) [reply app/controllers/cards_controller.rb]: Thanks for the suggestions @brianp . I went ahead and implemented the `before_action` in the controllers. I encountered some nuances in the cards controller and the issues' merge controller but overall the implementation looks much cleaner.

File: app/controllers/concerns/form_dradifier.rb [hunk 4]
```diff
@@ -0,0 +1,39 @@
...
+  def convert_form_content
+    return unless params[:item_form]
+
+    item = instance_variable_get("@#{controller_name.singularize}")
+    # Assign the issue for the Issues::MergeController
+    item = @issue if controller_name.singularize == 'merge'
+
+    content_attribute =
+      case item
+      when Card; :description
```
brianp, line 13: One might argue this block knows to much about other classes. Where the ruby way would be to use duck-typing. I'm impartial on type-safety vs duck typing arguments so I'll let you decide.

app/controllers/issues/merge_controller.rb
```diff
@@ -21,7 +21,9 @@ def create
       # create new issue if existing issue not given
       if @issue.new_record?
         @issue.author ||= current_user.email
-        if @issue.save && @issue.update_attributes(issue_params)
+        if @issue.save &&
+            (params[:issue] && @issue.update_attributes(issue_params))
```
brianp, line 25: What was the reason for this change?

app/views/evidence/_form.html.erb
```diff
@@ -7,12 +7,11 @@
       f.input :content,
       input_html: {
         class: 'textile',
-        data: {
-          behavior: 'auto-save',
+        data: textile_form_urls.merge({
+          allow_dropdown: action_name == 'new',
```
brianp, line 11:
```suggestion
          allow_dropdown: @evidence.persisted?,
```

I'd really like if we could move `action_name == 'new'` down into the view instead of having to make it a transient variable that gets passed through everything. If we can't though let make it relevant to the resource on hand.

app/views/issues/_form.html.erb
```diff
@@ -6,12 +6,11 @@
...
       input_html: {
         class: 'textile',
-          data: {
-            behavior: 'auto-save',
-            autosave_key: "#{current_project.id}-issue-#{ @issue.persisted? ? "edit-#{@issue.id}" : 'new' }",
-            help_url: markup_path,
-            preview_url: preview_path
-          }
+        data: textile_form_urls.merge({
+          allow_dropdown: action_name == 'new',
```
brianp, line 10:
```suggestion
          allow_dropdown: !@issue.persisted?,
```

app/views/notes/_form.html.erb
```diff
@@ -6,12 +6,11 @@
       label: false,
       input_html: {
         class: 'textile',
-        data: {
-          behavior: 'auto-save',
+        data: textile_form_urls.merge({
+          allow_dropdown: action_name == 'new',
```
brianp, line 10:
```suggestion
          allow_dropdown: !@note.persisted?,
```

app/controllers/issues/merge_controller.rb
aapomm: For context, we build the form view params like: `{ item_form: [params] }` while we have the write view params as: `{issue: { text: ''} }`. I'd like the `item_form` to also be under the `{ issue: ... }` but this doesn't work in rails by default because it expects the `issue` instance to respond to `item_form` and to me, adding virtual attributes seems excessive. 

So, this means that when passing the form view for the issue merge, `params[:issue]` will not exist and `issue_params` will throw an error when invoked. We only try to call the `update_attributes` when the issue params exist. Otherwise, we used the "dradified" form.

Note that this does not happen in other controllers since we also pass other parameters even if the user uses the form view. For example, if the user submits the form view in the evidence, the evidence's issue is also passed so the overall params looks something like:
```
{ item_form: [data], evidence: { issue_id: ... } }
```

app/views/textile/form.js.erb
```diff
@@ -0,0 +1,22 @@
...
+
+$('[data-behavior~=delete-field]').click( function(){
+  $(this).closest('[data-behavior~=textile-form-field]').remove();
+});
+
+$('[data-expand~=auto]').each( function() {
+  $(this).css({'padding': '0.375rem 0.75rem', 'height': this.scrollHeight});
+});
+
+$('[data-expand~=auto]').on('keyup', function(e) {
```
brianp, line 14: There's a lot happening here. I can't follow the purpose.

Textareas have their height set to scrollHeight. and then to 1px, and then back to scrollheight?
What's the reasoning?

aapomm: Gonna have to phone a friend for this one. @MattBudz ?

app/controllers/concerns/form_dradifier.rb
aapomm: I'd like to go with adding an alias method for the models, but I remember @etdsoft not being a fan of this approach because ideally we should have a uniform naming for the model attributes.
I could also go with the approach of the controllers declaring this attribute but this ties in with our previous conversation that controllers are doing too much just to include this module.
So, I'm electing to keep this as is.

app/views/textile/form.js.erb
MattBudz: This is for cross-browser compatibility for automatically expanding and shrinking textareas as the user types. If the height isn't set it to 1px first then the browser can't set it to the scroll height properly and it also prevents the textbox from shrinking when the user is backspacing.

config/routes.rb
```diff
@@ -121,8 +121,11 @@
 
   # -------------------------------------------------------------- Static pages
   # jQuery Textile URLs
-  post '/preview' => 'home#textilize',  as: :preview
-  get '/markup-help' => 'home#markup_help', as: :markup
+  get '/field', to: 'textile#field', as: :field, defaults: { format: :js }
```
etdsoft, line 124: Isn't there a RESTful way to approach this other than 5 different individual entries?

spec/features/evidence_spec.rb
```diff
@@ -79,13 +79,14 @@
   end
 
 
-  describe "edit page" do
+  describe "edit page", js: true do
```
etdsoft, line 82: quotes

spec/features/evidence_spec.rb
```diff
@@ -115,8 +125,11 @@
       it_behaves_like "a page which handles edit conflicts"
     end
 
-    describe "submitting the form with invalid data" do
-      before { fill_in :evidence_content, with: "a"*65536 }
+    pending "submitting the form with invalid data" do
```
etdsoft, line 128: 1. why pending?
2. quotes

spec/features/evidence_spec.rb
```diff
@@ -128,8 +141,8 @@
   end
 
 
-  describe "new page" do
-    let(:content) { "This is example evidence" }
+  describe "new page", js: true do
```
etdsoft, line 144: quotes

spec/features/evidence_spec.rb
```diff
@@ -184,7 +205,7 @@
         end
       end
 
-      describe "submitting the form with invalid information" do
+      context "submitting the form with invalid information" do
```
etdsoft, line 208: please fix all the quotes in the file.

app/controllers/notes_controller.rb
```diff
@@ -3,13 +3,15 @@
...
   include ConflictResolver
+  include FormDradifier
   include Mentioned
   include MultipleDestroy
   include NodesSidebar
   include NotificationsReader
 
   before_action :find_or_initialize_note, except: [:index, :new, :multiple_destroy]
   before_action :initialize_nodes_sidebar, only: [:edit, :new, :show]
+  before_action :convert_form_content, only: [:create, :update]
```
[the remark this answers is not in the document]
brianp: We've got some [context around this](https://github.com/dradis/dradis-ce/pull/626#discussion_r401600473) that lead to what you see now.

The idea of a form object doesn't solve the problem of having to use it repetitively. It sounds like a step in addition to, not something in replacement of. 
Typically the form object would be used in setting the initial form, and maybe serializing the data, which isn't the repetitiveness here. Here the problem is strictly serializing the form into data we want to persist, and we have to do that for 4 different record types, which all have their own controller. We had originally thought to add the before action in the module which created an order of operations issue when before actions are defined. 

I've also been looking at this code for days now, so if you've got a suggestion I'm all ears. Right now I don't have a clear path to reduce these calls unless we want to go virtual attribute that accepts the form fields, and uses a service included into the model to translate the data... 🤔

config/routes.rb
```diff
@@ -121,8 +121,15 @@
 
   # -------------------------------------------------------------- Static pages
   # jQuery Textile URLs
-  post '/preview' => 'home#textilize',  as: :preview
-  get '/markup-help' => 'home#markup_help', as: :markup
+  resource :textile, controller: :textile, only: [] do
+    collection do
```
brianp, line 125: This `collection do` block doesn't actually do anything if we're still specifying `get` `post` inside it.

app/controllers/textile_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+    @allow_dropdown = JSON.parse(params[:allow_dropdown] || 'false')
+  end
+
+  # Returns the markup cheatsheet that is used by the jQuery.Textile plugin Help
+  # button.
+  def markup_help
+    render layout: false
+  end
+
+  # Returns the source test given a form data
```
etdsoft, line 19:
```suggestion
  # Returns the source text given a form data
```

app/controllers/textile_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  private
+
+  # Convert serialized form data to Dradis-style item content
+  def convert_to_source(form_data)
+    form_data.each_slice(2).map do |field_name, field_value|
+      field = field_name[1]
+      value = field_value[1]
+      next if field.empty? || (field.empty? && value.empty?)
```
etdsoft, line 43: how can the second `field.empty?` ever be triggered?

app/controllers/textile_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      "#[#{field}]#\n#{value}"
+    end.compact.join("\n\n")
+  end
+
+  # Reformatted form parameters to be dradified
+  def form_params
+    JSON.parse(params[:form]).map do |field|
```
etdsoft, line 51: why are we doing our own JSON parsing here? Should we be using the Rails :json format facilities?

app/helpers/textile_helper.rb
```diff
@@ -0,0 +1,10 @@
+module TextileHelper
+  def textile_default_data
```
etdsoft, line 2: Still this method name doesn't describe what this method is for, can we please do our future selves a favour and be more thoughtful about our own naming decisions?

app/views/comments/_form.html.erb
```diff
@@ -1,10 +1,13 @@
...
+<%= form_with model: [current_project, @comment] do |f| %>
   <%= f.text_area :content,
                   value: @comment.content,
                   rows: 5,
                   required: 'required',
                   class: 'form-control mt-3',
-                  data: { behavior: 'mentionable' } %>
+                  data: {
+                    behavior: 'mentionable rich-toolbar',
+                    'rich-toolbar': 'block-code bold italic link list-ol list-ul table'
```
etdsoft, line 9: I think this attribute name could be more descriptive, perhaps `rt-buttons` or something? Also, what about nesting multiple?

```ruby
data: {
  behavior: 'mentionable rich-toolbar',
  rt: {
    buttons: 'block-code bold italic'
  }
}
```

Would Rails turn that into: `data-rt-buttons`?

app/views/cards/_form.html.erb
```diff
@@ -7,9 +7,14 @@
   <%=
     f.input :description,
       input_html: {
-        class: 'textile',
+        class: 'textile h-100',
```
etdsoft, line 10: I think comments should also get `h-100`

etdsoft: @MattBudz what's `h-100` about and should it be added to the comment's box?

app/controllers/textile_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+    render plain: convert_to_source(form_params)
+  end
+
+  # Returns the Textile version of a text passed as parameter
+  def textilize
+    @text =
+      if params[:form]
+        convert_to_source(form_params)
+      else
+        params[:text]
```
etdsoft, line 30: When is this path taken?

aapomm: This is triggered when the Source view is being used,

app/views/comments/_form.html.erb
brianp: Posted in another thread but it won't nest as we would like it to.

PR #632, opened 2020-04-01
Editor toolbar

MattBudz:
### Summary

This PR add's an editor toolbar to Dradis.

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/assets/javascripts/shared/editor_toolbar.js
+++ b/app/assets/javascripts/shared/editor_toolbar.js
@@ -192,7 +192,7 @@ class EditorToolbar {
   injectMarkdown($element, prefix, suffix, placeholder) {
     var adjustedPrefixLength = prefix.length, adjustedSuffixLength = suffix.length;
     var startIndex = $element[0].selectionStart, endIndex = $element[0].selectionEnd;
-    var elementText = $element.val(), selectedText = $element.val().substring(startIndex, endIndex).split('\n');
+    var selectedText = $element.val().substring(startIndex, endIndex).split('\n');
     var markdownText = '';
 
     // create string to inject with markdown added
@@ -213,16 +213,15 @@ class EditorToolbar {
     });
 
     // remove the original selection (if there was one) and add new markdown string in it's place
-    $element.val(elementText.slice(0, startIndex) + markdownText + elementText.slice(endIndex)); 
-
+    $element.focus(); // bring focus back to $element from the toolbar
+    document.execCommand("insertText", false, markdownText);
+    
     // post-injection cursor location
     if (startIndex == endIndex) { // no text was selected, select injected placeholder text
       $element[0].setSelectionRange(startIndex + prefix.length, startIndex + markdownText.length - suffix.length);
-      $element.focus();
     }
     else { // text was selected, place cursor after the injected string
       $element[0].setSelectionRange(adjustedPrefixLength + endIndex + adjustedSuffixLength, adjustedPrefixLength + endIndex + adjustedSuffixLength);
-      $element.focus();
     }
   }
 }
--- a/app/assets/javascripts/shared/editor_toolbar.js
+++ b/app/assets/javascripts/shared/editor_toolbar.js
@@ -86,21 +86,19 @@ class EditorToolbar {
         endIndex = $element[0].selectionEnd,
         elementText = $element.val(),
         selectedText = $element.val().substring(startIndex, endIndex).split('\n'),
-        markdownText = affix.asString;
+        markdownText = affix.asPlaceholder;
 
     // create string to inject with markdown added
-    selectedText.map(function(selection, index) {
-      if (selection !== '') {
-        markdownText = affix.withSelection(selection);
-        
-        // if not the last line of selection, add new line and account for prefix/suffix length injected on that line
-        if (index < selectedText.length - 1) {
-          markdownText += '\n';
-          adjustedPrefixLength *= 2;
-          adjustedSuffixLength *= 2;
-        }
-      }
-    });
+    markdownText = selectedText.filter(Boolean).reduce(function(text, selection, index) {
+      return (index == 0 ? affix.withSelection(selection) : text + '\n' + affix.withSelection(selection))
+    }, '');
+
+    // Adjust the lengths based on number of lines selected - the last line.
+    if (selectedText.length > 1) {
+      var adjustment = (selectedText.length-1);
+      adjustedPrefixLength *= adjustment;
+      adjustedSuffixLength *= adjustment;
+    }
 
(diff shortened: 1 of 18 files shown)


app/assets/javascripts/shared/editor_toolbar.js
```diff
@@ -0,0 +1,232 @@
...
+    // find all child input text and textarea elements
+    this.$target.find('input[type="text"], textarea').each(function() {
+      // wrap each element in a div and add toolbar element as sibling
+      $(this).wrap('<div class="editor-field" data-behavior="editor-field"></div>');
+      $(this).parent().append('<div class="editor-toolbar" data-behavior="editor-toolbar"></div>');
+    });
+
+    // create toolbar buttons for each toolbar
+    $('[data-behavior~=editor-toolbar]').each(function() {
+      if ($(this).prev('textarea').length) { // toolbar for textarea elements
```
brianp, line 30: Not a huge fan of these big blocks inside conditionals. Feels like we could split them into option groups or methods that return this content when asked for.

app/assets/javascripts/shared/editor_toolbar.js
```diff
@@ -0,0 +1,232 @@
...
+    });
+
+    // when a toolbar button is clicked
+    $('[data-btn]').click(function () {
+      var $element = $(this).parents('[data-behavior~=editor-field]').children('textarea, input[type=text]');
+      var prefix = '', suffix = '', placeholder = '';
+  
+      // set the appropriate prefix, placeholder and suffix depending on which button was clicked
+      switch ($(this).data('btn')) {
+        case 'header':
```
brianp, line 116: Same idea here. Big case statements make my brow sweat. 

Instead we could define a dictionary of element data. In this case it could be a plain javascript object with the appropriate naming.

```javascript
affixes = {}
affixes['bold'] = {'prefix': '*', placeholder: 'suffix', 'suffix': '*'}
// cont...
```

then instead of the case statement we can simply reach into the dictionary for the settings:
```javascript
affixes[$(this).data('btn')].prefix
> *
```

app/assets/javascripts/shared/editor_toolbar.js
```diff
@@ -0,0 +1,232 @@
...
+        $('[data-btn~=italic]').click();
+      }
+      else if (e.metaKey && key === 75 ) { // 75 = k
+        e.preventDefault();
+        $('[data-btn~=link]').click();
+      }
+    };
+  }
+
+  injectMarkdown($element, prefix, suffix, placeholder) {
```
brianp, line 192: We're not using Markdown. We're using Textile.

vendor/assets/javascripts/jquery.textile.js
```diff
@@ -27,7 +27,7 @@
         resize: true,
         // HTML templates
         tpl: {
-          wrap: '<div class="textile-wrap"><ul class="textile-toolbar"></ul><div class="textile-inner row"></div></div>',
+          wrap: '<div class="textile-wrap"><ul class="textile-toolbar"></ul><div class="textile-inner row" data-behavior="textile-inner"></div></div>',
```
brianp, line 30: @MattBudz is "textile-inner" really a behaviour. As far as semantics go it doesn't tell me what the _behaviour_ *is*. I know we're clearly binding it to something but I wouldn't know what.

app/assets/javascripts/shared/editor_toolbar.js
```diff
@@ -0,0 +1,233 @@
...
+    // Trigger a change event because javascript manipulation doesn't trigger
+    // them. The change event will reload the preview
+    $element.trigger('textchange');
+  }
+
+  affixesLibrary() {
+    return {
+      'block-code':  new BlockAffix('bc. ', 'Code markup'),
+      'bold':        new Affix('*', 'Bold text', '*'),
+      'header':      new Affix('#[', 'Header text', ']#'),
```
rachkor, line 121:
```suggestion
      'header':      new Affix('#[', 'Header', ']#'),
```
Suggesting to change this to a single word field name as field names without spaces are a best practice (to allow for filtering if we ever need to)

app/assets/javascripts/shared/editor_toolbar.js
```diff
@@ -0,0 +1,233 @@
...
+      'highlight':   new Affix('$${{', 'Highlighted text', '}}$$'),
+      'inline-code': new Affix('@', 'Inline code', '@'),
+      'italic':      new Affix('_', 'Italic text', '_'),
+      'link':        new Affix('"', '', '":http://'),
```
rachkor, line 125:
```suggestion
      'link':        new Affix('"'Link text, '', '":http://'),
```

rachkor: ^ Suggesting the above as it was hard for me to understand where to put the link text in the `"":http://` example.

app/assets/javascripts/tylium/behaviors.js.coffee
```diff
@@ -61,6 +61,9 @@ document.addEventListener "turbolinks:load", ->
   # Activate jQuery.Textile
   $('.textile').textile()
 
+  # Activate Editor Toolbar after Textile
+  new EditorToolbar($('[data-id~=textile-inner], [data-behavior~=comment-feed]'));
```
etdsoft, line 65: This seems like a "hard-coded" approach? It's not a behaviour that we sprinkle around the DOM, but a hard-coded list of elements we want to apply the behaviour to.

I'd say `data-behaviour="rich-editor"` and then apply that anywhere we want it? Then we can make the component that renders "Write/Form-View" to include that behaviour in the rendered items it manages.

vendor/assets/javascripts/jquery.textile.js
```diff
@@ -27,7 +27,7 @@
         resize: true,
         // HTML templates
         tpl: {
-          wrap: '<div class="textile-wrap"><ul class="textile-toolbar"></ul><div class="textile-inner row"></div></div>',
+          wrap: '<div class="textile-wrap"><ul class="textile-toolbar"></ul><div class="textile-inner row" data-id="textile-inner"></div></div>',
```
etdsoft, line 30: using an "ID" as a front-end behaviour is against the style guide.

app/assets/javascripts/tylium/behaviors.js.coffee
brianp: Yup this pattern sounds familiar 😄 

Updated with new changes.

vendor/assets/javascripts/jquery.textile.js
brianp: Resolved with new entry point changes.

PR #637, opened 2020-04-06
Form view preview update

aapomm:


app/views/textile/_js_handlers.js.erb
```diff
@@ -0,0 +1,32 @@
+$(function() {
```
brianp, line 1: I don't really like this as a new file. We also have handlers and bindings going on in the textile.js already. Now we've just split them across two files without a real distinction about what's what. My current feeling is that this feels like setup, which we currently do in the `_buildContainer` function. 
I'm not against splitting it up, but we should really move everything, or move nothing. 
I'd prefer to see it all in the same file, maybe we can call an additional _bindHandlers event or similar that does all this for us.

PR #640, opened 2020-04-08
user_id for the activities table

aapomm:
### Spec
Currently, the activities table has a user column that's a string. This is because when the Activity feature was implemented, there is no User model yet. Now that we have a User model implemented, there's no reason to have this column. 

**Proposed solution** 
Remove the divergence from pro and add a user reference to the activities table

### How to test:
This is a refactor and no new feature is implemented, so only the current activities behavior should be tested.


app/presenters/activity_presenter.rb
```diff
@@ -2,7 +2,7 @@ class ActivityPresenter < BasePresenter
   presents :activity
 
   def avatar_with_link(size)
-    h.link_to(avatar_image(User.new(email: activity.user), size: size), 'javascript:void(0)')
+    h.link_to(avatar_image(User.new(email: activity.user.email), size: size), 'javascript:void(0)')
```
etdsoft, line 5: This seems too roundabout, why is it needed?

```suggestion
    h.link_to(avatar_image(activity.user, size: size), 'javascript:void(0)')
```

PR #643, opened 2020-04-10
Add note title to breadcrumb

sean-yeoh:
### Summary
Currently, notes and evidence don't have complete breadcrumbs. This PR adds note and evidence title on the breadcrumb.

### How To Test

**For note**
1. Visit a note inside a node
2. Check if breadcrumbs contain the note title
3. Check if breadcrumbs contain the link 'Notes'
4. Check if clicking the 'Notes'  breadcrumb will redirect to the notes index page of the node

**For evidence**
1. Visit a note inside a evidence
2. Check if breadcrumbs contain the evidence title
3. Check if breadcrumbs contain the link 'Evidence'
4. Check if clicking the 'Evidence'  breadcrumb will redirect to the evidence index page of the node

- [x] Added a CHANGELOG entry


CHANGELOG
```diff
@@ -10,6 +10,7 @@ v3.17 (Month 2020)
   * Remove tag color from issue titles in issue summary
   * Use shared noscript partial
   * Upgraded gems: puma, sass-rails
+  * Display note and evidence title on breadcrumb
```
MattBudz, line 13:
```suggestion
  * Display note and evidence titles in breadcrumbs
```

MattBudz: please also move this entry up in the changelog so it is in alpha order.

sean-yeoh: Thanks for the suggestion @MattBudz. Will update it.

spec/support/node_breadcrumbs_shared_examples.rb
```diff
@@ -0,0 +1,53 @@
...
+shared_examples 'nodes pages breadcrumbs' do |action, klass|
+  it 'displays breadcrumbs' do
+    expect(page).to have_selector('.breadcrumb li a', text: 'Nodes')
+    expect(page).to have_selector('.breadcrumb li a', text: @node.label)
+    expect(page).to have_selector('.breadcrumb li a', text: klass.to_s.pluralize)
+  end
+
+  it 'redirects to node page when node label breadcrumb is clicked' do
+    find('.breadcrumb li a', text: @node.label).click
+    expect(page).to have_current_path("/projects/#{current_project.id}/nodes/#{@node.id}")
```
aapomm, line 25: Can we use path helpers for this instead of a hard-coded path?

spec/support/node_breadcrumbs_shared_examples.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  it 'redirects to node page with tab params when node label breadcrumb is clicked' do
+    find('.breadcrumb li a', text: klass.to_s.pluralize).click
+    expect(page).to have_current_path(
+      "/projects/#{current_project.id}/nodes/#{@node.id}?"\
```
aapomm, line 31: ditto

app/views/layouts/nested_node_resource.html.erb
```diff
@@ -1,23 +1,3 @@
-<% content_for :breadcrumbs do %>
```
aapomm, line 1: What's the reason for moving this away from the layout? Looking at both the `_breadcrumbs` partials for Notes and Evidence, it seems to be very similar that they can be put into one file. The advantage of this is that we can reduce the `render partial: 'breadcrumbs'` declarations.

spec/features/evidence_spec.rb
```diff
@@ -76,6 +76,9 @@
       include_examples "recover deleted item", :evidence
       include_examples "recover deleted item without node", :evidence
     end
+
+    let(:model) { @evidence }
+    include_examples "nodes pages breadcrumbs", :show, Evidence
```
aapomm, line 81:
```suggestion
    include_examples 'nodes pages breadcrumbs', :show, Evidence
```
Use single quotes here. Also make this adjustment for the changes below.

spec/support/node_breadcrumbs_shared_examples.rb
sean-yeoh: Thanks, I've changed it!

spec/support/node_breadcrumbs_shared_examples.rb
sean-yeoh: Thanks, I've changed it!

spec/features/evidence_spec.rb
sean-yeoh: Fixed!

app/views/layouts/nested_node_resource.html.erb
sean-yeoh: I decide to move it away from this layout because Nodes, Notes and Evidence are using the same breadcrumbs initially. I can make it into one file, but it'll be like something below.

```
if controller == 'nodes'
  render 'node' breadcrumb partial
elsif controller == 'notes'
  render 'notes' breadcrumb partial
else
  render 'evidence' breadcrumb partial
end
```

aapomm: Adding the following to this file seems to work:
```
<% content_for :breadcrumb do %>
  <%= render 'breadcrumbs' %>
<% end %>
```
without adding a bunch of elsifs. Can you confirm? 

Also, we can move the `render 'nodes/breadcrumb_dropdown'` here too.

sean-yeoh: Thanks @aapomm , it works like a charm!

PR #651, opened 2020-04-21
Add a default user with the welcome setup data

brianp:
### Summary
Previously we did not need to create a default user with the welcome data.
With a recent [activities change](https://github.com/dradis/dradis-ce/commit/fce25affc1739071af519efee535e0bef19a7d3f) we moved from using a denormalized email value to a relation. The relation can not be null and requires a valid id. So now we create a default user along with the welcome setup to satisfy the requirement.

Currently in master running `ruby bin/setup` fails because the template in dradis:setup:welcome can not find a user to apply the fake activity to so the setup errors out.

### Check List

- [x] Added a CHANGELOG entry


etdsoft:
I don't love this approach: creating a fake user in order for the fake activity to show up.

I think we'd be better off with a better Activity Feed blank slate that gives a sense of what the feed would look like instead of adding a bunch of random users and activities.

Maybe I'm overthinking it though.

> This also solves the current issue where the template in the welcome setup can not find a user and the setup errors out.

Can you please expand?

brianp:
@etdsoft Description line expanded.

lib/tasks/thor/setup.rb
```diff
@@ -83,6 +83,7 @@ def welcome
         default_user_id: 1
       )
 
+      User.create(email: 'adama@dradisframework.com')
```
etdsoft, line 86: This seems to belong before the previous call, so L83 can be replaced with the actual `user.id` which probably will be 1 all the time anyway, but it seems a bit weird to invert the order...

CHANGELOG
```diff
@@ -16,6 +16,7 @@ v3.17 (Month 2020)
   * Remove tag color from issue titles in issue summary
   * Use shared noscript partial
   * Use user model reference for activities instead of user email
+    - Add a default user during welcome setup to satisfy model relations
```
etdsoft, line 19: this is an "internals" type of change, I don't think it's a user-facing CHANGELOG worthy one, no?

brianp:
Tested 👍🏻

PR #656, opened 2020-04-25
Server side auto save

brianp:
### Summary

Implement server side auto save. When the editor is opened initialize a web sockets channel for the editor. Every time a change occurs in the editor send the content through the socket. The socket will update the resource, create a papertrail revision marked as `auto-save` and return the new `updated_at` time to update the form with. This will help keep the revision conflict resolver operating as it currently does.

### Other information
This does not collapse auto save revisions. So expect there to be many auto save's in the revision history, as well as auto save's in the resource activity.

### Check List

- [x] Added a CHANGELOG entry


app/assets/javascripts/tylium/modules/server_auto_save.js
```diff
@@ -0,0 +1,75 @@
...
+(function($, window) {
+  function ServerAutoSave(form) {
+    this.form = form;
+
+    this.projectId = form.dataset.asProjectId;
+    this.resourceType = form.dataset.asResourceType;
+    this.resourceId = form.dataset.asResourceId;
+    this.originalUpdatedAt = form.querySelector('[data-behavior~=updated-as-auto-save]');
+    this._doneTypingInterval = 500;
+    this._autoSaveTimedInterval = 60000;
```
aapomm, line 16: Are we still using this?

app/assets/javascripts/tylium/modules/server_auto_save.js
```diff
@@ -0,0 +1,75 @@
...
+// receive timestamps back from the server to update the `original_updated_at`
+// field which helps us with conflict resolution.
+(function($, window) {
+  function ServerAutoSave(form) {
+    this.form = form;
+
+    this.projectId = form.dataset.asProjectId;
+    this.resourceType = form.dataset.asResourceType;
+    this.resourceId = form.dataset.asResourceId;
+    this.originalUpdatedAt = form.querySelector('[data-behavior~=updated-as-auto-save]');
```
aapomm, line 14: `alpha`

app/channels/editor_channel.rb
```diff
@@ -0,0 +1,59 @@
...
+    reject and return unless find_resource(params)
+
+    stream_for [current_user, current_project, resource]
+  end
+
+  def save(params)
+    resource.paper_trail_event = AUTOSAVE_EVENT
+
+    if resource.update_attributes parsed_params(params['data'])
+      track_activity(resource, AUTOSAVE_EVENT)
```
aapomm, line 23: I don't think we need to track the activity for the auto-save. On the other hand, revisions are okay once we have a way to collapse the revisions.

One idea is to have an activity that announces that someone started editing a resource like: `admin@securityroots.com started editing Issue #1` instead. But, I don't think the current implementation of activity permits this, so I'm inclined to skip activities for now.

app/assets/javascripts/tylium/modules/server_auto_save.js
```diff
@@ -0,0 +1,75 @@
...
+        resource_id: this.resourceId,
+        resource_type: this.resourceType
+      },{
+        connected: function() {
+          console.log('Subscribed to EditorChannel');
+        },
+        rejected: function() {
+          console.log('Error subscribing to EditorChannel');
+        },
+        save: function() {
```
aapomm, line 36: `alpha`

app/channels/editor_channel.rb
```diff
@@ -0,0 +1,59 @@
...
+  private
+
+  def find_resource(params)
+    return unless %w[evidence issue note card].include? params['resource_type']
+
+    if params['resource_type'] == 'card'
+      authorized_list_ids = List.where(board_id: current_project.boards.select(:id)).select(:id)
+      @resource ||= Card.find_by(id: params['resource_id'], list_id: authorized_list_ids)
+    else
+      @resource ||= current_project.send(params['resource_type'].pluralize).find_by id: params['resource_id']
```
aapomm, line 38: Doesn't look safe calling `send` with `params`. Can we use `parsed_params` here?

app/channels/editor_channel.rb
```diff
@@ -0,0 +1,59 @@
...
+      require(@params['resource_type']).
+      permit(permissable_params(@params['resource_type']))
+  end
+
+  def permissable_params(type)
+    case type
+    when 'evidence' then %i[author content issue_id node_id]
+    when 'issue' then %i[tag_list text]
+    when 'note' then %i[category_id text node_id]
+    when 'card' then %i[name description due_date assignee_ids]
```
aapomm, line 55: `alpha`

app/channels/editor_channel.rb
```diff
@@ -0,0 +1,59 @@
...
+    ActionController::Parameters.
+      new(wrapped_params).
+      require(@params['resource_type']).
+      permit(permissable_params(@params['resource_type']))
+  end
+
+  def permissable_params(type)
+    case type
+    when 'evidence' then %i[author content issue_id node_id]
+    when 'issue' then %i[tag_list text]
```
aapomm, line 53: We can add `state` to the list for the issue states.

app/channels/editor_channel.rb
```diff
@@ -0,0 +1,59 @@
...
+    stream_for [current_user, current_project, resource]
+  end
+
+  def save(params)
+    resource.paper_trail_event = AUTOSAVE_EVENT
+
+    if resource.update_attributes parsed_params(params['data'])
+      track_activity(resource, AUTOSAVE_EVENT)
+
+      self.class.broadcast_to([current_user, current_project, resource], resource.reload.updated_at.to_i)
```
aapomm, line 25: Can you fill me in on what we're broadcasting here?

app/channels/editor_channel.rb
```diff
@@ -0,0 +1,59 @@
+# The Editor Channel is opened when a user navigates to the `Edit` page of an
+# Evidence, Issue, Note, or Card. The channel is closed when the user navigates
```
aapomm, line 2: Are we closing the channel or simply unsubscribing the user? The difference can have an implication to how we would handle conflict management

app/assets/javascripts/tylium/modules/server_auto_save.js
brianp: No we are not. I'll remove it.

app/channels/editor_channel.rb
brianp: Oh you're so right!

All the auto-save activity would just be straight noise. It's not useful to anyone. I'll pull this out asap.

app/channels/editor_channel.rb
brianp: I'm going to say it is safe. On line [#32](https://github.com/dradis/dradis-ce/pull/656/files#diff-3c89c0c2231ade55a853b21327ca6785R32) I do a check of approved possible values. We only let in what we want to let in.

app/channels/editor_channel.rb
brianp: Like:  `when 'issue' then %i[tag_list text state]` ?

app/channels/editor_channel.rb
brianp: it's the updated_at timestamp. Each form has an `original_updated_at_time` used for conflict resolution. So every time autosave happens we broadcast back the updated_at timestamp so we can update it in the form. Then when the user hits save and conflict resolution happens it'll work as expected.

app/channels/editor_channel.rb
brianp: > The difference can have an implication to how we would handle conflict management
Can you explain. I'm not sure how it would make a difference.

I'd say it's actually just unsubscribing them though. The channel would technically still be available for someone else to subscribe to.

Note: a subscription to this channel can't be "shared". IE two users who subscribed to the same channel, with the same resource will not receive eachother updates. They'll both be updating the same record yes, but the broadcasting will always be specific to the user who performed an action.
IE if you and I were both updating the same issue.
I make a change
 - the issue updates
 - the channel broadcasts the updated time
 - only I receive the broadcast
Your editor does not know that a change has occured.

We do no real time conflict management. Conflict resolution happens exactly as it does now. When a user saves, they will see a conflict warning.

aapomm: Thanks for the explanation! My original thought was: if we're not closing the channel, we can broadcast an error/alert message to other users subscribed to the channel that a changed as been made. Though, it seems we're kinda doing something similar with the `updated_at` time in your other comment. 👍

app/channels/editor_channel.rb
brianp: Seems worthy of more in file comments. I'm going to update with that.

app/channels/editor_channel.rb
brianp: We definitely have an opportunity to broadcast conflicts in real time. Could be a good enhancement if we have time.

app/channels/concerns/paper_trail_activity.rb
```diff
@@ -0,0 +1,17 @@
+# This module exists to assist with PaperTrail::Version properties in Channels.
+# Typically the controllers take care of this but this request process does not
+# utilize controllers. In controllers we set PaperTrail metadata in
+# ProjectScoped#info_for_paper_trail. For channels we will do it here as a
+# before_subscribe action.
+module PaperTrailActivity
```
etdsoft, line 6: PaperTrail is an implementation detail, not part of the domain model. I'm sure you can find a better name for this Class and the job it does!

app/channels/concerns/paper_trail_activity.rb
```diff
@@ -0,0 +1,17 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    before_subscribe :set_info
+  end
+
+  def set_info
```
etdsoft, line 13: if this is equivalent to `#info_for_paper_trail` let's keep the name consistent ;-)

app/channels/editor_channel.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  private
+
+  def find_resource(params)
+    return unless %w[evidence issue note card].include? params['resource_type']
+
+    if params['resource_type'] == 'card'
+      authorized_list_ids = List.where(board_id: current_project.boards.select(:id)).select(:id)
+      @resource ||= Card.find_by(id: params['resource_id'], list_id: authorized_list_ids)
```
etdsoft, line 39: This looks like 3 queries for a single fetch: board IDs, list IDs, Card

What about (not tested):

```suggestion
      @resource ||= current_project.boards.includes(:list, list: cards).lists.select{ list == params}.cards.select{ card == params}.first
```
Either way it's a big chain, but I'd say less SQL is better than more SQL?

app/channels/editor_channel.rb
```diff
@@ -0,0 +1,62 @@
...
+    else
+      @resource ||= current_project.send(params['resource_type'].pluralize).find_by id: params['resource_id']
```
etdsoft, line 41: 1. Why the use of `||=` here and in :39 ?

2. I'm not sure the use of `#send` is buying us much here. Yes, it's reducing the number of calls, but I'm not sure if we should go for readability of the code vs lines of code as a metric?

Should L35 and this be replaced with a `switch` statement? It's probably a smell anyway, but with a clear `switch` you know what you're reading right away, as opposed to figuring out the trickery.

We're also going to be adding at least ContentBlocks to the list in Pro too.

app/channels/editor_channel.rb
```diff
@@ -0,0 +1,62 @@
...
+  def parsed_params(data)
+    wrapped_params = Rack::Utils.parse_nested_query data
+    ActionController::Parameters.
+      new(wrapped_params).
+      require(@params['resource_type']).
+      permit(permissable_params(@params['resource_type']))
+  end
+
+  def permissable_params(type)
+    case type
```
etdsoft, line 54: I swear I hadn't seen the `case` here before my comment above 😇

app/channels/editor_channel.rb
etdsoft: Isn't this duplicating the standard app's Strong parameters definitions? I think that's a slippery slope. We're going to end up modifying in one place and forgetting to do so in the other.

Is there a way to reuse the strong parameter definition from each controller?

app/views/cards/_form.html.erb
```diff
@@ -1,4 +1,11 @@
-<%= simple_form_for [current_project, @board, @list, @card] do |f| %>
+<%= simple_form_for [current_project, @board, @list, @card], html: {
+  data: {
+    'as-project-id': current_project.id,
```
etdsoft, line 3: can't you nest this?

```
data: {
  as: {
    project_id: current_project.id,
    resource_id:...
  }
}
```

app/views/evidence/_form.html.erb
```diff
@@ -19,7 +26,7 @@
   </div>
 
   <% if @evidence.persisted? %>
-    <%= hidden_field :evidence, :original_updated_at, value: @evidence.updated_at.to_i %>
+    <%= hidden_field :evidence, :original_updated_at, value: @evidence.updated_at.to_i, data: { behavior: 'updated-as-auto-save' } %>
```
etdsoft, line 29: If this is an "auto save" related behaviour, let's lead with that on the behaviour name, perhaps "auto-save-updated-as"

CHANGELOG
```diff
@@ -9,9 +9,10 @@ v3.17 (Month 2020)
   * Comments now show author's name instead of email
   * Display note and evidence titles in breadcrumbs
   * Editor improvements:
-    - New form-view to edit each field individually
     - Formatting toolbar to help with markup
+    - New form-view to edit each field individually
     - Side-by-side editor preview that auto-updates
+    - Server side auto saving with revision history
```
etdsoft, line 15: :alpha 🙈

app/assets/javascripts/tylium.js
```diff
@@ -34,6 +34,7 @@
 //= require tylium/keyboard_shortcuts
 
 //= require tylium/modules/auto_save
+//= require tylium/modules/server_auto_save
```
etdsoft, line 37: is it worth doing a `/auto_save/server` and `/auto_save/local`? I think @sean-yeoh's work is adding a `/local_auto_save`

app/assets/javascripts/tylium/modules/server_auto_save.js
```diff
@@ -0,0 +1,74 @@
...
+    init: function() {
+      var that = this;
+      this.editorChannel = window.App.cable.subscriptions.create({
+        channel: 'EditorChannel',
+        project_id: this.projectId,
+        resource_id: this.resourceId,
+        resource_type: this.resourceType
+      },{
+        connected: function() {
+          console.log('Subscribed to EditorChannel');
```
etdsoft, line 30: Do we need some `<meta>` tag that lists the `Rails.environment` so we can disable/guard all these `console.log` calls to the Dev environment?

app/assets/javascripts/tylium/modules/server_auto_save.js
```diff
@@ -0,0 +1,74 @@
...
+        },
+        received: function(newUpdatedAtTime) {
+          if (that.originalUpdatedAt !== null) {
+            that.originalUpdatedAt.value = newUpdatedAtTime;
+          }
+        },
+        rejected: function() {
+          console.log('Error subscribing to EditorChannel');
```
etdsoft, line 38: `console.error`?

app/assets/javascripts/tylium/modules/server_auto_save.js
```diff
@@ -0,0 +1,74 @@
...
+// data-behavior=server-auto-save. This module creates an action cable
+// subscription bound to the editor it's initialized from. The socket will
+// accept content for resources to update, and create revisions from. It will
+// receive timestamps back from the server to update the `original_updated_at`
+// field which helps us with conflict resolution.
+(function($, window) {
+  function ServerAutoSave(form) {
+    this.form = form;
+
+    this._doneTypingInterval = 500;
```
etdsoft, line 11: I was going to suggest to move this interval to a `<meta>` somewhere so we can change the typing timeout value all over the place in one go.

However, now I'm thinking, shouldn't this be something that's triggered from the Editor?

Otherwise we have two separate JS modules (the Editor and ServerAutoSave), checking for a single input, and firing actions and Ajax/ActionCable calls on intervals after the user stops typing.

It's not the end of the world, but it seems there's something weird about this being an "editor functionality" that's separate from the editor?

Maybe having all the editor preview, field-source sync events and a couple of millis later having the auto-save events fire is not such a big deal?

app/assets/javascripts/tylium.js
brianp: I like the idea.

app/assets/javascripts/tylium/modules/server_auto_save.js
brianp: I'm okay to drop the console logs. They were in the last WS we used, so I added them here to stay consistent.

app/channels/concerns/paper_trail_activity.rb
brianp: I don't mind combining them. I'll make that change.

app/channels/editor_channel.rb
brianp: As long as we don't mind the consistent use of repetitive card, note, issue, evidence case statements _everywhere_. I'll make that change too.

app/channels/editor_channel.rb
brianp: That was the first adverse feeling I had to it as well. So here's my current thought:
It's not a slippery slope. We have duplicated text, but the domain is entirely different. Strong Params exists to control the flow of data to an entry point. Remember this functionality was removed from `attr_accessbile` and placed into controllers for more finite control at the point of entry. WebSocket channels are not controllers, it is a different entry point into our application, and may or may not share commonalities with the controller entry points. We _could_ find some dirty trick to re-use the code from the controller but I think it's not needed. We already have multiple entry points, and each controls it's own flow. The WebApp has it's own strong params, the API also declares it's own strong params, and now web sockets. It definitely has potential to be missed in future updates, but that wasn't untrue before.

That said. I know where you're coming from, but I believe it's the wrong angle.
So as it stands I have no plans to change this.

app/views/cards/_form.html.erb
brianp: Yeah I thought so too. It did not work. It mangled a stringified `Hash#inspect` into the `as` data attribute.

app/views/evidence/_form.html.erb
brianp: It was not introduced *as* auto save behavior. It's existing conflict resolution behavior that auto-save continues to play well with.

etdsoft: I'm not going to buy this one. This `data-behavior` is handled by the AutoSave code. As Such, I'd think it conventional to start with the name of the module that handles it.

app/channels/editor_channel.rb
etdsoft: What about re-using the StrongParameters syntax with `#require` and `#permit`? That way if Rails evolves something will break here and we'll remember to evolve this endpoint too.

I'm concerned about the proliferation of non-standard entry points. The API is bad enough, and the chances of anyone considering / remembering the web sockets endpoint are slimer.

brianp: You mean the `#require` and `#permit` on line 49 and 50? 😄 

But the plays into the comment on those lines, about it being hard to read. So I think moving the permissable_params case statement directly into a strong params usage might make it simpler to grok.

app/views/evidence/_form.html.erb
brianp: I'll bite on that completely.

app/channels/concerns/paper_trail_activity.rb
brianp: Removed this module and combined it into `ProjectScopedChannels`.

app/channels/concerns/paper_trail_activity.rb
brianp: Renamed method.

app/channels/editor_channel.rb
brianp: Not better.
It's actually still 3 queries
```
  Board Load (0.2ms)  SELECT "boards".* FROM "boards"
  List Load (0.1ms)  SELECT "lists".* FROM "lists" WHERE "lists"."board_id" = ?  [["board_id", 1]]
  Card Load (0.3ms)  SELECT "cards".* FROM "cards" WHERE "cards"."list_id" IN (?, ?)  [["list_id", 1], ["list_id", 2]]
```
And then requires filtering the ruby arrays. The original format is 3 queries and returns the record directly.

app/channels/editor_channel.rb
brianp: 1. Habit.
2. Send lets us use the power of ruby! Is really all. Otherwise we end up with:

```ruby
    case params['resource_type']
    when 'card'
      authorized_list_ids = List.where(board_id: current_project.boards.select(:id)).select(:id)
      @resource = Card.find_by(id: params['resource_id'], list_id: authorized_list_ids)
    when 'evidence'
      @resource = current_project.evidence.find_by id: params['resource_id']
    when 'issue'
      @resource = current_project.issues.find_by id: params['resource_id']
    when 'note'
      @resource = current_project.notes.find_by id: params['resource_id']
    end
```

Which I'm okay with. It's just repetitive.

app/views/evidence/_form.html.erb
brianp: Updated to `auto-save-updated-at`

app/assets/javascripts/tylium.js
brianp: Created `/auto_save` and moved the related server auto save js to `/auto_save/server.js`

app/assets/javascripts/tylium/modules/server_auto_save.js
brianp: I'm commented out the console log messages but left the empty functions for future reference.

app/assets/javascripts/tylium/modules/server_auto_save.js
brianp: This discussion is happening on another platform. Marking this as resolved because we will move forward with the current implementation ensuring to revisit the same methods before final merge.

app/channels/editor_channel.rb
brianp: I've refactored the channel for readability. We are still defining the allowed params in the channel, and continuing to use StrongParams. Overall the file should be easier to understand.

brianp:
duplicate pr

PR #658, opened 2020-04-27
add empty preview placeholder

MattBudz:
### Summary

This PR adds a placeholder for the preview pane when there are no fields defined. 

### Check List

### DIFF
--- a/vendor/assets/javascripts/jquery.textile.js
+++ b/vendor/assets/javascripts/jquery.textile.js
@@ -192,8 +192,10 @@
       $.post(this.$element.data('preview-url'),
         data,
         function(result) {
-          this.options.$preview.removeClass('loading-indicator')
-            .html(result);
+          this.options.$preview.removeClass('loading-indicator').html(result);
+          if (result == '\n') {
+            this.options.$preview.append('<div class="preview-placeholder"><h5>Add some fields to see a live preview here</h5></div>')
+          }
           this.options.$preview.children(':first').addClass('textile-preview');
           this._previewRendered = true;
         }.bind(this)
@@ -277,9 +279,6 @@
         this.options.$wrap.removeClass('textile-fullscreen');
         this.options.$wrap.css('width', 'auto');
         this.options.tmpspan.after(this.options.$wrap).remove();
-
-        this.options.$preview.css('height', '100%');
-        this.$element.css('height', this.options.height);
 
         // update button icon
         $btnFS.removeClass('fa-compress').addClass('fa-expand');
--- a/app/assets/stylesheets/tylium/modules/textile.scss
+++ b/app/assets/stylesheets/tylium/modules/textile.scss
@@ -214,6 +214,18 @@
       padding: 0;
       overflow: auto;
       height: 100%;
+
+      &.preview-placeholder {
+        align-items: center;
+        display: flex;
+        height: 100%;
+        justify-content: center;
+
+        h5 {
+          color: $mutedText;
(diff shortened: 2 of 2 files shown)


vendor/assets/javascripts/jquery.textile.js
```diff
@@ -192,8 +192,10 @@
       $.post(this.$element.data('preview-url'),
         data,
         function(result) {
-          this.options.$preview.removeClass('loading-indicator')
-            .html(result);
+          this.options.$preview.removeClass('loading-indicator').html(result);
+          if (result == '\n') {
+            this.options.$preview.append('<div class="preview-placeholder"><h5>Add some fields to see a live preview here</h5></div>')
```
brianp, line 197: Does the default state on line 32 not need this change as well?
Actually the entire loading-indicator stuff- does it do anything? Can we pull that out in favour of the blank slate work being applied.

If we need the blank slate in two places we'll probably have to pull that content out into a function or variable to utilize in both places.

vendor/assets/javascripts/jquery.textile.js
```diff
@@ -278,9 +280,6 @@
         this.options.$wrap.css('width', 'auto');
         this.options.tmpspan.after(this.options.$wrap).remove();
 
-        this.options.$preview.css('height', '100%');
-        this.$element.css('height', this.options.height);
```
brianp, line 282: No more auto-height?

MattBudz: @brianp this didn't do anything because the height is handling in CSS.

vendor/assets/javascripts/jquery.textile.js
MattBudz: @brianp 
> Does the default state on line 32 not need this change as well?
Nope. Loading happens first, then the preview is rendered (blank or otherwise).

> Actually the entire loading-indicator stuff- does it do anything?
Yes, it shows a loading indicator while loading is happening, once the loading done the indicator is hidden and the content is displayed. If there is no content the placeholder is rendered.

> Can we pull that out in favour of the blank slate work being applied.
No, because the loading needs to happen first. If we take that out then when slow loading happens the user will be told to add fields when really fields do exist but just haven't been rendered yet. It would cause a bad/confusing user experience.

> If we need the blank slate in two places
It is not. We only need it here in `_loadPreview` because that is executed on load and on change so it executes in both situations where it is needed. 

The flow is: 
1. Attempt to load the pane and show the loading indicator while it is loading.
2. Once it is loaded show the content and remove the loading indicator.
3. If there is no content, show the placeholder.

vendor/assets/javascripts/jquery.textile.js
```diff
@@ -192,8 +192,10 @@
       $.post(this.$element.data('preview-url'),
         data,
         function(result) {
-          this.options.$preview.removeClass('loading-indicator')
-            .html(result);
+          this.options.$preview.removeClass('loading-indicator').html(result);
+          if (result == '\n') {
```
brianp, line 196: Can the result actually be empty string and not just a new line?

We've also got to think about new line variations. `\r` vs `\n` vs `\r\n`.

brianp: I take it back. `\n` should be fine. It looks like the server side rendering is creating the new line which I think will mean it always results in a unix newline. So just checking \n should be fine. If someone on a windows machine wants to feed in a blank `\r` and have it rendering nothing that's fine with me.

PR #659, opened 2020-04-27
Issue states crud

aapomm:
### Spec
Implement CRUD for issue states

**Proposed solution**
1. Add state column to issues
2. Update issues CRUD to handle states
3. Update dradis-projects to handle states for export/import
4. Update dradis-plugins to handle states for plugin imports

### How to test

**Test CRUD**

1. Create an issue and choose a state for it
2. Confirm that the issue was created with the state
3. Update the issue state
4. Confirm that the issue was updated with the state

**Test Export/Import**

1. Add multiple issues in your project
2. Export the project
3. Open the XML file and confirm that it contains issue states
4. Import the XML in another project
5. Confirm that the issue’s state was imported correctly

**Test Plugin Upload**

1. Go to the uploads page
2. Select a plugin
3. Select a state
4. Upload a sample plugin file
5. Confirm that the created issues have the selected state


app/controllers/upload_controller.rb
```diff
@@ -57,6 +63,14 @@ def parse
 
   private
 
+  def default_issue_state
```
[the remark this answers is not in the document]
aapomm: This method also serves as a parameter validation and it checks if the passed state is among the ones we declared. I renamed the method to `validate_issue_state` so it's closer to what it actually does.

app/assets/javascripts/tylium/modules/state_menu.js
```diff
@@ -0,0 +1,37 @@
+document.addEventListener('turbolinks:load', function(){
```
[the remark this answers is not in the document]
aapomm: Seems like YAGNI to me. I don't think there's been a situation where we needed this. I do agree that we keep on adding things on our JS, but I don't see the problem if we simply scope the modules properly.

app/assets/javascripts/tylium/modules/state_menu.js
```diff
@@ -0,0 +1,37 @@
...
+    
+    $('[data-state~=' + state + ']').addClass('selected');
+
+    btnText = prefix + ' ' + verboseState + ' Issue';
+
+    $submitBtn.prop('value', btnText);
+  }
+
+  if ($('[data-behavior~=state-menu]').length) {
+    var currentState = $('[data-behavior~=issue-state]').val();
```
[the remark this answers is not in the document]
aapomm: Good point! I updated the jquery selections with scope

app/assets/javascripts/tylium/modules/state_menu.js
```diff
@@ -0,0 +1,37 @@
+document.addEventListener('turbolinks:load', function(){
+  function setBtnText(prefix, state) {
+    var btnText;
+    var $submitBtn = $('[data-behavior~=state-submit-button]');
+    var verboseState = (state == 'review') ? 'ready for review' : state;
```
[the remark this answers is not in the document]
aapomm: We don't need to capitalize here because we already have `text-transform: capitalize;` declared in the css

CHANGELOG
```diff
@@ -1,6 +1,7 @@
 v3.17 (Month 2020)
   * Add author to evidence and notes views
   * Adjusted Uploads layout to provide more visibility to the output console
+  * Boards can be renamed and deleted through their dots menu
```
etdsoft, line 4: ???

app/views/issues/_form.html.erb
```diff
@@ -23,7 +23,28 @@
   <%= render 'tag_input', f: f %>
 
   <div class="form-actions">
-    <%= f.button :submit, nil, class: 'btn btn-primary' %> or
+    <div class="btn-group dropdown state-menu" data-behavior="state-menu">
```
etdsoft, line 26: We're already using a partial in L23, probably that's a good idea here.

Maybe time to organise both inside a ./form/ folder as the're not used elsewhere? Not too fussed about this though.

app/views/issues/_table.html.erb
```diff
@@ -1,4 +1,4 @@
-<% cache ['issues-table', @columns, issues.map(&:id), @issues.map(&:updated_at).map(&:to_i).sort.last] do %>
+<% cache ['issues-table', @columns, issues.map(&:id), issues.map(&:state), @issues.map(&:updated_at).map(&:to_i).sort.last] do %>
```
etdsoft, line 1: why do we need this in the key, and under what conditions would this part of the cache key change?

app/views/issues/_table.html.erb
```diff
@@ -16,7 +16,7 @@
       </thead>
       <tbody>
         <% issues.each do |issue| %>
-        <% cache [issue, @columns, issue.affected_count, 'issues-table'] do%>
+        <% cache [issue, @columns, issue.affected_count, issue.state, 'issues-table'] do %>
```
etdsoft, line 19: wouldn't a change in the `issue.state` be part of a change in the `issue` (first parameter of the key)

app/views/issues/_table.html.erb
aapomm: Removed this one and the other cache key update. Apparently, `update_all` (in states_controller) does not call any callbacks and does not _touch_ the records. Manually updating the `updated_at` attribute in the `update_all` call fixes the issue.

app/assets/javascripts/tylium/modules/state_menu.js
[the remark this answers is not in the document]
aapomm: I've centralized this and added a method in the Issue model to provide the state name.

app/views/issues/_table.html.erb
aapomm: Not needed anymore. See comment: https://github.com/dradis/dradis-ce/pull/659#discussion_r422035116

aapomm:
Breaking this PR into smaller pieces.

PR #660, opened 2020-04-27
Move resource action links to dots-menu

MattBudz:
### Summary

Currently, we are using the dots-menu actions inconsistently across views.

This PR:
- Moves the action links for the resource listed in the breadcrumbs to the dots-menu
- Adds a `Subscribe` link to the Subscribers section.

### Check List

- [x] Added a CHANGELOG entry


app/views/subscriptions/_actions.html.erb
```diff
@@ -1,3 +1,5 @@
+<% css_class = local_assigns[:dots_menu] ? 'dropdown-item' : '' %>
```
aapomm, line 1: Is it correct to say that `dots_menu` is true for `#show` actions? If so, we can drop this local variable and use something like:
```
<% css_class = action_name == 'show' ? 'dropdown-item' : '' %>
```

this way, we won't have to pass `dots_menu` in the views anymore.

MattBudz: Yes, but that's not the reason for this. 

There subscribe/unsubscribe link is rendered in two places in `#show`. It needs to have a `dropdown-item` class in the dots-menu but it can't use that class in the subscriptions sections.

I used this local variable approach to render the partial in two slightly different ways on the same view.

aapomm: In that case, I think this should be good 👍

PR #661, opened 2020-04-29
Issue states api

aapomm:
### Spec

With the implementation of states for issues, we need to update the API to accommodate the change.

**Proposed solution**

We need to:

- Add state as an issue field in the API
- Allow to set the state when creating the issue
- Allow to set the state when updating the issue

### How to test

**Create the issue**

```
curl --location --request POST 'localhost:3000/api/issues' \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic <token>' \
--data-raw '{ "issue": { "text": "h5. Title]#\nIssue API Test Create", "state": "published" } }'
```

Confirm that the issue is created with the published state.

**Update the issue**

```
curl --location --request PUT 'localhost:3000/api/issues/<issue_id>' \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic <token>' \
--data-raw '{ "issue": { "text": "#[Title]#\nIssue API Test Update", "state": "review" } }'
```

Confirm that the issue is updated with the review state.

**Show the issue**

```
curl --location --request GET 'localhost:3000/api/issues/<issue_id>' \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic <token>'
```

Confirm that the issue is shown with the correct state.

**List the issues**

```
curl --location --request GET 'localhost:3000/api/issues' \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic <token>'
```

Confirm that the issue is listed with the correct state.


engines/dradis-api/app/controllers/dradis/ce/api/v1/issues_controller.rb
```diff
@@ -13,12 +13,14 @@ def show
...
 
       def create
-        @issue = current_project.issues.new(issue_params)
+        @issue = current_project.issues.new
         @issue.author   = current_user.email
         @issue.category = Category.issue
         @issue.node     = current_project.issue_library
 
-        if @issue.save
+        if is_valid_state? &&
```
etdsoft, line 21: I'm not sure about this construct. Validations belong to the model, and `save` should do it's thing.

I've thread through the Rails issue, and I think that we should implement a custom validator in the model that does this check and adds the error if needed. That way the controller code would feel more familiar and similar to others.

aapomm: Unfortunately, a custom validator will not work here because the `ArgumentError` is thrown _during_ assignment. This means that an issue instance will always have a valid state.

I tried an alternative and added the following method in the model, overwriting the `state` setter:
```ruby
def state=(value)
  if Issue.states.keys.include?(value.to_s) || Issue.states.values.include?(value)
    self[:state] = value
  else
    self.errors.add(:state, 'is not valid.')
    return nil
  end
end
```

This works until we get to the controller. A `valid?` call, even if the errors exist in the instance, will return true because [it actually clears the errors](https://github.com/rails/rails/blob/master/activemodel/lib/active_model/validations.rb#L336).

In the Rails issue, the Rails team member said:

>  The current focus of AR enums is to map a set of states (labels) to an integer for performance reasons. Currently assigning a wrong state is considered an application level error and not a user input error. That's why you get an ArgumentError.

So I'm inclined to think the validation in the controller is the correct path.

Note that we add model errors in the controller level to make it work seamlessly with the `render_validation_errors`. This could be done differently, but it will involve some additional work.

brianp:

PR #662, opened 2020-04-29
Form view missing fields

aapomm:
### Spec
The main problem here is that the parse_fields method does not recognize text that does not follow the field format or any text that is not preceded by a field header.

**Proposed solution**
The textile module should be able to recognize this format and add the necessary field in the inline view.

### How to test

1. Open an issue
2. Click the source view and add any text without any field headers
3. Click back on the inline view
4. Confirm that the text still appears.


app/controllers/textile_controller.rb
```diff
@@ -7,6 +7,11 @@ def field
   # Returns the form view given a source text
   def form
     @form_data = Note.parse_fields(params[:source]) if params[:source]
+
+    if fieldless_string && !fieldless_string.empty?
```
sean-yeoh, line 11: Is it okay if we use `.present?` instead of double negation of `!empty?`? `.present?` seems more intuitive

app/controllers/textile_controller.rb
```diff
@@ -52,4 +60,10 @@ def form_params
       [field['key'], field['value']]
     end || []
   end
+
+  # Field-less strings are strings that do not have a field header (#[Field]#).
+  def fieldless_string
+    # Extract all characters before a field header (#[Field]#)
+    @fieldless_string ||= params[:source][/^([\s\S]*?)(?=\n{,2}#\[.+?\]#|\z)/, 1]
```
etdsoft, line 67: this looks like a field parsing string, that a) is a duplicated that we have defined elsewhere; b) probably shouldn't be in this controller at all.

PR #668, opened 2020-05-05
Auto Save revision collapsing

brianp:
### Summary

Every time an auto-save is triggered we currently create a new revision. Instead we would like to collapse revisions to make them more manageable. We collapse revisions under three circumstances:

- When a new auto-save occurs we collapse the auto-saves into a single auto-save
- When an Update occurs we remove old auto-saves
- When Discard is clicked we revert to the last Update, and discard old auto-saves

### Other Information

This PR is build on top of the `server-caching` PR.

### Check List

### DIFF
--- a/app/channels/editor_channel.rb
+++ b/app/channels/editor_channel.rb
@@ -12,7 +12,7 @@ class EditorChannel < ApplicationCable::Channel
 
   attr_accessor :resource
 
-  AUTOSAVE_EVENT = 'auto-save'.freeze
+  AUTOSAVE_EVENT = Activity::VALID_ACTIONS[:autosave].freeze
 
   def subscribed
     reject and return unless find_resource
--- a/app/channels/editor_channel.rb
+++ b/app/channels/editor_channel.rb
@@ -24,6 +24,7 @@ class EditorChannel < ApplicationCable::Channel
     resource.paper_trail_event = AUTOSAVE_EVENT
 
     if resource.update_attributes resource_params(params)
+      RevisionCollapser.call(resource)
       self.class.broadcast_to([current_user, current_project, resource], resource.reload.updated_at.to_i)
     end
   end
--- a/app/channels/editor_channel.rb
+++ b/app/channels/editor_channel.rb
@@ -22,7 +22,7 @@ class EditorChannel < ApplicationCable::Channel
     resource.paper_trail_event = RevisionTracking::REVISABLE_EVENTS[:autosave]
 
     if resource.update_attributes resource_params(params)
-      RevisionCollapser.collapse(resource)
+      RevisionCollapser.collapse(resource, RevisionTracking::REVISABLE_EVENTS[:autosave])
       self.class.broadcast_to([current_user, current_project, resource], resource.reload.updated_at.to_i)
     end
   end
--- a/app/channels/editor_channel.rb
+++ b/app/channels/editor_channel.rb
@@ -9,6 +9,7 @@
 # of changes from other users.
 class EditorChannel < ApplicationCable::Channel
   include ProjectScopedChannels
+  include RevisionCollapsing
 
   attr_accessor :resource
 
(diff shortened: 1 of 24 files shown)


app/controllers/cards_controller.rb
```diff
@@ -85,7 +86,7 @@ def destroy
   private
 
   def card_params
-    params.require(:card).permit(:name, :description, :due_date, assignee_ids: [])
+    params.require(:card).permit(:name, :description, :due_date, assignee_ids: []).merge(updated_at: Time.now)
```
brianp, line 89: There's a comment about this in the PR.

I'd hear an argument to move it only into the `update` action though.

app/models/activity.rb
```diff
@@ -16,9 +16,15 @@ def project=(new_project); end
 
   validates_presence_of :action, :trackable_id, :trackable_type, :user
 
-  VALID_ACTIONS = %w[auto-save create update destroy recover]
-
-  validates_inclusion_of :action, in: VALID_ACTIONS
+  VALID_ACTIONS = {
+    autosave: 'auto-save',
```
brianp, line 20: I had added this here when I was creating activities with auto-save. We're no longer doing that and I wonder if that means I should refactor this back out and put this constant somewhere more appropriate.

app/controllers/cards_controller.rb
brianp: I wanted to double check, it doesn't seem to change anything in regards to `#create` actions `created_at` and `updated_at` are the same on create.

app/services/revision_collapser.rb
```diff
@@ -0,0 +1,52 @@
...
+# update to the original state before auto-save started. The incremental changes
+# can be discarded with out concern. Orignal <=> Current is what's important.
+# Anytime the previous revision is an auto-save we want to carry the original
+# state forward so we can always have it accessible to compare against. This
+# happens when the new revision is an update, or autosave.
+class RevisionCollapser
+  def self.call(resource)
+    return unless resource.versions.any? # Lots of specs run without versioning
+
+    last_revision = resource.versions.reorder('created_at DESC').first
```
etdsoft, line 22: My understanding is that `paper_trail` already returns this in order?

https://github.com/paper-trail-gem/paper_trail#1c-basic-usage

brianp: I'll double check this before I trust it. My gut says versions start from the first version created.
Which would be `created_at ASC`. I'll validate that and see if this needs changing.

brianp: Paper trail default sorting is as expected `created_at ASC`.

etdsoft: Isn't their example (created_at ASC, #last):

```
v = widget.versions.last
```

equivalent to yours (created_at DESC, #first)?

```
last_revision = resource.versions.reorder('created_at DESC').first 
```

app/channels/editor_channel.rb
```diff
@@ -12,18 +12,17 @@ class EditorChannel < ApplicationCable::Channel
...
 
     stream_for [current_user, current_project, resource]
   end
 
   def save(params)
-    resource.paper_trail_event = AUTOSAVE_EVENT
+    resource.paper_trail_event = RevisionTracking::REVISABLE_EVENTS[:autosave]
 
     if resource.update_attributes resource_params(params)
+      RevisionCollapser.collapse(resource, RevisionTracking::REVISABLE_EVENTS[:autosave])
```
etdsoft, line 25: I get this is a Channel, but in the interest of having the code in one place, wouldn't we be better off including the `RevisionCollapsing` concern here and calling `#collapse_revisions`?

It does add the background Job, and Channel is already a background channel, I know.

If you feel strongly about that, then maybe referencing this file in the RevisionCollapser form as a "source" or "caller" section, and hoping that'll be enough in future refactors for us to look here.

app/services/revision_collapser.rb
brianp: Yes. You are correct. 😆 

Change incoming.

PR #670, opened 2020-05-06
Fix v3 boards templates import and add boards spec

aapomm:
### Spec

Currently, nothing happens when creating a methodology with a v3 template.

Note that a boards v3 template is generated from a v3 projects template. See: https://github.com/dradis/dradis-projects/blob/master/lib/dradis/plugins/projects/export/v3/template.rb#L16

**Proposed solution**

Use the >=2 version check instead when importing the boards template in the controller.

### How to test

1. Add a v3 template to your templates/methodologies directory
2. Open the boards view and create a new board
3. Use the From template select and choose the template in Step 1.
4. Confirm that the board was created successfully.

### DIFF
--- a/spec/features/board_pages/board_pages_spec.rb
+++ b/spec/features/board_pages/board_pages_spec.rb
@@ -47,17 +47,14 @@ describe 'Board pages:' do
           board
           visit project_boards_path(current_project)
 
-          click_link 'Create new methodology...'
+          expect {
+            click_link 'Create new methodology...'
 
-          find('#modal-board-new', visible: true)
-          find('#template').find(:option, 'Methodology Template v3').select_option
+            find('#modal-board-new', visible: true)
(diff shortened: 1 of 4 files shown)


spec/features/board_pages/board_pages_spec.rb
```diff
@@ -38,6 +38,28 @@
...
+
+          board
+          visit project_boards_path(current_project)
+
+          click_link 'Create new methodology...'
+
+          find('#modal-board-new', visible: true)
+          find('#template').find(:option, 'Methodology Template v3').select_option
+
+          click_button 'Add methodology'
```
sean-yeoh, line 55:
```suggestion
          expect {
            click_button 'Add methodology'
          }.to change(Board, :count).by(1)
```
Aaron, I think changing the expectation to assert a board is _created_ is more concrete than just checking for the attributes of the board.

PR #672, opened 2020-05-06
Auto Save timed save

brianp:
### Summary

Currently auto-save happens after keystroke events are completed, and when navigating away. This closely mirrors the local-caching methods and isn't entirely useful. Instead lets keep auto saves to navigating away, and on an interval of every 5 minutes.

### Check List

### DIFF
--- a/app/assets/javascripts/tylium/modules/auto_save/server.js
+++ b/app/assets/javascripts/tylium/modules/auto_save/server.js
@@ -8,7 +8,7 @@
   function ServerAutoSave(form) {
     this.form = form;
 
-    this._doneTypingInterval = 500;
+    this._timedIntervalInMS = 300000; // 5 minutes in MS
     this.originalUpdatedAt = form.querySelector('[data-behavior~=auto-save-updated-at]');
     this.projectId = form.dataset.asProjectId;
     this.resourceId = form.dataset.asResourceId;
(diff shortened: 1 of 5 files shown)


spec/support/controller_macros.rb
```diff
@@ -26,6 +26,15 @@ def visit(arg)
...
 
     if RSpec.current_example.metadata[:js]
+      # When visiting a form we've already entered data into we will get the
+      # cached data prompt. We won't always, and we can't check for it. So we have
+      # to try and close the prompt every time and rescue if it errors.
+      begin
+        page.driver.browser.switch_to.alert
+        page.dismiss_prompt
+      rescue
+      end
```
aapomm, line 36: Can we omit this and use [implicit begin](https://github.com/rubocop-hq/ruby-style-guide#begin-implicit) instead?
```suggestion
      page.driver.browser.switch_to.alert
      page.dismiss_prompt
      rescue
      end
```

brianp: I'll give that a try.

Ideally if we get the local caching changes merged, this can be removed entirely because the prompt won't exist anymore

brianp: Ahhh what your suggesting is to add the rescue at the end of the method block, and that won't work. We need the method to continue execution regardless of this possible error.

It's ugly but we can remove it as soon as we merge the local caching stuff.

brianp: I have removed this completely as the prompt no longer exists.

app/assets/javascripts/tylium/modules/auto_save/server.js
```diff
@@ -4,11 +4,14 @@
...
 // field which helps us with conflict resolution.
+// Auto-saved forms are set to be saved at two points:
+//  - When the user navigates away from the page
+//  - Every 5 minutes
 (function($, window) {
   function ServerAutoSave(form) {
     this.form = form;
 
-    this._doneTypingInterval = 500;
+    this._timedIntervalInMS = 300000; // 5 minutes in MS
```
etdsoft, line 14:
```suggestion
    this._timedIntervalInMS = 3*60*1000;
```

brianp: I appreciate the readability change.

app/assets/javascripts/tylium/modules/auto_save/server.js
```diff
@@ -4,11 +4,14 @@
 // accept content for resources to update, and create revisions from. It will
 // receive timestamps back from the server to update the `original_updated_at`
 // field which helps us with conflict resolution.
+// Auto-saved forms are set to be saved at two points:
+//  - When the user navigates away from the page
+//  - Every 5 minutes
```
brianp, line 9:
```suggestion
//  - Every 3 minutes
```

PR #673, opened 2020-05-07
Add issue tags in API json builder and update specs

aapomm:
### Spec
Currently, there are no tags when fetching issues from the API.

**Proposed Solution**
Add tags in the issue json builder.

### How to test:
1. Add an issue in a project
2. Use the API endpoint to fetch the issue
3. Confirm that the issue tags was returned


spec/factories/issues.rb
```diff
@@ -4,5 +4,11 @@
     author { "factory_bot" }
     category { Category.issue }
     node { Project.new.issue_library }
+
+    factory :tagged_issue do
```
etdsoft, line 8: why `factory` and not `trait`?

aapomm: No actual reason, it's the first thing that came to mind. I do agree that trait is better here for the following reasons: 1) it can be mixed and matched with future traits we want to implement and 2) semantics.

PR #674, opened 2020-05-08
Unlimited recent activity

MattBudz:
### Summary

This PR adds the ability to view all recent activity.

### Check List

- [x] Added a CHANGELOG entry


app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -0,0 +1,22 @@
+document.addEventListener( "turbolinks:load", function(){
+
+  var $dateHeader = $('[data-behavior~=activity-date-header]')
+
+  // remove date header if the previous date header is the same
+  $dateHeader.each(function() {
+    var currentDate = $(this).find('[data-behavior~=activity-day-value]').text();
```
brianp, line 7: Scoping 👌

app/assets/stylesheets/tylium/modules/_activities.scss
```diff
@@ -39,3 +47,104 @@
...
+    }
+  }
+
+  .activities-group {
+    width: 50%;
+
+    .activity {
+      border: 1px solid transparent;
+      padding: 1rem;
+      transition: background-color 0.2s ease-in-out, border-color 0.2s ease-in-out;
```
brianp, line 71: Do we have a standard transition that gets used, and should it become a variable if yes?

app/controllers/activities_controller.rb
```diff
@@ -4,6 +4,10 @@
 class ActivitiesController < AuthenticatedController
   include ProjectScoped
 
+  def index
+    @activities = Activity.all_latest
```
brianp, line 8: This could be far to many activities. When imports from services happen and they end up created 100,000 issues, there's also 100,000 activities.

I think the approach needs to go back to the drawing board. Inifinite scroll doesn't mean we load everything up front and scroll it, it means we load the next "set" of records and append them to the list, once scrolled we make a new query, and append ~20 more items. This could have performance implications I'm worried about. 

I suppose no more than the `Issues#index`  would 🤔 /cc @aapomm

app/models/activity.rb
```diff
@@ -26,6 +26,10 @@ def project=(new_project); end
     includes(:trackable).order("activities.created_at DESC").limit(20)
   end
 
+  scope :all_latest, -> do
+    includes(:trackable).order("activities.created_at DESC")
```
brianp, line 30: To deal with the date issue we can actually perform a group_by `group_by { |item| item.created_at }` which will separate activity items into groups by date.

The above code likely needs fine tuning as it may try to sort down to the second, instead of just the Day.

MattBudz: @sean-yeoh I believe this could be removed now, correct?

sean-yeoh: Yeah, i've removed it

app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -0,0 +1,62 @@
...
+              $activitiesGroupsContainer.data('page', page);
+
+              var activityDayValues = $.map($('[data-behavior~=activity-day-value]'), function(element, index) {
+                return $(element).text();
+              })
+
+              $.each($activitiesGroups, function(index, activitiesGroup) {
+                var $activitiesGroup = $(activitiesGroup);
+                var timeElementDatetime = $activitiesGroup.find('[data-behavior~=activity-day-value]').attr('datetime');
+                var timeElementInDOM = $activitiesGroupsContainer.find(`[datetime="${timeElementDatetime}"]`);
```
aapomm, line 37:
```suggestion
                var $timeElementInDOM = $activitiesGroupsContainer.find(`[datetime="${timeElementDatetime}"]`);
```

since this is a jquery object.

app/presenters/activity_presenter.rb
```diff
@@ -1,6 +1,14 @@
 class ActivityPresenter < BasePresenter
   presents :activity
 
+  def activity_day
+    h.local_date(activity.created_at.to_date, format: '%A, %B %e %Y', data: { behavior: 'activity-day-value' })
```
aapomm, line 5: Can we use `ACTIVITIES_STRFTIME_FORMAT` here?

app/controllers/activities_controller.rb
```diff
@@ -4,6 +4,23 @@
 class ActivitiesController < AuthenticatedController
   include ProjectScoped
 
+  def index
+    @activities_groups = Activity.includes(:trackable)
```
aapomm, line 8: Let's split this into two lines instead of chaining the `group_by`:

```
activities = Activity....
@activities_groups = activities.group_by do |activity|
...
end
```

app/views/activities/index.html.erb
```diff
@@ -0,0 +1,28 @@
...
+      </li>
+      <li class="breadcrumb-item active">All Activity</li>
+    </ol>
+  </nav>
+<% end %>
+
+<div class="content-container mt-3 <%= 'pt-3' if @activities_groups.any? %>">
+  <% if @activities_groups.any? %>
+    <div class="container">
+      <div class="activities-groups-container" data-behavior="activities-groups-container" data-page="<%= params[:page].present? ? params[:page] : 1 %>" data-url="<%= project_activities_path(current_project, format: :js) %>">
```
aapomm, line 17: Let's evaluate the `data-page` at the controller level, instead of in this view. Something like:
```
@page = params[:page].present? ? params[:page] : 1
```

app/controllers/activities_controller.rb
```diff
@@ -4,6 +4,23 @@
...
+                                 .order(created_at: :desc)
+                                 .page(params[:page])
+                                 .group_by do
+
+      |activity| activity.created_at.strftime(Activity::ACTIVITIES_STRFTIME_FORMAT)
```
aapomm, line 13: Let's use a more consistent block syntax:

```
group_by do |activity|
..
end
```

spec/features/activity_pages/activity_pages_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+
+      it 'shows paginated records' do
+        expect(page).to have_selector('.activity', count: Kaminari.config.default_per_page)
+      end
+
+      it 'shows unique date headers' do
+        activities_groups = Activity.order(created_at: :desc).limit(Kaminari.config.default_per_page).group_by do
+          |activity| activity.created_at.strftime(Activity::ACTIVITIES_STRFTIME_FORMAT)
+        end
```
aapomm, line 43:
```suggestion
        activities_groups = Activity.order(created_at: :desc).limit(Kaminari.config.default_per_page).group_by do |activity|
           activity.created_at.strftime(Activity::ACTIVITIES_STRFTIME_FORMAT)
        end
```

You can also split the `activities_group` query into multiple lines instead of one chain for better readability.

spec/features/activity_pages/activity_pages_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+            sleep 0.25 # Needed due to specs failing occasionally
+          end
+        end
+
+        it 'loads more records' do
+          expect(page).to have_selector('.activity', :count => Activity.count)
+        end
+
+        it 'shows unique date headers' do
+          activities_groups = Activity.order(created_at: :desc).group_by do
```
aapomm, line 68: Ditto

app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -0,0 +1,62 @@
...
+      if (canLoadMore && !loading && viewHeight + scrollTop >= (scrollHeight - 64)) {
+        var page = $activitiesGroupsContainer.data('page') + 1;
+        var url = $activitiesGroupsContainer.data('url');
+        $('[data-behavior="activities-spinner"]').show();
+        loading = true;
+
+        $.ajax({
+          url: url,
+          data: { page: page },
+          success: function(data, status, XHR) {
```
aapomm, line 24:
```suggestion
          success: function(data) {
```

Or are we using the other 2 arguments?

spec/features/activity_pages/activity_pages_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+          times_to_scroll.times do
+            page.execute_script('$("[data-behavior=\'view-content\']").scrollTop(100000)')
+            wait_for_ajax
+            sleep 0.25 # Needed due to specs failing occasionally
+          end
+        end
+
+        it 'loads more records' do
+          expect(page).to have_selector('.activity', :count => Activity.count)
```
aapomm, line 64:
```suggestion
          expect(page).to have_selector('.activity', count: Activity.count)
```

spec/features/activity_pages/activity_pages_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+        end
+      end
+
+      describe 'infinite scroll' do
+        before do
+          times_to_scroll = (Activity.count / Kaminari.config.default_per_page.to_f).ceil
+
+          times_to_scroll.times do
+            page.execute_script('$("[data-behavior=\'view-content\']").scrollTop(100000)')
+            wait_for_ajax
```
aapomm, line 58: I'm not convinced on adding `wait_for_ajax` and `sleep` here. I believe we can use the capybara `find` to naturally wait for the loaded activities as long we know what it needs to find.

spec/rails_helper.rb
```diff
@@ -80,7 +80,7 @@
   # config.include SupportHelper,    type: :feature
   # config.include SupportHelper,    type: :request
   config.include FactoryBot::Syntax::Methods
-  # config.include WaitForAjax, type: :feature
+  config.include WaitForAjax, type: :feature
```
aapomm, line 83: See comment above

app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -0,0 +1,62 @@
...
+  if ($activitiesGroupsContainer.length) {
+    var $viewContent = $('[data-behavior="view-content"]');
+
+    $viewContent.on('scroll', function() {
+      var viewHeight = $viewContent.height();
+      var scrollTop = $viewContent.scrollTop();
+      var scrollHeight = $viewContent[0].scrollHeight;
+
+      // 64px is the sum of margin bottom and padding bottom of the content-container class
+      if (canLoadMore && !loading && viewHeight + scrollTop >= (scrollHeight - 64)) {
```
brianp, line 15: Can we make this functionality a generic class? The functionality seems useful and maybe i'm getting ahead of myself (@MattBudz ?) but if there's anywhere else we could use this (notifications?) then making a generic Infinite Scroll module would be useful.

If it's something we would like I think most of this could stay the same. Once an infinite scroll container is scoped stuff like `activities-spinner` would just be `spinner`. The module would be initialized with `[data-behavior="infinite-scroll"]` and the container would have properties for `url` etc.

app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -0,0 +1,62 @@
...
+  var loading = false;
+  var canLoadMore = true;
+
+  if ($activitiesGroupsContainer.length) {
+    var $viewContent = $('[data-behavior="view-content"]');
+
+    $viewContent.on('scroll', function() {
+      var viewHeight = $viewContent.height();
+      var scrollTop = $viewContent.scrollTop();
+      var scrollHeight = $viewContent[0].scrollHeight;
```
aapomm, line 12:
```suggestion
      var scrollTop = $viewContent.scrollTop();
      var scrollHeight = $viewContent[0].scrollHeight;
      var viewHeight = $viewContent.height();
```
alpha order

app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -0,0 +1,62 @@
+document.addEventListener('turbolinks:load', function() {
+  var $activitiesGroupsContainer = $('[data-behavior="activities-groups-container"]');
+  var loading = false;
+  var canLoadMore = true;
```
aapomm, line 4:
```suggestion
  var canLoadMore = true;
  var loading = false;
```
alpha order

app/assets/javascripts/tylium/pages/activities.js
MattBudz: The variable is storing the datetime data, not the object itself.

app/assets/javascripts/tylium/pages/activities.js
MattBudz: @brianp I think we can extract it once we find another use for it. I don't think we need to over-engineer this just in case we may need it at some point elsewhere else.

I do like this idea, I just don't think it's necessary right now.

app/assets/javascripts/tylium/pages/activities.js
sean-yeoh: Thanks, fixed!

app/presenters/activity_presenter.rb
sean-yeoh: Yeah we can!

app/assets/javascripts/tylium/pages/activities.js
sean-yeoh: No we're not, i'll remove them

spec/rails_helper.rb
sean-yeoh: Thanks, i've removed the wait_for_ajax completely and it passes

app/assets/javascripts/tylium/pages/activities.js
sean-yeoh: @brianp I think this can be easily extracted out, but i agree with @MattBudz  that we should do that when it is required in multiple pages.

app/controllers/activities_controller.rb
```diff
@@ -4,6 +4,25 @@
 class ActivitiesController < AuthenticatedController
   include ProjectScoped
 
+  def index
+    activities = Activity.includes(:trackable)
```
etdsoft, line 8: As in L28 activities should be loaded through `current_project`.

app/controllers/activities_controller.rb
```diff
@@ -4,6 +4,25 @@
...
+                         .order(created_at: :desc)
+                         .page(params[:page])
+
+    @page = params[:page].present? ? params[:page] : 1
```
etdsoft, line 12: Since we're initialising a `@page` variable anyway, let's do it before the previous call, that way you can pass it as argument to `page()`.

If this is a number, call `.to_i` on the param.

app/controllers/activities_controller.rb
```diff
@@ -4,6 +4,25 @@
...
+
+    @activities_groups = activities.group_by do |activity|
+      activity.created_at.strftime(Activity::ACTIVITIES_STRFTIME_FORMAT)
+    end
+
+    respond_to do |format|
+      format.html
+      format.js do
+        render 'activities/_activities_groups.html.erb', locals: { activities_groups: @activities_groups }
```
etdsoft, line 21: Lets use the rails convention and have an `index.js.erb` template please. We can remove this entire block and handle the render logic in the view.

Otherwise we're using `format.js` to render a `.html.erb`. We should probably also javascript-scape the result. There are other examples throughout the code where we've followed this pattern.

app/controllers/activities_controller.rb
sean-yeoh: Thanks, fixed!

app/controllers/activities_controller.rb
sean-yeoh: Got it!

app/controllers/activities_controller.rb
sean-yeoh: Done, much cleaner now :+1:

app/views/activities/index.js.erb
```diff
@@ -0,0 +1,29 @@
...
+
+<% if @activities_groups.present? %>
+  var template = '<%= j render 'activities/activities_groups', activities_groups: @activities_groups %>';
+  var $template = $(template);
+
+  $.each($template, function(index, activitiesGroup) {
+    var $activitiesGroup = $(activitiesGroup);
+    var timeElementDatetime = $activitiesGroup.find('[data-behavior~=activity-day-value]').attr('datetime');
+    var $timeElementInDOM = $activitiesGroupsContainer.find(`[datetime="${timeElementDatetime}"]`);
+    // console.log($timeElementInDOM);
```
aapomm, line 11: Remove comment

app/views/activities/index.js.erb
```diff
@@ -0,0 +1,29 @@
...
+    if ($timeElementInDOM.length) {
+      var $activitiesGroupContainer = $timeElementInDOM.parents('[data-behavior="activities-group"]');
```
aapomm, line 13: Can we use better variable naming here? It's easy to mistake `$activitiesGroupsContainer` and `$activitiesGroupContainer` and they are two very different elements. `$activitiesGroupsContainer` is the parent container while `$activitiesGroupContainer` seems to be an activities group within a specific date.

app/views/activities/index.js.erb
sean-yeoh: Done!

app/views/activities/index.js.erb
sean-yeoh: I've updated it to infinite-scroll-container instead, Thanks!

PR #675, opened 2020-05-08
Textile endpoint

aapomm:
PR to apply the review comments from: https://github.com/dradis/dradis-ce/pull/626

Checklist:

- [x] Split the textile controller
- [x] [Fix conditional](https://github.com/dradis/dradis-ce/pull/626/files/f94f810aed3108ed69baedef20e84440890058e8#r407515882)
- [x] [Fix method name](https://github.com/dradis/dradis-ce/pull/626/files/f94f810aed3108ed69baedef20e84440890058e8#r407517021) 
- [x] [Use format: js](https://github.com/dradis/dradis-ce/pull/626/files#diff-63b5ee2963c0a8d2ab786eb261a04f8cR194)
- [ ] [Data attribute in toolbar](https://github.com/dradis/dradis-ce/pull/626/files/f94f810aed3108ed69baedef20e84440890058e8#r407519151)
- [x] Remove params[:text] flow in [controller](https://github.com/dradis/dradis-ce/pull/626/files#r419285399)
- [x] [Use Rails JSON parsing](https://github.com/dradis/dradis-ce/pull/626/files#diff-14dfef9f01460f69f175a8e6bd7d143dL11)


app/views/cards/_form.html.erb
```diff
@@ -8,11 +8,12 @@
     f.input :description,
       input_html: {
         class: 'textile h-100',
-        data: textile_default_data.merge(
+        data: {
           allow_dropdown: @card.new_record?,
           behavior: 'rich-toolbar',
+          paths: editor_paths,
           'rich-toolbar': 'block-code bold field italic link list-ol list-ul table'
```
sean-yeoh, line 15: Even though this is not supposed to be in this PR, but do you mind updating the 'rich-toolbar': to rich_toolbar: instead?
`rich_toolbar: 'block-code bold field italic link list-ol list-ul table'`

app/views/issues/merge/new.html.erb
```diff
@@ -47,7 +47,7 @@
             </div>
             <div id="<%= "preview_issue_new" %>" class="accordion-body collapse">
               <div id="new_issue">
-                <%= text_area_tag "issue[text]", nil, class: 'textile form-control', data: textile_default_data %>
+                <%= text_area_tag "issue[text]", nil, class: 'textile form-control', data: { paths: editor_paths }%>
```
sean-yeoh, line 50:
```suggestion
                <%= text_area_tag "issue[text]", nil, class: 'textile form-control', data: { paths: editor_paths } %>
```
Just missing a space here

app/views/cards/_form.html.erb
aapomm: I have nothing against the change but this will affect a bunch of other parts of the app which will bloat this PR. I'm leaning on making this change in a separate PR to keep things simple. Also, this is related to Daniel's comment in the other PR https://github.com/dradis/dradis-ce/pull/626/files/f94f810aed3108ed69baedef20e84440890058e8#r407519151.

app/helpers/textile_helper.rb
```diff
@@ -1,10 +1,10 @@
 module TextileHelper
```
etdsoft, line 1: Is this the right file and module name?

app/controllers/fields_controller.rb
```diff
@@ -0,0 +1,32 @@
+class FieldsController < AuthenticatedController
+  include FieldParser
+
+  # Returns the form view given a source text
+  def form
+    @form_data = parse_fields(params[:source])
+    @allow_dropdown = JSON.parse(params[:allow_dropdown] || 'false')
```
etdsoft, line 7: If we are just checking against 'true' or 'false' let's make this a whitelist of string values.

I don't want risking JSON parsing errors by sombebody sending some random stuff via the `:allow_dropdown` parameter.

app/controllers/markups_controller.rb
```diff
@@ -0,0 +1,13 @@
+class MarkupsController < AuthenticatedController
```
etdsoft, line 1: - This is not a "collection" resources controller, seems like a singular `Markup` would be a better fit.

- if all actions of the controller are using `layout: false` maybe it should be a controller setting.

app/services/field_parser.rb
```diff
@@ -0,0 +1,16 @@
+module FieldParser
```
etdsoft, line 1: why is a module the right vehicle for this behaviour?

app/helpers/textile_helper.rb
brianp: I think the urls it generates are a good sense of direction for naming: `markup_*`

app/services/field_parser.rb
brianp: It's interesting. It's code extracted from a Module, into a Module, included in a controller, a model concern, and a library.

It sounds like a utility helper. For a utility helper a lib, or service can be an okay place for it but then it shouldn't be included like a module it should be accessed as a class method or a field parsing object than knows how to return parsed data.

The second option a `parsing object` would be more conducive to the original implementation. The `HasFields` concern was included on objects like Notes, Cards etc. objects that have fields that need parsing. We've now got data coming in to a controller that needs fields parsed. OOP would say we pass the data into an object that also `HasFields` and can return the parsed data for us, instead of trying to apply the function to different code entry points.

app/views/cards/_form.html.erb
brianp: > Even though this is not supposed to be in this PR

That's part of the "don't do too many things in a single PR issue"

Let's not change areas the PR doesn't intend to handle.

app/services/field_parser.rb
aapomm: > OOP would say we pass the data into an object that also HasFields and can return the parsed data for us, instead of trying to apply the function to different code entry points.

This is the current implementation in the controller: `Note.parse_fields(params[:source])`

But @etdsoft mentioned in the task:

> I’m not a fan of the current implementation of calling Note.parse_fields to return a Hash though (as it has nothing to do with Note). I think we’d be better of with HasFields::ClassMethods::parse_fields but it seems like a smell anyway.

I'm thinking the best middleground here is to define a _class_ method for the `HasFields` module:

```
module HasFields
  def self.parse_fields
   ..
```

and use it as `HasFields.parse_fields`, which sounds like your first option.

brianp: For the sake of conversation:

> as it has nothing to do with Note

I'd agree with that. Using a note to do the work is a misuse of `Note`.
Creating an intermediary object that also `HasFields` for this purpose is what I was thinking. 

```
class EditorData
  include HasFields

  def initialize(data)
    @data = data
  end

  def fields
    parse_fields(@data)
  end
end

Now we have an object that can act like the other `HasFields` objects, but stands alone in purpose, and description.

Additionally, I'm not opposed to `parse_fields` becoming a class method. The fact it only operates on non-internal data means it was never really an object method in the firstplace.

File: app/controllers/fields_controller.rb [hunk 7]
```diff
@@ -0,0 +1,30 @@
+class FieldsController < AuthenticatedController
+  # Returns the form view given a source text
+  def form
+    @form_data = HasFields.parse_fields(params[:source])
+    @allow_dropdown = params[:allow_dropdown] == 'true'
```
brianp, line 5: I'm trying really hard to figure out a way around this. It's scattered in so many places. I can see why, I totally get why it's here I just wish we didn't need it.

Just to prevent how we're currently spreading it through each view etc. Why not just reference it directly?

`<% @allow_dropdown && options = value.split(' | ') %>`

The partials have access to the variable it would tidy up the forced passing through local assigns.

app/helpers/markup_helper.rb
```diff
@@ -0,0 +1,10 @@
+module MarkupHelper
+  def editor_paths
+    {
+      form_url: form_fields_path(format: :js),
+      help_url: help_markup_path,
+      preview_url: preview_markup_path,
+      source_url: source_fields_path
+    }
```
brianp, line 8: What if instead of setting `allow_dropdown` on the form, and passing it through the request to the render we change the url of the request slightly to represent a new vs existing record in a restful-esque manner. Then in the controller based on whether we sent an id or now we set `@allow_dropdown` and render accordingly? /cc @etdsoft. It's pretty much the same concept we're doing now we're just not creating a specific data attribute for it.

```suggestion
  def editor_paths(resource)
    {
      form_url: form_fields_path(resource, format: :js),
      help_url: help_markup_path,
      preview_url: preview_markup_path,
      source_url: source_fields_path
    }
 end
```

aapomm: Unfortunately this doesn't work, specifically for the case when a user switches from Source -> Fields. We're defining the `form_url` to also have the `allow_dropdown` param so every request we make to render the Fields view will have that same param. This doesn't work for the Source -> Fields because we need to pass a `false` `allow_dropdown` here even if the resource is a new record.

app/controllers/fields_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  # Returns the source text given a form data
+  def source
+    render plain: convert_to_source
+  end
+
+  private
+
+  # Convert serialized form data to Dradis-style item content
+  def convert_to_source
```
brianp, line 21: As per 

> ... `HasFields#parse_fields`” and it seems this functionality is the opposite transformation (e.g. “give me source from fields”). Those two operations should probably be together.

Seems like the convert to source method should exist in has fields as a class method as well.

brianp: Maybe as `HasFields.parse_source`

Actually now it feels like the naming is backwards. The two method names should be opposite of each other.
`HasFields.parse_fields` isn't actually parsing fields it's parsing source into fields. So maybe the convert naming is better.

`HasFields.content_to_source` & `HasFields.content_to_fields` or
`HasFields.fields_to_source` & `HasFields.source_to_fields`.

app/services/field_parser.rb
etdsoft: fwiw, I was thinking of a singleton class with a class method. You wouldn't be able to mix it in from other classes, but you'd be able to call the singleton anyway.

PR #679, opened 2020-05-12
Issue states

aapomm:
### Spec
Implement CRUD for issue states

**Proposed solution**
1. Add state column to issues
2. Update issues CRUD to handle states

### How to test

**Test CRUD**

1. Create an issue and choose a state for it
2. Confirm that the issue was created with the state
3. Update the issue state
4. Confirm that the issue was updated with the state


app/assets/javascripts/tylium/pages/issues/state_menu.js
```diff
@@ -0,0 +1,34 @@
...
+
+    $('[data-state]', $scope).click(function () {
+      var selectedState = $(this).data('state');
+      var $stateField = $('[data-behavior~=issue-state]', $scope);
+
+      setBtnText.call($scope[0], textPrefix, selectedState);
+      $stateField.val(selectedState);
+    });
+  });
+});
```
sean-yeoh, line 34: @aapomm The JS seems a bit complicated for rendering 'Create'/'Update' text for the dropdown items. I've used the code below and move the changing text logic to the view.

```javascript
document.addEventListener('turbolinks:load', function(){
  function setBtnText(state) {
    var $submitBtn = $('[data-behavior~=state-submit-button]', $(this));

    $('[data-state]').removeClass('selected')

    var $selectedDropdownItem = $('[data-state~=' + state + ']', $(this))
    $selectedDropdownItem.addClass('selected');

    $submitBtn.prop('value', $selectedDropdownItem.find('[data-state-header]').text());
  }

  $('[data-behavior~=state-menu]').each( function() {
    var $scope = $(this);

    $('[data-state]', $scope).click(function () {
      var selectedState = $(this).data('state');
      var $stateField = $('[data-behavior~=issue-state]', $scope);

      setBtnText.call($scope[0], selectedState);
      $stateField.val(selectedState);
    });
  });
});
```

app/views/issues/form/_form_actions.html.erb
```diff
@@ -0,0 +1,25 @@
...
+      </div>
+      <div class="dropdown-item" data-state="published">
+        <span class="state-header" data-state-header="published"> Published Issue</span>
+        <span class="state-details">If this is the final form of the issue and it’s ready for the report.</span>
+      </div>
+    </div>
+  </div>
+  or
+  <%= link_to 'Cancel', @issue.new_record? ? project_issues_path(current_project) : [current_project, @issue], class: 'cancel-link' %>
+</div>
```
sean-yeoh, line 25: Since i changed the JS, i have to use the view logic below. For this to work, you'll have to add the below snippet to `Issue.rb`. We can either put the code here, or add it in a helper, or use a presenter. Whichever way works  too.

`app/models/issue.rb`
```ruby
  def self.state_descriptions
    {
      draft: 'Draft Issue',
      review: 'Ready for Review Issue',
      published: 'Published Issue'
    }
  end
```

`app/views/issues/form/_form_actions.html.erb`
We can even refactor the dropdown item logic further by including the long description in a helper and just loop over it.
```erb
<div class="form-actions">
  <div class="btn-group dropdown state-menu" data-behavior="state-menu">
    <%= f.button :submit, "#{@issue.new_record? ? 'Create' : 'Update'} #{Issue.state_descriptions[@issue.state.to_sym]}", class: 'btn btn-primary text-capitalize', data: { behavior: 'state-submit-button' } %>
    <%= f.hidden_field :state, data: { behavior: 'issue-state' } %>
    <button type="button" class="btn btn-primary dropdown-toggle dropdown-toggle-split ml-0" data-toggle="dropdown">
      <i class="fa fa-caret-down"></i>
    </button>
    <div class="dropdown-menu">
      <div class="dropdown-item <%= 'selected' if @issue.state == 'draft' %>" data-state="draft">
        <span class="state-header" data-state-header="draft"><%= @issue.new_record? ? 'Create' : 'Update' %> <%= Issue.state_descriptions[:draft] %></span>
        <span class="state-details">If you feel there is still more work before this issue is ready.</span>
      </div>
      <div class="dropdown-item <%= 'selected' if @issue.state == 'review' %>" data-state="review">
        <span class="state-header" data-state-header="review"><%= @issue.new_record? ? 'Create' : 'Update' %> <%= Issue.state_descriptions[:review] %></span>
        <span class="state-details">If you think this issue is ready and would like someone to have a look.</span>
      </div>
      <div class="dropdown-item <%= 'selected' if @issue.state == 'published' %>" data-state="published">
        <span class="state-header" data-state-header="published"><%= @issue.new_record? ? 'Create' : 'Update' %> <%= Issue.state_descriptions[:published] %></span>
        <span class="state-details">If this is the final form of the issue and it’s ready for the report.</span>
      </div>
    </div>
  </div>
  or
  <%= link_to 'Cancel', @issue.new_record? ? project_issues_path(current_project) : [current_project, @issue], class: 'cancel-link' %>
</div>
```

app/views/issues/new.html.erb
```diff
@@ -17,7 +17,7 @@
   <div id="issues_editor">
     <div class="note-text-inner">
       <h4 class="header-underline">Add new issue</h4>
-      <%= render 'form' %>
+      <%= render 'issues/form/form' %>
```
sean-yeoh, line 20: Just wondering is there a reason for moving `issues/_form` to `issues/form/_form`? `issues/form` seems more intuitive to me.

app/views/issues/form/_form_actions.html.erb
aapomm: Thanks @sean-yeoh. There already is a method for that: `Issue.state_names`. I've updated the JS and this file to use that method and simplify.

app/views/issues/new.html.erb
aapomm: It's not aesthetic but it groups the form-building partials into one director, which was originally suggested by Daniel.

PR #680, opened 2020-05-12
Implement multi edit of issue states in issues#index

aapomm:
### Spec

With the implementation of issue states, we need to add a way to edit multiple issue states, otherwise users will have to open and edit each issue to update their states.

**Proposed solution**

Add a “State” functionality in the issues table toolbar

### How to test

1. Add multiple issues in a project
2. Visit the Issues Summary page
3. Select multiple issues in the table and select a State.
4. Confirm that the issues have their state changed.

### DIFF
--- a/app/controllers/issues/states_controller.rb
+++ b/app/controllers/issues/states_controller.rb
@@ -1,0 +1,24 @@
+class Issues::StatesController < AuthenticatedController
+  include ProjectScoped
+
+  def update_states
+    issues = current_project.issues.where(id: params[:ids])
+
+    if issues.update_all(state: state_params, updated_at: Time.now)
+      render json: { state: Issue.state_names[state_params] }
+    else
+      render json: { message: 'Something went wrong.' }
+    end
+  end
+
+  private
+
+  def state_params
+    @state_params ||=
+      if Issue.states.keys.include?(params[:state])
+        params[:state]
+      else
+        :published
+      end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -49,6 +49,9 @@ Rails.application.routes.draw do
       collection do
         post :import
         resources :merge, only: [:new, :create], controller: 'issues/merge'
+        resources :states, only: [], controller: 'issues/states' do
+          collection { put :update_states }
+        end
       end
 
       resources :nodes, only: [:show], controller: 'issues/nodes'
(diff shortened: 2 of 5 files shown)


app/controllers/issues/states_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  private
+
+  def state_params
+    @state_params ||=
+      if Issue.states.keys.include?(params[:state])
+        params[:state]
+      else
+        :published
```
sean-yeoh, line 21: Nothing to be changed here, just curious if there's any reason for the default state to be published

config/routes.rb
```diff
@@ -49,6 +49,9 @@
       collection do
         post :import
         resources :merge, only: [:new, :create], controller: 'issues/merge'
+        resources :states, only: [], controller: 'issues/states' do
+          collection { put :update_states }
+        end
```
sean-yeoh, line 54: I think the controller name can be more descriptive, since we're just using it to update states. Maybe:

```suggestion
        resource :bulk_update_states, only: [:update]
```
Then we can PUT to `/projects/:project_id/issues/bulk_update_states`.

What do you think?

app/controllers/issues/states_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  include ProjectScoped
+
+  def update_states
+    issues = current_project.issues.where(id: params[:ids])
+
+    if issues.update_all(state: state_params, updated_at: Time.now)
+      render json: { state: Issue.state_names[state_params] }
+    else
+      render json: { message: 'Something went wrong.' }
+    end
```
sean-yeoh, line 11: Will the return of `issues.update_all` ever be false or nil?
If yes i think we should also handle on:error in the ajax call, if not we can just  do
```
issues.update_all(state: state_params, updated_at: Time.now)
render json: { state: Issue.state_names[state_params] }
```

app/controllers/issues/states_controller.rb
MattBudz: @sean-yeoh The reason for this is to eliminate the impact on users who do not want to use Dradis for QA. Since issues are only visible in reports when they are published, users who do not want to use QA would have to always set each issue to published in order to see it in their reports if the default was anything other than published.

sean-yeoh: @MattBudz Got it, thanks for the explanation!

PR #681, opened 2020-05-12
Implement uploading of issue states from plugins

aapomm:
### Spec

With the implementation of issue states, we need to implement a way for users to automatically set issue states when uploading from plugins. Otherwise, they will have to manually set each issue to their desired state.

**Proposed solution**

Implement a dropdown in the uploads view that sets the incoming findings’ state.

### How to test

Before testing, pull all the necessary branches of CE and the plugins and update the Gemfile to point to the plugins local directory:

gem 'dradis-plugins', path: '<local_directory>'
gem 'dradis-projects', path: '<local_directory>'

Then run bundle install to install the local plugins.

1. Go to the uploads page
2. Select a plugin (nessus, netsparker, etc.)
3. Select a state
4. Upload a sample plugin file
5. Confirm that the created issues have the selected state


brianp:

PR #687, opened 2020-05-13
Local Caching Base

brianp:
### Summary
This PR improves the current implementation of local cache for textareas. The new implementation removes the prompt when the user navigates back to the form after editing, and also cache the whole form instead of only textareas. This only applies when creating new Card, Evidence, Issue and Note. 

## **Card**
**Scenario 1**
1. Create new **blank** card
2. Enter name, select a due date and enter something in description, assign some one
3. Navigate to another board
4. Create new **blank** Card on the new board
5. Ensure the fields entered previously is not on the card form on the new board
6. Go back to the first board and list and create **blank** card
7. Ensure the fields entered previously appears

**Scenario 2**
1. Create new **blank** card
2. Enter name, select a due date and enter something in description, assign some one
3. Click Create Card
4. Create new **blank** card
5. Ensure the fields entered previously is not on the form anymore

**Scenario 3**
1. Create new **blank** card
2. Enter name, select a due date and enter something in description, assign some one
3. Click 'Cancel'
4. Create new **blank** card
5. Ensure the fields entered previously is not on the form anymore

## **Issue**
**Scenario 1**
1. Create new **blank** issue
2. Enter description, select a tag
3. Navigate away
4. Create new **blank** issue again
5. Ensure the description and the selected tag appears 

**Scenario 2**
1. Create new **blank** issue
2. Enter description, select a tag
3. Click Create Issue
4. Create new **blank** issue
5. Ensure the fields entered previously is not on the form anymore

**Scenario 3**
1. Create new **blank** issue
2. Enter description, select a tag
3. Click 'Cancel'
4. Create new **blank** issue
5. Ensure the fields entered previously is not on the form anymore

## **Note**
**Scenario 1**
1. Create new **blank** note
2. Enter something in textarea, select a category
3. Navigate away
4. Create new **blank** note on the same node
5. Ensure the textarea and the selected category appears 

**Scenario 2**
1. Create new **blank** note
2. Enter something in textarea, select a category
3. Click Create Note
4. Create new **blank** note
5. Ensure the fields entered previously is not on the form anymore

**Scenario 3**
1. Create new **blank** note
2. Enter something in textarea, select a category
3. Click 'Cancel'
4. Create new **blank** note
5. Ensure the fields entered previously is not on the form anymore

## **Evidence**
**Scenario 1**
1. Create new **blank** evidence
2. Enter something in textarea, select an issue
3. Navigate away
4. Create new **blank** note on the same node
5. Ensure the textarea and the selected issue appears 

**Scenario 2**
1. Create new **blank** evidence
2. Enter something in textarea, select an issue
3. Click Create Evidence
4. Create new **blank** evidence
5. Ensure the fields entered previously is not on the form anymore

**Scenario 3**
1. Create new **blank** evidence
2. Enter something in textarea, select an issue
3. Click 'Cancel'
4. Create new **blank** evidence
5. Ensure the fields entered previously is not on the form anymore

### Check List

- [x] Added a CHANGELOG entry


app/assets/javascripts/tylium/modules/auto_save/local.js
```diff
@@ -0,0 +1,127 @@
...
+    var clearCacheElement = this.target.querySelector('[data-behavior~=clear-local-auto-save]');
+
+    if (clearCacheElement) {
+      clearCacheElement.addEventListener('click', function(event) {
+        that.cancelled = true;
+        localStorage.removeItem(that.key);
+      })
+    }
+
+    this._debounceTimer = 500;
```
brianp, line 37: @sean-yeoh This should get moved up top with the other constructor values.

app/assets/javascripts/tylium/modules/auto_save/local.js
```diff
@@ -0,0 +1,127 @@
...
+      // Don't add event handler for submit button
+      if (input.type === 'button') { return true; }
+
+      // we're using a jQuery plugin for :textchange event, so need to use $()
+      $(input).on('textchange change', that.handleTextChange.bind(that));
+    })
+  }
+
+  handleTextChange() {
+    clearTimeout(this._debounceTimer);
```
brianp, line 49: I've double used a variable in this example and shouldn't have. 

It should read:

```javascript
 handleTextChange() {
    clearTimeout(this._debounceTimeout);

    this._debounceTimeout = setTimeout(function() {
      this.setData();
    }.bind(this), this._debounceTimer);
  }
```

`this._debounceTimer` should represent the timing interval and `this._debounceTimeout` should be the timeout. One was clobbering the other.

app/assets/javascripts/tylium/modules/auto_save/local.js
```diff
@@ -0,0 +1,126 @@
...
+    })
+
+    if (tagInputData) {
+      var hiddenInput = document.querySelector('#issue_tag_list');
+      hiddenInput.value = tagInputData[1];
+
+      var $target = $(`.js-taglink[data-tag='${tagInputData[1]}']`);
+      new SelectTagDropdown($target)
+    }
+  }
```
etdsoft, line 116: There's still something here that I feel could be improved.

We don't have to, but me may _want_ to.

First L105-L107 are repeating a loop that we already did in L72. Can't we just add the `if issue[tag_list]' conditional there, and only then make the call to `SelectTagDrop`?

Second, if we're adding a separate class whose responsibility is to "select the tag", then `LocalAutoSave` has no business touching the `#issue_tag_list`, that responsibility is a lot closer to the component that "selects the tag" than to the auto-save component. Auto-save cares about persisting the data, but if we have another object whose responsibility is to manage the state of the Tag select control, let it do it.

Particularly L111 is also in `SelectTagDropdown:14`, which reinforces my thinking that we're splitting/duplicating reponsibilities.

Right now the `SelectTagDropdown` class only does some cosmetics (CSS, label, etc.) I'd say we move L14 up to the `SelectTagDropdown` class, removed it from `LocalAutoSave` and let `SelectTagDropdown` handle the hidden input value and the CSS of the select control.

sean-yeoh: Thanks @etdsoft , I've refactored it to be inside the restoreData() loop and it's much cleaner now!
```javascript
if (element.name === 'issue[tag_list]') {
  var $tagDropdownItem = $(`.js-taglink[data-tag='${item[1]}']`);
  new SelectTagDropdown($tagDropdownItem)
}
```
Initially, I tried to trigger a onclick event programmatically on the $tagDropdownItem but it doesn't seem to work on a hidden element, so i went with the above instead

PR #692, opened 2020-05-14
Auto Save Whodunnit

brianp:
### Summary

At the point we use a background job we no longer have access to PaperTrail.request.whodunnit because we're not operating on a different thread. To correctly apply who did it, we need to pass in the param when the job is serialized.

### How to test

- Log in with two tabs, and two different users
- Edit a resource with user A
- Trigger auto save by waiting 3 minutes, or navigating to a different page
- With User B validate the the auto-save occured
- With User B edit the resource, make no changes and press "Update"
- Verify on the revision history user B is credited with the update

### Check List


brianp:
I'll bite on the named parameter.
I'll validate the whodunnit. I'm 99% sure it's a string of the email, not a user_id, nor a user. If it's the email string I'll leave it as it is. If it's an entire user and the record is doing something magical to pull the email string, I'll make the suggested change. Alternatively I may change it to `user_email` but passing the id and serializing or fetching a user etc seems like un-needed work.

PR #700, opened 2020-05-19
update return_to before redirect to login

MattBudz:
### Spec
Currently, when a user is on any view and the session times out, the user is redirected back to the login view. Once logged back in the user goes to projects#show in ce or projects#index in pro.

Besides that, when they are using the editor while timed out, the preview on the right does not reflect their changes. (because they have been logged out). This isn't good UX and would confuse users.

#### Proposed solution
Return the user back to the page they were on when the session timed out.
While the user is on the editor while timed out, show an alert message to inform users to login again

### Check List

- [x] Added a CHANGELOG entry


spec/features/sessions_spec.rb
```diff
@@ -55,6 +55,38 @@
...
+        fill_in 'login', with: @user.email
+        fill_in 'password', with: password
+        click_button 'Let me in!'
+      end
+
+      it 'redirects to previous page' do
+        login
+
+        Timecop.freeze(Time.now + 1.hour) do
+          visit new_project_issue_path(Project.find(1))
```
aapomm, line 70: `Project.find(1)` will have to change in Pro, so I suggest we store this in a variable like:

```
let (:project) { Project.find(1) }
```

So in Pro, all we have to change is that line.

spec/features/sessions_spec.rb
```diff
@@ -56,6 +56,40 @@ def login
...
+        fill_in 'password', with: password
+        click_button 'Log in'
+      end
+
+      before do
+        login
+      end
+
+      it 'redirects to previous page' do
+        Timecop.freeze(Time.now + 1.hour) do
```
aapomm, line 74: Is there an alternative to using `Timecop.freeze`? Introducing more `#freeze` calls might cause issues in CI.

app/controllers/sessions_controller.rb
```diff
@@ -15,11 +15,26 @@ def destroy
...
       end
       format.json { head :not_found }
       format.js { head :not_found }
     end
   end
+
+  private
+
+  def return_to
+    if request.get?
```
aapomm, line 34: Instead of adding another brakeman ignore, would it help if we follow what's recommended here: https://brakemanscanner.org/docs/warning_types/http_verb_confusion/ and add an explicit `request.post?` check?

spec/rails_helper.rb
```diff
@@ -73,6 +73,8 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 
+  config.include ActiveSupport::Testing::TimeHelpers
```
sean-yeoh, line 76: To use https://api.rubyonrails.org/v5.2.4/classes/ActiveSupport/Testing/TimeHelpers.html instead of Timecop helpers: https://github.com/dradis/dradis-ce/pull/700#discussion_r972857626

app/controllers/sessions_controller.rb
sean-yeoh: Gotcha!

app/controllers/sessions_controller.rb
```diff
@@ -15,11 +15,26 @@ def destroy
...
+      warden_options[:attempted_path]
+    elsif request.post?
```
sean-yeoh, line 36: See: https://brakemanscanner.org/docs/warning_types/http_verb_confusion/

app/controllers/sessions_controller.rb
MattBudz: The warning comes up because our if block is checking `request.get?` and then we have an `else`. Brakeman is assuming the else is for `request.post?` but that's not the case in our code. This is just a false positive. If we don't want to add another ignore, then we can add an `elsif` and be explicit about what we are checking.

This fixes the warning:

```ruby
  def return_to
    if request.get?
      warden_options[:attempted_path]
    elsif request.env['HTTP_REFERER'].present?
      request.env['HTTP_REFERER']
    end
  end
```

PR #705, opened 2020-05-25
Shared assets

MattBudz:
### Summary

This PR shares select tylium assets with mintcreek and provides bs3 support.

### Check List

### DIFF
--- a/app/assets/javascripts/tylium.js
+++ b/app/assets/javascripts/tylium.js
@@ -20,7 +20,6 @@
 
 //= require cable
 
-//= require shared/behaviors
 //= require shared/charts
 //= require shared/console_updater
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/tylium.js
```diff
@@ -30,6 +30,8 @@
 //= require tylium/plugins/jquery.treenav
 
 //= require tylium/behaviors
+//= require shared/behaviors
```
etdsoft, line 33: any reason not to include this in the `shared/` block (around L26)?

MattBudz: If EditorToolbar init's before textile it breaks.

PR #713, opened 2020-06-03
Add textile comments filter and comments formatter

aapomm:
### Spec
Mentions in comments are borked when there is a period `.` present in a users email. This is because the textile filter evaluates `@not.code@asdf` as an inline code *@not.code@asdf*.

**Proposed Solution**
Make inline code no-op for textile in comments for now.

### How to test:
1. Add a user with email: `test.user@gmail.com`
2. Mention that user
3. Confirm that the user's avatar correctly displays


lib/html/comments_formatter.rb
```diff
@@ -0,0 +1,9 @@
+module HTML
+  module CommentsFormatter
```
etdsoft, line 2: This should Reference Textile in the name

lib/html/pipeline/dradis_textile_comments_filter.rb
```diff
@@ -0,0 +1,9 @@
+module HTML
+  class Pipeline
+    class DradisTextileCommentsFilter < TextFilter
+      def call
+        "<div>#{RedCloth.new(@text, [:filter_html, :no_span_caps]).to(HTML::CommentsTextileFormatter)}</div>"
```
brianp, line 5: `to(HTML::CommentsTextileFormatter)`

Can you detail why we need this class to perform this action? Wouldn't `HTML::Pipeline::TextileFilter` be able to perform this. I'm not seeing why we needed two different classes for this workaround.

aapomm: The [HTML::Pipeline::TextileFilter](https://github.com/jch/html-pipeline/blob/628d4c00dcc4ca7ba6d9b11a4244d255050169f9/lib/html/pipeline/textile_filter.rb) is not enough to do what we want (add `<div>`s, redcloth options, etc.)

If you mean the `DradisTextileFilter` class, we still need to have a filter that does not remove the `inline code` functionality to make it work in situations outside of comments.

brianp: Couldn't this be done by passing an argument to `DradisTextileFilter`? `:no_inline_code` that will perform the change needed instead of the required logic to properly wrap textile ending up in two places?

brianp:
Once green. :shipit:

PR #716, opened 2020-06-08
redirect to the correct resource on cancel

MattBudz:
### Summary

Currently, when a user edits evidence from an issue, clicking `update `will return to the issue but clicking `cancel` brings the user to the node.

This PR updates the cancel link to bring the user back to the issue if the user started editing the evidence from an issue.

### Check List

- [x] Added a CHANGELOG entry


app/controllers/evidence_controller.rb
```diff
@@ -10,6 +10,7 @@ class EvidenceController < NestedNodeResourceController
   before_action :initialize_nodes_sidebar, only: [ :edit, :new, :show ]
   skip_before_action :find_or_initialize_node, only: [:create_multiple]
   before_action :set_auto_save_key, only: [:new, :create, :edit, :update]
+  before_action :cancel_path, only: [:new, :create, :edit, :update]
```
brianp, line 13: Sorry dudes I'm going to disapprove. We're adding an entire controller method, with a before_action- to generate a _link_.

At most this is essentially a URL helper, which may qualify is a view helper, but does not require bulking out controller functionality.

app/helpers/evidence_helper.rb
```diff
@@ -0,0 +1,11 @@
...
+  def cancel_path(back_to)
+    if back_to == 'issue'
+      [current_project, @evidence.issue]
+    elsif @evidence.persisted?
+      [current_project, @node, @evidence]
+    else
+      [current_project, @node]
+    end
+  end
+end
```
MattBudz, line 11: @sean-yeoh is there a way to further improve this so we are not partly duplicating some of this code in the evidence controllers update action?

app/controllers/evidence_controller.rb
```diff
@@ -5,6 +5,7 @@ class EvidenceController < NestedNodeResourceController
   include MultipleDestroy
   include NodesSidebar
   include NotificationsReader
+  include EvidenceHelper
```
brianp, line 8: I have a feeling eventually we'll end up with our own routes helper file. We do enough switches to generate paths it'll be helpful. I'm good with this as is for now.

PR #721, opened 2020-06-24
Sidebar v2

MattBudz:
### Summary

This PR brings collapsed sidebar improvements:
- Labels below the icons
- Nodes can now be added with the sidebar collapsed
- Animations and transitions have been removed.

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,5 @@
 v3.#.# (MMM’YY)
-  * [feature #1]
-  * [feature #2]
-  * [feature ...]
+  * Update main sidebar
   * Upgraded gems: rack, sanitize
   * Bugs fixed:
     - Comments:
--- a/app/assets/javascripts/tylium/modules/sidebar.js
+++ b/app/assets/javascripts/tylium/modules/sidebar.js
@@ -6,50 +6,55 @@
     this.$viewContent = $sidebar.siblings('[data-behavior~=view-content]');
     this.storageKey = $sidebar.data('storage-key');
 
+    // Must match $sidebarCollapsed and $sidebarExpanded values from tyluim/variables.scss
+    this.collapsedWidth = '6rem';
+    this.expandedWidth =  '15.5rem';
+
     this.init();
   }
 
   Sidebar.prototype = {
     init: function() {
-      this.toggle(this.isSidebarOpen(), 'no-animation');
+      this.toggle(this.isSidebarOpen());
 
       var that = this;
-      $('[data-behavior~=sidebar-toggle]').on('click', function() {
-        if (that.isSidebarOpen() && $(this).is('[data-behavior~=open-only]')) return;
 
-        that.toggle(!that.isSidebarOpen(), 'animate');
+      $('[data-behavior~=sidebar-toggle]').on('click', function(e) {
+        if ((that.isSidebarOpen() && $(this).is('[data-behavior~=open-only]')) || ($(e.target).data('behavior') == 'add-node')) return;
+
+        that.toggle(!that.isSidebarOpen());
       });
     },
     changeState: function(state) {
       localStorage.setItem(this.storageKey, state);
       Turbolinks.clearCache();
     },
-    close: function(animationClass) {
-      this.$navbar.css('left', '0px');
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/tylium/modules/sidebar.js
```diff
@@ -6,50 +6,55 @@
     this.$viewContent = $sidebar.siblings('[data-behavior~=view-content]');
     this.storageKey = $sidebar.data('storage-key');
 
+    // Must match $sidebarCollapsed and $sidebarExpanded values from tyluim/variables.scss
```
aapomm, line 9: Is there a way we can calculate this from the view/CSS so we don't have to set this in multiple places?

app/assets/javascripts/tylium/modules/sidebar.js
```diff
@@ -6,50 +6,55 @@
...
-      this.toggle(this.isSidebarOpen(), 'no-animation');
+      this.toggle(this.isSidebarOpen());
 
       var that = this;
-      $('[data-behavior~=sidebar-toggle]').on('click', function() {
-        if (that.isSidebarOpen() && $(this).is('[data-behavior~=open-only]')) return;
 
-        that.toggle(!that.isSidebarOpen(), 'animate');
+      $('[data-behavior~=sidebar-toggle]').on('click', function(e) {
+        if ((that.isSidebarOpen() && $(this).is('[data-behavior~=open-only]')) || ($(e.target).data('behavior') == 'add-node')) return;
```
aapomm, line 23: A couple of things on this line:

- Inconsistency between `[data-behavior~=` and `data('behavior')`
- Is `$(e.target)` the same as `$(this)` in this click event?

We can invert this conditional so that we don't have to return:
```
function(e) {
  if (!condition){
    that.toggle(!that.isSidebarOpen());
  }
})...
```

app/assets/javascripts/tylium/modules/sidebar.js
```diff
@@ -6,50 +6,55 @@
...
 
       this.changeState(false);
     },
     isSidebarOpen: function() {
       return JSON.parse(localStorage.getItem(this.storageKey))
     },
-    open: function(animationClass) {
-      this.$navbar.css('left', '207px');
+    open: function() {
+      this.$navbar.css('left', '9.5rem');
```
aapomm, line 45: Can we name this magic number? Example:
```
this.navbarOffset = '9.5rem';
...
this.$navbar.css('left', this.navbarOffset);
```

CHANGELOG
```diff
@@ -1,7 +1,5 @@
 v3.#.# (MMM’YY)
-  * [feature #1]
-  * [feature #2]
-  * [feature ...]
+  * Update main sidebar
```
aapomm, line 2: Since the changelog is now front-facing/customer-facing, we'll benefit more from being more descriptive here. What exactly is updated in the sidebar?

aapomm, line 2: Since the changelog is now front-facing/customer-facing, we'll benefit more from being more descriptive here. What exactly is updated in the sidebar?

app/assets/javascripts/tylium/modules/sidebar.js
MattBudz: Yep, made this change. Thanks.

app/assets/javascripts/tylium/modules/sidebar.js
MattBudz: I fixed the inconsistency and inverted the conditional. 

`$(e.target)` is a child of `$(this)`, we need to check for it to prevent event bubbling. Otherwise, the collapsed sidebar will expand when a user wants to add a node.

app/assets/javascripts/tylium/modules/sidebar.js
MattBudz: This number is the result of subtracting the collapsed width from the expanded width. 
I went ahead and changed this to use a calculated value as this would need to be adjusted if the widths are changed in the future. This change now eliminates the need to adjust this value manually.

PR #729, opened 2020-07-29
Absolute email notification timestamp

brianp:
### Summary

When email notifications get sent out each comment has a timestamp written in a relative format "8 minutes ago" etc. This relative timestamp is accurate the moment the email is sent, but if the email isn't read for 15 hours then the real time the comment was made was 15 hours and 8 minutes ago, not 8 minutes ago. This can be confusing when reading the email and attempting to build a mental timeline of when events occur.

### Solution

Change relative time to actual time (ie. Jul 9th at 2:45pm)

### Copyright assignment

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/presenters/digest_presenter.rb
+++ b/app/presenters/digest_presenter.rb
@@ -2,7 +2,7 @@ class DigestPresenter < NotificationPresenter
   attr_reader :current_project, :notifications, :template
 
   # Jul 9th at 2:45pm
-  Time::DATE_FORMATS[:digest_format] = ->(time) { time.strftime "%h #{time.day.ordinalize} at %l:%M%P" }
+  Time::DATE_FORMATS[:digest_format] = ->(time) { time.strftime "%h #{time.day.ordinalize} at %l:%M%P #{time.zone}" }
 
   def initialize(notifications, project, template)
     @current_project = project
@@ -23,7 +23,7 @@ class DigestPresenter < NotificationPresenter
   end
 
   def created_at
-    notification.created_at.to_s(:digest_format)
+    notification.created_at.localtime.to_s(:digest_format)
   end
 
   def text_title
--- a/app/presenters/digest_presenter.rb
+++ b/app/presenters/digest_presenter.rb
@@ -1,9 +1,6 @@
 class DigestPresenter < NotificationPresenter
   attr_reader :current_project, :notifications, :template
 
-  # Jul 9th at 2:45pm
-  Time::DATE_FORMATS[:digest_format] = ->(time) { time.strftime "%h #{time.day.ordinalize} at %l:%M%P #{time.zone}" }
-
   def initialize(notifications, project, template)
     @current_project = project
     @notifications = notifications
@@ -23,7 +20,8 @@ class DigestPresenter < NotificationPresenter
   end
 
   def created_at
-    notification.created_at.localtime.to_s(:digest_format)
+    time = notification.created_at.localtime
+    time.strftime("%h #{time.day.ordinalize} at %l:%M%P #{time.zone}") # Jul 9th at 2:45pm CEST
   end
 
   def text_title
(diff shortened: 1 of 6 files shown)


app/presenters/digest_presenter.rb
```diff
@@ -1,6 +1,9 @@
 class DigestPresenter < NotificationPresenter
   attr_reader :current_project, :notifications, :template
 
+  # Jul 9th at 2:45pm
+  Time::DATE_FORMATS[:digest_format] = ->(time) { time.strftime "%h #{time.day.ordinalize} at %l:%M%P" }
```
sean-yeoh, line 5: @brianp Is it possible to add timezone? I'm not too sure how to use the user's timezone, but maybe we can use UTC as default?

brianp: It's a good idea, but we don't really do that anywhere else. We use the local_time gem which does some transformations on the front-end for timestamps. I think for now it's safe for everyone to assume it's the local system time of the VM without distinctly clarifying. Once we begin to offer a timezone setting for notifications etc it'll matter a little more.

app/presenters/digest_presenter.rb
```diff
@@ -1,6 +1,9 @@
 class DigestPresenter < NotificationPresenter
   attr_reader :current_project, :notifications, :template
 
+  # Jul 9th at 2:45pm
+  Time::DATE_FORMATS[:digest_format] = ->(time) { time.strftime "%h #{time.day.ordinalize} at %l:%M%P #{time.zone}" }
```
etdsoft, line 5: Why is it better to add a global `DATE_FORMAT` here than to just do an `strftime` call in L26?

app/views/notification_mailer/digest.html.erb
```diff
@@ -65,17 +65,21 @@
...
               <table cellspacing="0" cellpadding="0">
                 <tr>
+                  <td width="20px">&nbsp;</td>
                   <td width="590px">
                     <hr style="border-top: 0.5px solid #ddd; border-bottom: 0.5px solid #ddd; border-right: 0.5px solid transparent; border-left: 0.5px solid transparent;">
                   </td>
+                  <td width="20px">&nbsp;</td>
                 </tr>
                 <tr>
                   <td height="10px"></td>
```
etdsoft, line 75: no `&nbsp;`? I'm assuming this is email client related, but it caught my eye it's missing here and also in L85, is it because we're using `height` and not `width` so it isn't needed?

brianp: Sounds like a @MattBudz question. If you can throw in :D

app/presenters/digest_presenter.rb
brianp: `strftime` is the way it's been done through the application
`DATE_FORMATS` is a way we can enhance how it's being done. I like the idea of using the global formatter and initializing the formats we want in a single place. Helps spots consistency problems, makes things easier for re-use. Can be a smart default.

app/views/notification_mailer/digest.html.erb
MattBudz: >is it because we're using height and not width so it isn't needed?

That's right. These cells are used to create vertical white space so we don't need `&nbsp;`.

`&nbsp;` is used to show horizontal white space correctly in FF.

app/presenters/digest_presenter.rb
etdsoft: I don't disagree, but if we're deciding to introduce a "consistent format across the app", a) this doesn't feel like the right PR for that; when we open the PR for that b) we should sweep the code to make sure we're adding the consistent formatting across the board, and not in this one place...

brianp: I'll change it back.

My thinking is the debate between "Making small improvements which break consistency, or continuing doing things the same way simply because it's the way we have been doing it".
I like the idea of grand sweeping changes but they don't always end up happening.

PR #730, opened 2020-07-29
Editor Screenshots

brianp:
### Summary

Improve screenshot workflow in the editor. Support drag & drop, copy & paste, and a new editor toolbar button for uploading images.

### Copyright assignment

### Check List

- [x] Added a CHANGELOG entry


app/assets/javascripts/tylium/modules/fileupload.js
```diff
@@ -1,33 +1,64 @@
...
+  $('[data-behavior~=jquery-upload]').fileupload({
+     autoUpload: true,
+     dropZone: function(){ return $('[data-behavior~=drop-zone]') },
+     pasteZone: function(){ return $('[data-behavior~=drop-zone]') },
+     singleFileUploads: true,
+     destroy: function(e, data) {
+       if (confirm('Are you sure?\n\nProceeding will delete this attachment from the associated node.')) {
+         $.blueimp.fileupload.prototype.options.destroy.call(this, e, data);
+       }
+     }
```
MattBudz, line 13: I don't think we need this anymore since we are auto-uploading. This was for the staged area delete button in case users wanted to remove files from the queue.

app/assets/javascripts/tylium/modules/fileupload.js
```diff
@@ -1,33 +1,64 @@
...
+  }).on('fileuploadadd', function (e, data) {
+    if (e.originalEvent !== undefined) {
+      var $textarea = $(e.originalEvent.target)
+
+      if ($textarea.is($('[data-behavior~=rich-toolbar]'))) {
+        var editorToolbar = $textarea.data('editorToolbar'),
+            that = this;
```
MattBudz, line 20: we don't use `that` anywhere

app/assets/javascripts/shared/editor_toolbar.js
```diff
@@ -64,13 +75,27 @@ class EditorToolbar {
...
+        $textarea: that.$editorField.find('textarea, input[type=text]')
+      });
+
+      // Clear the $fileField so it never submit unexpected filedata
+      $(this).val('');
+    });
+
     // when a toolbar button is clicked
-    this.$editorToolbar.find('[data-btn]').click(function () {
+    this.$editorToolbar.find('[data-btn]').click(function (e) {
```
MattBudz, line 89: I don't think the `e` parameter here

app/assets/javascripts/tylium/modules/fileupload.js
brianp: > Proceeding will delete this attachment from the associated node.

Isn't this the delete button to remove the file completely?

brianp: Verified, this is for the button to delete an existing already uploaded attachment, and should not be removed.

brianp: This did remind me though. We hadn't removed the Add, Start, and Remove buttons for the staging area. So I've removed those now as well.

app/views/issues/_form.html.erb
```diff
@@ -15,9 +15,11 @@
         class: 'textile h-100',
         data: {
           allow_dropdown: @issue.new_record?,
-          behavior: 'rich-toolbar',
+          autosave_key: "#{current_project.id}-issue-#{ @issue.persisted? ? "edit-#{@issue.id}" : 'new' }",
```
sean-yeoh, line 18: @brianp What's this `autosave_key` here for?

app/assets/javascripts/shared/editor_toolbar.js
```diff
@@ -15,18 +15,28 @@ class EditorToolbar {
...
     this.init();
   }
 
   init() {
     this.$target.wrap('<div class="editor-field" data-behavior="editor-field"><div class="textarea-container"></div></div>');
-    this.$editorField = this.$target.parents('[data-behavior=editor-field]');
-    this.$editorField.prepend('<div class="editor-toolbar" data-behavior="editor-toolbar"></div>');
-    this.$editorToolbar = this.$editorField.find('[data-behavior=editor-toolbar]');
+    this.$editorField = this.$target.parents('[data-behavior~=editor-field]');
+    this.$editorField.prepend('<div class="editor-toolbar" data-behavior="editor-toolbar" data-edit="#'+ this.$target.data('id') + '"></div>');
```
sean-yeoh, line 32: Found an interesting part where the first editor toolbar will have a `data-edit=#undefined"`. But it doesn't seem to break anything

image

app/assets/javascripts/shared/editor_toolbar.js
```diff
@@ -15,18 +15,28 @@ class EditorToolbar {
...
+    this.$editorToolbar = this.$editorField.find('[data-behavior~=editor-toolbar]');
 
     this.$editorToolbar.append(this.textareaElements(this.opts.include));
+    this.$fileField = $('<input type="file" name="editor-toolbar-' + Math.random().toString(36) + '[]" multiple style="display: none">');
```
sean-yeoh, line 36: What does the `name="editor-toolbar-' + Math.random().toString(36) + '[]` name do? Don't see it used anywhere

app/views/issues/_form.html.erb
brianp: Not sure exactly. Merge conflict? Issues should have one though yeah?

app/assets/javascripts/shared/editor_toolbar.js
```diff
@@ -15,18 +15,28 @@ class EditorToolbar {
...
+    this.$editorToolbar.append(this.$fileField);
```
sean-yeoh, line 37: And also, possible to add a check to append the fileField only if there's an `image` listed in rich-toolbar? Now it's always prepending the fileField even if the editor doesn't have an image uploader.

app/assets/javascripts/shared/editor_toolbar.js
brianp: It's designed to assign random names to the input field, ensuring no conflict as an input would exist for every field on the page.

app/assets/javascripts/shared/editor_toolbar.js
brianp: Good suggestion.

app/views/attachments/_attachment_box.html.erb
```diff
@@ -62,30 +62,11 @@
...
-        <button type="submit" class="btn btn-success start" data-toggle="tooltip" data-placement="bottom" title="Start">
-          <i class="fa fa-upload fa-inverse"></i>
-          <span></span>
-        </button>
-        <button type="reset" class="btn btn-warning cancel" data-toggle="tooltip" data-placement="bottom" title="Cancel">
-          <i class="fa fa-ban fa-inverse"></i>
-          <span></span>
-        </button>
-
-      </div>
```
sean-yeoh, line 88: Why was the upload button removed?

MattBudz: We are auto-uploading files. As soon as a user selects a file via upload button, drops onto a drop zone or pastes into a text area, the file(s) are uploaded right away rather than being staged like they were previously. This is to facilitate a more seamless user experience when users drag/drop or paste into text areas (we want the files to be available right away not staged in the sidebar).

The buttons that should have been removed were just the `start` and `cancel` buttons, along with the buttons that should up in the staged queue. The `add` button should remain to give users the option to upload by browsing their file system.

MattBudz: I re-added the `add` button in the sidebar and removed the `start` button from the staged queue

app/assets/javascripts/shared/editor_toolbar.js
brianp: @MattBudz do we need this?

app/assets/javascripts/shared/editor_toolbar.js
brianp: > And also, possible to add a check to append the fileField only if there's an image listed in rich-toolbar? Now it's always prepending the fileField even if the editor doesn't have an image uploader.

Good idea!

>  if we're using the dropzone on the sidebar to upload, why is this file input field necessary? Is it possible to trigger the dropzone programmatically?

So we actually are sending the files programatically to the dropzone. The difference is about events. If we just trigger the file upload on the dropzone and get the upload select box there's no way to identify what toolbar it came from.
Using an input box associated to the toolbar, and then sending the file to the dropzone lets us middle-man the process and inject the data regarding *where* the upload needs to call back and inject syntax into.

app/views/issues/_form.html.erb
brianp: Confirmed. It's a merge conflict and not-needed.

app/assets/javascripts/shared/editor_toolbar.js
brianp: Added `accept="image/*"` on the toolbar button. Technically the attachment has always supported any attachment kind so I've not enforced it any further. You could drag and drop another file type, but the `Image` button on the toolbar only supports images.

app/assets/javascripts/shared/editor_toolbar.js
brianp: Changed to only append the uploaded if the image button is used.

app/assets/javascripts/shared/editor_toolbar.js
MattBudz: removed

PR #732, opened 2020-08-06
Node properties longtext

aapomm:
### Spec
Currently, the node properties column in the database is of type TEXT. This can only support 65k+ bytes. This is a problem especially since importing Nessus/Nmap files can easily exceed the limit.

**Proposed solution**
Update the node properties column to LONGTEXT
Alternatively, we can truncate the extra data instead when uploading.

### How to test

1. Before testing, run rails db:migrate to update the database.
2. Upload a Nessus/Nmap export file with large data
3. Confirm that the file imports successfully.

### DIFF
--- a/config/initializers/z_11_plugin_templates.rb
+++ b/config/initializers/z_11_plugin_templates.rb
@@ -7,7 +7,7 @@
 
 
 # Unless the DB is already migrated, do nothing
-if Configuration.table_exists?
+if (ActiveRecord::Base.connection rescue false) && Configuration.table_exists?
   plugin_dir = nil
   source_dir = nil
   destination_dir = nil
--- a/config/initializers/z_11_plugin_templates.rb
+++ b/config/initializers/z_11_plugin_templates.rb
@@ -5,6 +5,11 @@
 # When such facility exists we can replace the content of this loop with a
 # simple call to (for instance) <plugin>.copy_templates
 
+# Rails 5.2 now loads the environment on all db rake tasks. Previously it did
+# not for setup or create. This means if the db doesn't exist this code will
+# fail. Technically we shouldn't communicate with AR during initializers so this
+# is workaround.
+# https://github.com/rails/rails/issues/32870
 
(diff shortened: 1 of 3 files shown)


config/initializers/z_11_plugin_templates.rb
```diff
@@ -7,7 +7,7 @@
 
 
 # Unless the DB is already migrated, do nothing
-if Configuration.table_exists?
+if (ActiveRecord::Base.connection rescue false) && Configuration.table_exists?
```
sean-yeoh, line 10: @aapomm What's the purpose of this part added here?

db/migrate/20200806120434_change_node_properties_to_long_text.rb
```diff
@@ -0,0 +1,7 @@
+class ChangeNodePropertiesToLongText < ActiveRecord::Migration[5.2]
+  TEXT_BYTES = 1_073_741_823
+
+  def change
+    change_column :nodes, :properties, :text, limit: TEXT_BYTES
```
sean-yeoh, line 5: @aapomm Just wondering... how does 1_073_741_823 get translated to `limit: 4294967295` after migration?

aapomm: From the `db/migrate/20151119124120_create_versions.rb` file:

>   The largest text column available in all supported RDBMS is
  1024^3 - 1 bytes, roughly one gibibyte.  We specify a size
  so that MySQL will use `longtext` instead of `text`.  Otherwise,
  when serializing very large objects, `text` might not be big enough.

config/initializers/z_11_plugin_templates.rb
aapomm: > Rails 5.2 now loads the environment on all db rake tasks. Previously it did
> not for setup or create. This means if the db doesn't exist this code will
> fail. Technically we shouldn't communicate with AR during initializers so this
> is workaround.

Note that this is a backport from Pro.

PR #733, opened 2020-08-10
Image modal

MattBudz:
### Summary

This PR adds the ability to enlarge images by clicking on them.

### Check List

- [x] Added a CHANGELOG entry


app/assets/javascripts/tylium/modules/image_modals.js
```diff
@@ -0,0 +1,14 @@
...
+  $('[data-behavior~=content-textile] img').each(function() {
+    $(this).attr('data-toggle', 'modal').attr('data-target', '[data-behavior~=image-modal]');
+  });
+
+  $('[data-behavior~=content-textile] img').click(function() {
+    var title = $(this).attr('alt');
+
+    $modal.find('[data-behavior~=modal-title]').text(title);
+    $modal.find('[data-behavior~=image-modal-image]').attr('src', $(this).attr('src'));
+  });
```
brianp, line 13: Let me see if I understand this.

1. We loop the content area of textile rendered html and assign a data attribute to images for them to target the modal, I think this also binds the functionality for the modal to open?
2. Then assign the event to the images that swaps the modal content when the image is clicked.

I'm not in love with the process. It feels like we should be able to perform the behavior in a single loop, or potentially entirely on the click function. Maybe we need to find away to append the appropriate attributes somewhere else (within textile?)

MattBudz: @brianp I assume the code to add the data-attributes to the img tags would be in `lib/html/pipeline/dradis_textile_filter.rb`

I'm not sure how to add the attributes, I'm assuming we would have to patch the formatter? I chose this route since it seemed like a much simpler option but by all means, I'm open to another approach. 

I didn't see any options for this in the RedCloth gem.

CHANGELOG
```diff
@@ -1,4 +1,5 @@
 [v#.#.#] ([MMM’YY])
+  * Enlarge images on click
```
brianp, line 2: We could probably be more specific about _which_ images will enlarge, and how they'll enlarge.
"Images imported into issues will now expand to fill the screen when clicked" kinda thing. *Sell it!*

app/assets/javascripts/shared/behaviors.js
```diff
@@ -6,4 +6,27 @@ document.addEventListener('turbolinks:load', function(){
   $('[data-behavior~=rich-toolbar]').each(function() {
     new EditorToolbar($(this));
   });
-})
+
+  // Activate Modals for images within textile rendered areas
```
etdsoft, line 10: 20 lines of initialisation seems like enough to be moved to their own file instead of leaving them here. I think we need an easier "if this is a thing in the page, trigger the full init, which lives elsewhere" approach.

app/assets/javascripts/shared/behaviors.js
```diff
@@ -6,4 +6,27 @@ document.addEventListener('turbolinks:load', function(){
...
+  // The ImageModal class will control the navigation of the image carousel. In
+  // this case the modal is managed by bootstrap. Bootstrap brings the modal up
+  // on click, we access the needed images as some may have been ajaxed in
+  // (comments, or the editor preview) and add the images to the ImageModal
+  // object to be displayed within the modal.
+  const $modal = $('[data-behavior~=image-modal]');
```
etdsoft, line 16: this seems like a constant that is only used inside `ImageModal`, it should probably live inside that class too.

app/assets/javascripts/shared/behaviors.js
```diff
@@ -6,4 +6,27 @@ document.addEventListener('turbolinks:load', function(){
...
+    modal = new ImageModal([], 0, $modal);
+
+    scope = ($(this).parents('[data-behavior~=comment-feed]').length ? 'comment-feed' : 'content-textile');
+    $(this).parents('[data-behavior~='+scope+']').find(targetImgs).each(function() {
+      modal.addImage(this);
+    });
+
+    var index = modal.getIndex(this);
+    modal.loadImage(index);
+  });
```
etdsoft, line 31: I'm not sure of our architecture tbh.

Click to enlarge? Cool!

Do we really need a slideshow of all the screenshots in a given stream? Not so sure.

If I click on the screenshot on a comment from Alice, I don't think I care about being able to navigate to a screenshot 7 comment up posted by Bob using the chevron. It seems we went too far for a v1 of this feature.

From a technical perspective:

- We're installing a new `click` handler for every image.
- However in each handler, we're reusing a global variable (`modal`) that we `unbind()` and reinistantiate every single time.
- But this `ImageModal` instance *always* does the same in the same page: load all images available. If we click on 3 images, we unbind and reinstantiate 3 times, doing exactly the same every time.
- L24 is a smell, `data-behaviors` can be combined, here we don't need to choose which one is on the page, we need to add a *new* data-behavior for this *new* behavior we're adding.
- L 25 also feels out of place, we're already looking for all `targetImgs` in L20, then we're in a single image, click handler context, but we do so again. Seems to me that the `ImageModal` init and `addImage` don't belong to the click handler. But as I already mentioned, I don't think we need the extra bells and whistles at all.

app/assets/javascripts/tylium.js
```diff
@@ -41,6 +41,7 @@
 //= require tylium/modules/configurations
 //= require tylium/modules/export
 //= require tylium/modules/fileupload
+//= require tylium/modules/image_modals
```
etdsoft, line 44: I think this should be /shared/

app/assets/stylesheets/tylium/modules/_modals.scss
```diff
@@ -3,6 +3,45 @@
   .modal-dialog {
     max-width: 560px;
 
+    &.image-modal {
```
etdsoft, line 6: probably /shared/ too

app/views/cards/show.html.erb
```diff
@@ -51,7 +51,7 @@
                 <h4 class="mb-4 header-underline"><%= @card.name %></h4>
 
                 <div class="card-description">
-                  <div class="content-textile">
+                  <div class="content-textile" data-behavior="content-textile">
```
etdsoft, line 54: As mentioned above, I'm not sure about this technique to "scope" the work.

app/views/comments/_comment.html.erb
```diff
@@ -38,7 +38,7 @@
 
       </div>
 
-      <div class="content content-textile">
+      <div class="content content-textile" data-behavior="content-textile">
```
etdsoft, line 41: ditto

app/views/layouts/tylium.html.erb
```diff
@@ -62,7 +62,8 @@
     <%= render 'activities/poller' %>
 
     <!-- All modals go here -->
-    <%= render 'layouts/tylium/modals' %>
+    <%= render partial: 'layouts/tylium/image_modals' %>
+    <%= render partial: 'layouts/tylium/modals' %>
```
etdsoft, line 66: if we already have a `modals` and you're adding a modal... let's render it from inside it...

... or you may have the wrong name? you're adding a single modal not a bunch of modals, so then perhaps what you need is a `tylium/modals/image` or `tylium/modals/image_preview` or something along those lines?

app/views/nodes/show/_services_table.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="content-textile">
+<div class="content-textile" data-behavior="content-textile">
```
etdsoft, line 1: ditto and 👇

lib/dradis/ce.rb
```diff
@@ -14,3 +14,4 @@ def self.version
 require 'html/pipeline/dradis_fieldable_filter'
 require 'html/pipeline/dradis_mentions_filter'
 require 'html/pipeline/dradis_textile_filter'
+require 'html/pipeline/dradis_image_modal_attributes_filter'
```
etdsoft, line 17: :alpha

vendor/assets/javascripts/jquery.textile.js
```diff
@@ -30,7 +30,7 @@
         tpl: {
           fields: '<div class="textile-form h-100 col-6 col-sm-6"></div>',
           wrap: '<div class="textile-wrap"><ul class="textile-toolbar"></ul><div class="textile-inner row"></div></div>',
-          preview: '<div class="col-6 col-sm-6"><div class="textile-preview loading-indicator">Loading...</div></div>',
+          preview: '<div class="col-6 col-sm-6" data-behavior="content-textile"><div class="textile-preview loading-indicator">Loading...</div></div>',
```
etdsoft, line 33: ditto

etdsoft (changes requested):

PR #735, opened 2020-08-11
Upgrade Rails to 6.1.1 and Ruby to 2.7.2

brianp:
### Summary

Currently Rails is on version 5.2.4, and Ruby is version 2.4.1. Ruby 2.4 has reached end of life in March 2020. We would like to upgrade to Ruby ~2.7.1~ 2.7.2. Rails 5.2.4 does not support the newest version of ruby. So we would also like to upgrade Rails to ~6.0.3.2~ 6.1.1. This would bring us to both Rails and Ruby newest modern versions.

**NOTE:** Branch name is outdated but being kept the same despite now upgrading to a newer version.

### Solution

Step through version increments upgrading the application and removing deprecation warnings until we hit the newest versions.

### Notable Changes

- `update_attributes` to `update` Lots of small lines with this change. `update_attributes` is no longer supported.
- I updated documentation or inline comments created by the framework so our comments would be up to date. So lots of text changes that may or may not mean much to us. Really I tried to make it seem like we generated the project with this version of Rails.
- Creation of default files in bin/ and .git. Along with removal of some bin/ files

### Copyright assignment

### Check List

- [ ] Added a CHANGELOG entry


.gitattributes
```diff
@@ -0,0 +1,10 @@
+# See https://git-scm.com/docs/gitattributes for more about git attribute files.
```
sean-yeoh, line 1: TIL about git attributes and github linguist https://medium.com/@bolajiayodeji/introducing-github-linguist-8f09273ddea1

Gemfile
```diff
@@ -27,13 +27,13 @@ gem 'local_time', '>= 2.0.0'
 gem 'turbolinks', '~> 5'
 
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
-gem 'jbuilder', '~> 2.5'
+gem 'jbuilder', '~> 2.7'
 
-# Use ActiveStorage variant
-# gem 'mini_magick', '~> 4.8'
+# Use Active Storage variant
+# gem 'image_processing', '~> 1.2'
```
sean-yeoh, line 33: Can these be removed?

app/assets/javascripts/tylium/modules/issues.js.coffee
```diff
@@ -1,5 +1,5 @@
-//= require tylium/modules/issues/evidence
-//= require tylium/modules/issues/importer
-//= require tylium/modules/issues/merge
-//= require tylium/modules/issues/table
-//= require tylium/modules/issues/tag-input
+#= require tylium/modules/issues/evidence
+#= require tylium/modules/issues/importer
+#= require tylium/modules/issues/merge
+#= require tylium/modules/issues/table
+#= require tylium/modules/issues/tag-input
```
sean-yeoh, line 5: Just wondering, isn't the syntax `//=`?

brianp: The asset pipeline syntax is the same, but the coffeescript syntax changed in coffeescript 4. So any coffeescript files with requires need to use this.

sean-yeoh: Ahh okay, got it.

Gemfile
brianp: They can but we left them there as a nice reminder that someday we should probably move out propitiatory attachment management over to Active Storage 😄

.gitignore
```diff
@@ -13,12 +13,15 @@
 !/log/.keep
 !/tmp/.keep
 
-# Ignore uploaded files in development
+# Ignore pidfiles, but keep the directory.
+/tmp/pids/*
+!/tmp/pids/
+!/tmp/pids/.keep
```
etdsoft, line 19: It would seem these 3 lines can be combined into one and moved up to the /log/ and /tmp/ section?

etdsoft, line 19: It would seem these 3 lines can be combined into one and moved up to the /log/ and /tmp/ section?

Gemfile
```diff
@@ -56,15 +56,15 @@ gem 'html-pipeline'
 
 gem 'kaminari', '~> 1.2.1'
 
-gem 'paper_trail', '~> 10.3'
+gem 'paper_trail', git: 'git@github.com:paper-trail-gem/paper_trail.git', ref: '1e56afd'
```
etdsoft, line 59: 🙈

Gemfile
```diff
@@ -206,7 +208,7 @@ end
 #
 
 # Base framework classes required by other plugins
-gem 'dradis-plugins', '~> 3.20'
+gem 'dradis-plugins', git: 'git@github.com:dradis/dradis-plugins.git', branch: 'upgrade/rails-6.0.3.2'
```
etdsoft, line 211: No need, this is now merged to `master`

bin/rails
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
+load File.expand_path("spring", __dir__)
```
etdsoft, line 2:
```suggestion
load File.expand_path('spring', __dir__)
```

bin/rake
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
-require_relative '../config/boot'
-require 'rake'
+load File.expand_path("spring", __dir__)
+require_relative "../config/boot"
+require "rake"
```
etdsoft, line 4:
```suggestion
load File.expand_path('spring', __dir__)
require_relative '../config/boot'
require 'rake'
```

config/environments/development.rb
```diff
@@ -1,10 +1,11 @@
...
   # Settings specified here will take precedence over those in config/application.rb.
 
-  # In the development environment your application's code is reloaded on
-  # every request.  This slows down response time but is perfect for development
-  # since you don't have to restart the webserver when you make code changes.
+  # In the development environment your application's code is reloaded any time
+  # it changes. This slows down response time but is perfect for development
+  # since you don't have to restart the web server when you make code changes.
   config.cache_classes = false
+  config.action_controller.enable_fragment_cache_logging = true
```
etdsoft, line 8: This seems like a dup of L20?

Can't spot it here:
http://railsdiff.org/5.2.4.4/6.1.1

config/environments/development.rb
```diff
@@ -49,19 +54,26 @@
...
   # routes, locales, etc. This feature depends on the listen gem.
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
-  # In development, we want any emails to be opened in the browser
+  # In development, we want any emails to be opened in the browser.
   config.action_mailer.delivery_method = :letter_opener
 
+  # Uncomment if you wish to allow Action Cable access from any origin.
+  # config.action_cable.disable_request_forgery_protection = true
+  config.action_cable.url = "wss://localhost:3000/cable"
```
etdsoft, line 75: where is this coming from?

I'm concerned about the `-p` param to change port number and this being hard-coded.

config/initializers/assets.rb
```diff
@@ -15,4 +15,4 @@
 Rails.application.config.assets.precompile += %w( application.css application.js tylium.js tylium.css bootstrap.css)
 
 # Include font files
-Rails.application.config.assets.precompile += [/\.(?:svg|eot|woff|ttf)\z/]
+Rails.application.config.assets.precompile << ["*.svg", "*.eot", "*.woff", "*.ttf"]
```
etdsoft, line 18: as far as I can tell, we don't use .woff, or .eot; not even .svg...

etdsoft (approved):

config/environments/development.rb
brianp: hmmm. Maybe a copy and paste error.

config/initializers/assets.rb
brianp: Font awesome uses those formats so this is likely for vendor compatibility. We do package them in our assets.

image

Gemfile
brianp: :shipit: 🚀 🤞🏻

config/environments/development.rb
brianp: I've gone back through the steps that lead me to this and I think it was a db/page reload red-herring. I believe i've reproduced it, and this just isn't needed. Will remove.

PR #739, opened 2020-08-27
Diff by word rather than by line

rachkor:
### Summary
The revision history currently checks by line. This PR updates the revision history to see changes by word instead. This should offer more granularity as far as the changes made. 

### Check List
- [x] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,7 @@
   * Editor: Allow drag & drop, copy & paste, and direct image uploading
   * Increase the node properties column size by changing it to LONGTEXT
   * Layout: Breadcrumbs have a fixed position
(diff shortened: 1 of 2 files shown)


brianp:
Just going through PR's. Pasting comments from elsewhere: I've taken a look at this. The library we're using for diffing has 2 other by_word related bugs listed on github, one of which looks like this issue. The library is quite old and does not receive updates.

I'd say if by_word is more buggy, and less consistent the the best path forward is do nothing at all. by_line was likely chosen because it's more reliable.

We will not be implementing this patch at the moment as it looks like it will introduce more issues for users than it solves. It could absolutely be worth investigating better diffing at some point but that's more suitable for a pitch.

PR #755, opened 2020-12-02
Memoize pwd in file_backed_model

sean-yeoh:
### Summary
This PR memoizes pwd in file_backed_model.rb so that it is more efficient.

### Check List
- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/models/concerns/file_backed_model.rb
+++ b/app/models/concerns/file_backed_model.rb
@@ -51,10 +51,10 @@ module FileBackedModel
     # Returns a Pathname to location configured in the database through the
     # :setting_name setting
     def pwd
-      return @pwd if defined?(@pwd)
-
-      conf = Configuration.find_or_create_by(name: @setting_name, value: @default_path)
-      @pwd = Pathname.new(conf.value)
+      @pwd ||= begin
+        conf = Configuration.find_or_create_by(name: @setting_name, value: @default_path)
+        Pathname.new(conf.value)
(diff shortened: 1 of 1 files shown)


app/models/concerns/file_backed_model.rb
```diff
@@ -51,8 +51,10 @@ def from_file(filename)
     # Returns a Pathname to location configured in the database through the
     # :setting_name setting
     def pwd
+      return @pwd if defined?(@pwd)
```
aapomm, line 54: Let's follow the same format in methodology model: https://github.com/dradis/dradis-ce/blob/master/app/models/methodology.rb#L44

sean-yeoh: Updated.

app/models/concerns/file_backed_model.rb
```diff
@@ -51,8 +51,10 @@ def from_file(filename)
     # Returns a Pathname to location configured in the database through the
     # :setting_name setting
     def pwd
-      conf = Configuration.find_or_create_by(name: @setting_name, value: @default_path)
-      Pathname.new(conf.value)
+      @pwd ||= begin
+        conf = Configuration.find_or_create_by(name: @setting_name, value: @default_path)
```
etdsoft, line 55: does this work, don't you need for this to be `create_with`, also, shouldn't entire thing be in `Configuration`?

PR #756, opened 2020-12-03
Add file rename concern

sean-yeoh:
### Summary
This PR extracts `Attachment.available_name` into a concern so that it can be reused elsewhere.

### Check List

- [ ] Added a CHANGELOG entry


app/services/naming_service.rb
```diff
@@ -0,0 +1,18 @@
+class NamingService
+  # If filename exists within the directory,
+  # return a count based alternative, E.g. file_name-copy-01.png, file_name-copy-02.png
```
etdsoft, line 3:
```suggestion
  # return a count based alternative, E.g. filename_copy-01.png, filename_copy-02.png
```

app/services/naming_service.rb
```diff
@@ -0,0 +1,18 @@
...
+  def self.name_file(original_filename:, pathname:)
+    if File.exists?(pathname.join(original_filename))
+      extension = File.extname(original_filename)
+      basename = File.basename(original_filename, extension)
+      files = Dir.glob(pathname.join("#{basename}_copy-*#{extension}"))
+      sequence = files.map { |file| file.match(/_copy-([0-9]+)#{extension}\z/)[1].to_i }.max || 0
+      "%s_copy-%02i%s" % [basename, sequence + 1, extension]
+    else
+      original_filename
+    end
```
etdsoft, line 16: would the code be cleaner with an early return? (I don't know)

```ruby
  file_exists = File.exists?(pathname.join(original_filename))
  return original_filename if file_exists

  [...]
```

etdsoft (approved):
Looking good, minor stuff only.


(🤓 it's not a concern though! 🙈)

app/services/naming_service.rb
sean-yeoh: Good suggestion, thanks!

app/services/naming_service.rb
sean-yeoh: :+1:

PR #757, opened 2020-12-04
Use view hooks for export index view

aapomm:
Instead of using the `@plugins` variable and iterating through every export plugins, we are adding a render view hook that export plugins should use. 

One important change here is adding an argument to `render_view_hooks` to specify the addon feature:

```
def render_view_hooks(partial, feature = :addon, locals: {})
```

to include the export plugins.


app/views/export/index.html.erb
```diff
@@ -1,40 +1,26 @@
...
+      <p class="mb-0">Check the Export section of your <code><%= Rails.root.join('Gemfile.plugins') %></code> file.</p>
     </div>
-  <% else %>
+  </div>
+<% else %>
 
-    <ul id="plugin-chooser" class="tabs-container nav nav-tabs">
-      <% @plugin_info.each do |info| %>
+  <ul id="plugin-chooser" class="tabs-container nav nav-tabs">
+    <%= render_view_hooks 'export/plugin_description', :export%>
```
sean-yeoh, line 13:
```suggestion
    <%= render_view_hooks 'export/plugin_description', :export% >
```

app/helpers/application_helper.rb
```diff
@@ -16,8 +16,8 @@ def markup(text)
     result[:output].to_s.html_safe
   end
 
-  def render_view_hooks(partial, locals: {})
-    Dradis::Plugins::with_feature(:addon).sort_by(&:plugin_description).each do |plugin|
+  def render_view_hooks(partial, feature = :addon, locals: {})
```
sean-yeoh, line 19: @aapomm We've added this functionality in Pro during the permissions cycle https://github.com/secroots/dradis-pro/blob/develop/app/helpers/application_helper.rb#L23.

Can you take care of the merge conflicts once this is merged?

aapomm: I missed that one. Thanks!

etdsoft:
@aapomm why do we need the `:feature` change? 

Traditionally we've relied on plugins resitering the `:addon` if they'd like to provide a View extension (no matter where), and then on the implementation of `.render_view_hooks()` to rescue from `ActionView::MissingTemplate` if the given plugin didn't provide a particular view for a particular hook.

I'm worried about the proliferation of approaches around `render_view_hooks`, for instance: should we use `feature: :calculator` when building the Issues#show tab? Or `feature: :menu` when rendering the Mintcreek Add-ons menu?

I'm not sure what the best way forward here is.

app/views/export/index.html.erb
```diff
@@ -1,40 +1,26 @@
...
+      <p class="mb-0">Check the Export section of your <code><%= Rails.root.join('Gemfile.plugins') %></code> file.</p>
     </div>
-  <% else %>
+  </div>
+<% else %>
 
-    <ul id="plugin-chooser" class="tabs-container nav nav-tabs">
-      <% @plugin_info.each do |info| %>
+  <ul id="plugin-chooser" class="tabs-container nav nav-tabs">
+    <%= render_view_hooks 'export/plugin_description', feature: :export %>
```
etdsoft, line 13: I'd like for us to keep naming consistent: for Issues#show calculator tabs we're using:

```erb
<%= render_view_hooks 'issues/show-tabs' %>
```

and 

```
<%= render_view_hooks 'issues/show-content' %>
```

What about using `export/index-tabs` and `export/index-content` here?

PR #761, opened 2021-01-22
Fix deprecation warnings for autoloaded constants

sean-yeoh:
### Summary
This PR fixes the deprecation warnings for:
1. Autoloading constants in initializers https://github.com/rails/rails/issues/36363
2. Capybara Support locator type https://github.com/teamcapybara/capybara/pull/2143
3. Rendering partials with "." in template name https://github.com/rails/rails/commit/dd9991bac598bb5da312278a749cf85e19b027cc

We have to wrap constants that are referenced inside a `to_prepare` block.
```ruby
Rails.application.reloader.to_prepare do
# Some code that references constant here.
end
```
Reference:
https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoloading-when-the-application-boots

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/views/issues/update.json.erb
+++ b/app/views/issues/update.json.erb
@@ -1,7 +1,7 @@
 {
   "success": <%= @modified %>,
   <% if tag = @issue.tags.first %>
-  "issue_link": "<%= j render('issues/issue.html.erb', issue: @issue) %>",
+  "issue_link": "<%= j render(partial: 'issues/issue', issue: @issue, format: 'html') %>",
   "tag_cell": "<%= j "<td style=\"color:#{tag.color}\"><i class=\"fa fa-bug\"></i> #{tag.display_name}</td>".html_safe %>"
   <% end %>
 }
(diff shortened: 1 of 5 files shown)


app/views/issues/update.json.erb
```diff
@@ -1,7 +1,7 @@
 {
   "success": <%= @modified %>,
   <% if tag = @issue.tags.first %>
-  "issue_link": "<%= j render('issues/issue.html.erb', issue: @issue) %>",
+  "issue_link": "<%= j render(partial: 'issues/issue', locals: { issue: @issue }, formats: 'html') %>",
```
etdsoft, line 4: @sean-yeoh fyi you can replace this syntax (`{locals: { issue: @issue } }` with `object: @issue`)

etdsoft (approved):

app/views/issues/update.json.erb
sean-yeoh: Ehh.. cool, it works. I'll do that actually!

PR #771, opened 2021-01-29
Use 6.0 defaults

sean-yeoh:
### Summary
This PR uses Rails 6.1 defaults and removing defaults generated by `rails app:update`.

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


app/controllers/concerns/nodes_sidebar.rb
```diff
@@ -7,6 +7,6 @@ def initialize_nodes_sidebar
     # won't be able to generate a route for a Note with no id) So filter out
     # Notes that haven't been saved:
     @sorted_notes    = @node.notes.sort_by(&:title).select(&:persisted?)
-    @sorted_evidence = @node.evidence.sort_by { |e1| e1.issue.title }
+    @sorted_evidence = @node.evidence.select(&:persisted?).sort_by { |e1| e1.issue.title }
```
aapomm, line 10: small comment, but this should be consistent with line#9. can we call `select` before `sort` in that line too?

sean-yeoh: Updated, thanks!

PR #776, opened 2021-02-03
Empty states nodes show

MattBudz:
### Summary

Update empty states for each tab in `nodes#show` as well as the `Nodes` section in the main sidebar

### Check List


app/assets/javascripts/tylium/modules/nodes.js.coffee
```diff
@@ -20,3 +20,12 @@ document.addEventListener "turbolinks:load", ->
...
     $('[data-behavior~=services-extras-tab-pane]').first().addClass('active')
+
+    $('[data-behavior~=evidence-dropdown-toggle]').click (e) ->
+      e.stopPropagation()
+      $('[data-name~=evidence]').click()
+
+    $('[data-behavior~=notes-dropdown-toggle]').click (e) ->
+      e.stopPropagation()
+      $('[data-name~=notes]').click()
+
```
brianp, line 31: What's the deal here? We're taking a click from one place and pushing it onto something else because?

app/views/nodes/notes/_empty_state_actions.html.erb
```diff
@@ -0,0 +1,7 @@
+<p>
+  Use the 
+  <%= link_to 'javascript:void(0)', data: { behavior: 'notes-dropdown-toggle' } do -%>
```
brianp, line 3: Can't this partial and the one that's exactly the same be re-used. The only difference is `evidence-dropdown-toggle` vs `notes-dropdown-toggle`. I've pretty sure we can pull a type or resource type or something from somwhere.

app/assets/javascripts/tylium/modules/nodes.js.coffee
MattBudz: This is so we can show the sidebar menu using the empty state CTA. We can't just add the same data attribute to the CTA because bootstrap expects the menu to be a sibling of the button

etdsoft (approved):

PR #777, opened 2021-02-04
update boards#show empty state

MattBudz:
### Summary

Update boards#show empty state

### Check List


app/views/boards/_board.html.erb
```diff
@@ -11,85 +11,94 @@
...
-        </a>
-      </li>
-    </ul>
+        <li class="list list-new">
+          <a
+            class="js-list-modal"
+            href="#modal-list-new"
+          >
+            Add a list...
+          </a>
```
brianp, line 91: I realize this isn't a change in the PR but I just have to point at this and ask the world why 😄

brianp (approved):

PR #792, opened 2021-03-01
Shared subscriptions assets and controller

sean-yeoh:
### Summary
This PR moves the subscription js to shared/ and make subscription routes and controller to be independent.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


app/controllers/subscriptions_controller.rb
```diff
@@ -1,12 +1,10 @@
 class SubscriptionsController < AuthenticatedController
-  include ProjectScoped
-
   def create
     subscription = Subscription.new(subscription_params)
     subscription.user = current_user
     subscription.save
 
-    redirect_back fallback_location: project_path(current_project), notice: 'Subscribed!'
+    redirect_back fallback_location: params[:fallback_path], notice: 'Subscribed!'
```
etdsoft, line 7: let's use a sensible default (as before) instead of  user-controllable parameter.

`redirect_back` should already work in most normal cases, this is in case the browser didn't send a Referer header.

app/controllers/subscriptions_controller.rb
```diff
@@ -17,7 +15,7 @@ def destroy
     )
     subscription.destroy
 
-    redirect_back fallback_location: project_path(current_project), notice: 'Unsubscribed!'
+    redirect_back fallback_location: params[:fallback_path], notice: 'Unsubscribed!'
```
etdsoft, line 18: ditto

etdsoft (changes requested):

app/controllers/subscriptions_controller.rb
sean-yeoh: Fixed.

app/controllers/subscriptions_controller.rb
sean-yeoh: Fixed.

PR #797, opened 2021-03-05
Implement #to_xml for Board, List, Card, Comment, and Activity

etdsoft:
### Summary

Put the responsibility of knowing how to turn each of this classes into an XML representation on themselves (instead of on the Template Exporter of dradis/dradis-plugins.

### Check List

- [x] Added a CHANGELOG entry

Skipped, this is an internal refactor, not really user-facing.

### DIFF
--- a/app/models/card.rb
+++ b/app/models/card.rb
@@ -48,6 +48,42 @@ class Card < ApplicationRecord
     super(ids)
   end
 
+  def to_xml(xml_builder, includes: [], version: 3)
+    xml_builder.card do |card_builder|
+      card_builder.id(id)
+      card_builder.name(name)
+      card_builder.description do
+        card_builder.cdata!(description)
+      end
+      card_builder.due_date(due_date)
+      card_builder.previous_id(previous_id)
+
+      if includes.include?(:assignees)
+        card_builder.assignees do |assignee_builder|
+          assignees.each do |assignee|
+            assignee_builder.assignee(assignee.email)
+          end
+        end
+      end
+
+      if includes.include?(:activities)
+        card_builder.activities do |activity_builder|
+          activities.each do |activity|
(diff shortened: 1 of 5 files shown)


app/models/card.rb
```diff
@@ -48,6 +48,42 @@ def assignee_ids=(ids)
...
+      card_builder.due_date(due_date)
+      card_builder.previous_id(previous_id)
+
+      if includes.include?(:assignees)
+        card_builder.assignees do |assignee_builder|
+          assignees.each do |assignee|
+            assignee_builder.assignee(assignee.email)
+          end
+        end
+      end
```
brianp, line 67: Cam we not push these additions down onto their respective classes?

app/models/card.rb
```diff
@@ -48,6 +48,42 @@ def assignee_ids=(ids)
...
+
+      if includes.include?(:activities)
+        card_builder.activities do |activity_builder|
+          activities.each do |activity|
+            activity.to_xml(activity_builder)
+          end
+        end
+      end
```
brianp, line 75: And here.

app/models/card.rb
```diff
@@ -48,6 +48,42 @@ def assignee_ids=(ids)
...
+
+      if includes.include?(:comments)
+        card_builder.comments do |comment_builder|
+          comments.each do |comment|
+            comment.to_xml(comment_builder)
+          end
+        end
+      end
```
brianp, line 83: And here.

brianp (changes requested):
What about pulling these out into a classes that represent xml for these data types.

I'm thinking like jbuilder but for xml, and not views. We're already using a builder. We could reverse it- have a builder and pass in the data object? 

Have a directory for them `app/models/xml/activity.rb` etc

PR #805, opened 2021-03-17
Backport activity's project_id removal from Pro

aapomm:


app/models/project.rb
```diff
@@ -21,6 +21,10 @@ def eql?(item)
     end
   end
 
+  def ==(item)
+    eql?(item)
+  end
+
```
sean-yeoh, line 27: @aapomm What's this method for? I don't see it on the pro repo.

aapomm: This is added to make `project_instance == another_project_instance` work. This is used in the notifications `by_project` method. In Pro, this is already implemented since the Project is an ActiveRecord model.

PR #827, opened 2021-04-19
Convert SSH git remotes to HTTPS

imander:
### Summary

This PR fixes an issue with git remotes being referenced via SSH rather than HTTPS. Using SSH breaks unattended installs (i.e. docker) of Dradis due to host key verification and the dependency of a private key.

### Copyright assignment

### Check List

- [x] Convert SSH git remotes to HTTPS


imander:
Fixed in #858

PR #835, opened 2021-04-29
Api token

aapomm:
### Spec

The Basic Authentication currently used in the CE API does not work for the purpose of using the endpoints in-app. This is because to generate the token, we need a copy of the user’s password.

**Proposed solution**

Use a generated API token for authentication.

### How to test

1. Create a user if none exist yet.
2. Open the rails console and run the following command to show the API token:

```
> User.find(<user_id>).api_token
```

3. Run the command in your terminal to test the API:

```
$ curl -k -H 'Authorization: Token token="<token>"' https://localhost:3000/api/issues
```

4. Confirm you get the issues in the project.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -20,5 +20,12 @@ class User < ApplicationRecord
   scope :enabled, -> { all }
 
   # -- Class Methods --------------------------------------------------------
+  def self.authenticate(email, pass)
+    user = find_or_create_by(email: email)
+    return user if user && BCrypt::Password.new(::Configuration.shared_password) == pass
+  end
+
   # -- Instance Methods -----------------------------------------------------
+
+  ActiveSupport.run_load_hooks(:user_model, self)
 end
--- a/engines/dradis-api/db/migrate/20150722125412_add_api_token_to_users.rb
+++ b/engines/dradis-api/db/migrate/20150722125412_add_api_token_to_users.rb
@@ -1,0 +1,21 @@
+class AddAPITokenToUsers < ActiveRecord::Migration[5.1]
+  # Future-proof this migration by ignoring everything in the actual model.
+  class User < ActiveRecord::Base; end
+
+  def up
+    add_column :users, :api_token, :string
+    add_index :users, :api_token, unique: true
+
+    # Create an API token for all existing users
+    User.reset_column_information
+    User.find_each do |user|
+      user.update_column(:api_token, SecureRandom.base64(18).tr("+/", "-_"))
+    end
+    change_column :users, :api_token, :string, null: false
+  end
+
+  def down
+    remove_column :users, :api_token, :string
+  end
+
+end
--- a/engines/dradis-api/db/migrate/20150722125412_add_api_token_to_users.rb
+++ b/engines/dradis-api/db/migrate/20150722125412_add_api_token_to_users.rb
@@ -1,17 +1,14 @@
 class AddAPITokenToUsers < ActiveRecord::Migration[5.1]
-  # Future-proof this migration by ignoring everything in the actual model.
-  class User < ActiveRecord::Base; end
-
   def up
-    add_column :users, :api_token, :string
+    add_column :users, :api_token, :string, null: false
     add_index :users, :api_token, unique: true
 
-    # Create an API token for all existing users
-    User.reset_column_information
-    User.find_each do |user|
-      user.update_column(:api_token, SecureRandom.base64(18).tr("+/", "-_"))
+    User.transaction do
+      User.find_each do |user|
+        # Create an API token for all existing users
+        user.update_column(:api_token, SecureRandom.base64(18).tr("+/", "-_"))
+      end
     end
-    change_column :users, :api_token, :string, null: false
   end
 
   def down
--- a/engines/dradis-api/lib/dradis/ce/api.rb
+++ b/engines/dradis-api/lib/dradis/ce/api.rb
@@ -7,5 +7,6 @@ end
 require 'dradis/ce/api/catch_json_parse_errors'
 require 'dradis/ce/api/routing_constraints'
 require 'dradis/ce/api/version'
-require 'dradis/ce/api/warden_strategy'
+require 'dradis/ce/api/api_token_strategy'
+require 'dradis/ce/api/basic_auth_strategy'
 require 'dradis/ce/api/engine'
--- a/engines/dradis-api/lib/dradis/ce/api/api_token_strategy.rb
+++ b/engines/dradis-api/lib/dradis/ce/api/api_token_strategy.rb
@@ -1,13 +1,5 @@
-# Dradis::API::WardenStrategy
-#
-# HTTP Basic authentication strategy for Warden.
-#
-# See:
-#   https://github.com/hassox/warden
-#
 module Dradis::CE::API
-  class WardenStrategy < ::Warden::Strategies::Base
-
+  class ApiTokenStrategy < ::Warden::Strategies::Base
     attr_reader :email, :password
 
     # This strategy should be applied when we are requesting /api/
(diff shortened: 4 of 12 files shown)


engines/dradis-api/db/migrate/20150722125412_add_api_token_to_users.rb
```diff
@@ -0,0 +1,21 @@
+class AddAPITokenToUsers < ActiveRecord::Migration[5.1]
+  # Future-proof this migration by ignoring everything in the actual model.
```
brianp, line 2: This is neat but is it the way we normally handle this situation?

engines/dradis-api/lib/dradis/ce/api/warden_strategy.rb
```diff
@@ -32,11 +38,26 @@ def authenticate!
...
+    def secure_compare(a)
+      b = token
+      return false if a.blank? || b.blank? || a.bytesize != b.bytesize
+      l = a.unpack "C#{a.bytesize}"
+
+      res = 0
+      b.each_byte { |byte| res |= byte ^ l.shift  }
+      res == 0
+    end
+
```
brianp, line 60: We weren't using this before with token authentication in Pro, why add it now?

engines/dradis-api/lib/dradis/ce/api/warden_strategy.rb
```diff
@@ -16,12 +16,18 @@ def valid?
...
-      if auth.provided? && auth.basic? && auth.credentials
-        username = auth.credentials.first
+      if token
+        user = User.enabled.find_by_api_token(token)
+        if user && secure_compare(user.api_token)
+          success!(user)
+        else
+          custom!(unauthorized(403))
+        end
+      elsif auth.provided? && auth.basic? && auth.credentials
```
brianp, line 26: Another way we can do this is with two different warden strategies. One that accepts basic auth, and one that accepts token auth. Then at some point maybe in the controller, or in a routing strategy we can switch to the strategy we want to use. It keeps each strategy class specifics it it's _strategy_ where strategy is tokenized, or basic auth, or username/pw etc.

Up to you if you see value in that, or would prefer to keep it in a single location.

engines/dradis-api/lib/dradis/ce/api/warden_strategy.rb
aapomm: This was ported from the Pro warden strategy and used there too, so this isn't anything new.

engines/dradis-api/db/migrate/20150722125412_add_api_token_to_users.rb
aapomm: Updated to modern times ✅

engines/dradis-api/lib/dradis/ce/api/warden_strategy.rb
aapomm: This is now split into two strategies: `token_auth` and `basic_auth`.

app/models/user.rb
```diff
@@ -20,5 +20,12 @@ class User < ApplicationRecord
   scope :enabled, -> { all }
 
   # -- Class Methods --------------------------------------------------------
+  def self.authenticate(email, pass)
+    user = find_or_create_by(email: email)
+    return user if user && BCrypt::Password.new(::Configuration.shared_password) == pass
+  end
+
```
etdsoft, line 27: I'm not sure we need to add this method:

1. It seems to only be used by the Basic Auth strategy (more comments below)
2. It'd diverge from `User#authenticate` in Pro.

engines/dradis-api/db/migrate/20150722125412_add_api_token_to_users.rb
```diff
@@ -0,0 +1,18 @@
+class AddAPITokenToUsers < ActiveRecord::Migration[5.1]
+  def up
+    add_column :users, :api_token, :string, null: false
+    add_index :users, :api_token, unique: true
+
+    User.transaction do
+      User.find_each do |user|
+        # Create an API token for all existing users
+        user.update_column(:api_token, SecureRandom.base64(18).tr("+/", "-_"))
```
etdsoft, line 9: I know this is copied code, but in the spirit of leaving better than you found it:

1. This token is of a different length from the one defined in the file above
2. It doesn't use the `urlsafe` version
3. Applies a different set of 'readability' translations

engines/dradis-api/lib/dradis/ce/api.rb
```diff
@@ -7,5 +7,6 @@ module API; end
 require 'dradis/ce/api/catch_json_parse_errors'
 require 'dradis/ce/api/routing_constraints'
 require 'dradis/ce/api/version'
-require 'dradis/ce/api/warden_strategy'
+require 'dradis/ce/api/api_token_strategy'
+require 'dradis/ce/api/basic_auth_strategy'
```
etdsoft, line 11: two `_strategy` suffixes is a smell.

I think that we also have a problem int he main app where we define strategies inline in the initializers when they should live under `./lib/dradis/ce/warden`.

I think the API strategies should live inside the API engine's `./lib/dradis/ce/strategies/api` folder. Then you can remove the `_strategy` suffix.

engines/dradis-api/lib/dradis/ce/api/api_token_strategy.rb
```diff
@@ -0,0 +1,50 @@
+module Dradis::CE::API
+  class ApiTokenStrategy < ::Warden::Strategies::Base
+    attr_reader :email, :password
+
+    # This strategy should be applied when we are requesting /api/
+    def valid?
+      !!(request.path_info =~ /\A\/api\//)
```
etdsoft, line 7: The goal of `valid?` is to guard early against unmet conditions for the strategy.

I think that L11, the presence of `token` is a prerequisite for the strategy to apply and should be moved to the `valid?` method.

That way it becomes explicit that a token is required, and the the `authenticate!` method becomes cleaner.

engines/dradis-api/lib/dradis/ce/api/api_token_strategy.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    def authenticate!
+      if token
+        user = User.enabled.find_by_api_token(token)
+        if user && secure_compare(user.api_token)
```
etdsoft, line 13: nothing to change here, but I'm a bit stumped by the double check. I believe it's because `secure_compare` doesn't give away when the token doesn't match (as having a single DB query would)

engines/dradis-api/lib/dradis/ce/api/api_token_strategy.rb
etdsoft, line 7: on a separate note, our second API strategy is missing the `path_info` check we have here, I think we should add it as well.

engines/dradis-api/lib/dradis/ce/api/basic_auth_strategy.rb
```diff
@@ -1,13 +1,5 @@
...
-#
-# HTTP Basic authentication strategy for Warden.
-#
-# See:
-#   https://github.com/hassox/warden
-#
 module Dradis::CE::API
-  class WardenStrategy < ::Warden::Strategies::Base
-
+  class BasicAuthStrategy < ::Warden::Strategies::Base
```
etdsoft, line 2: this strategy quacks a lot like the [:shared_password](https://github.com/dradis/dradis-ce/blob/3377d759a15008e577be9cfb0af9938034415910/config/initializers/warden_00_shared_password.rb) one.

I think it's worth creating a `./lib/dradis/ce/warden/shared_password` strategy that has an `authenticate!` method that uses the `username` and `password` values to perform the secure check (removing the need to have `User.authenticate` as mentioned above). In the `SharedPassword` strategy the `username` method can take the value from the `params` and in the API streategy we can override the method to take the value from `auth.credentials`.

That way we end up with 1 place in the entire codebase where we're doing the `BCrypt` check for the shared password.

If you think that's too much (I could buy that argument particularly because the "fail" condition in both strategies isn't standard. Then I'd also buy two separate strategies classes in which the `authenticate!` method is pretty much identical (relies on `username`, `password`, and does the `BCrypt` check. Then we move the implementation of where the username comes from outside the `authenticate!` method to a separate `def username` method.

engines/dradis-api/lib/dradis/ce/api/basic_auth_strategy.rb
```diff
@@ -17,11 +9,10 @@ def valid?
 
     def authenticate!
       if auth.provided? && auth.basic? && auth.credentials
```
etdsoft, line 11: even if this is how the code _was_ these checks belong to `#valid?` and not to `#authenticate!`

etdsoft (changes requested):
The two things I'd like to see improved are:

* The unnecessary `User.authenticate` method
* The weirdness / duplication around the regular shared password strategy and the one used in the API endpoint

PR #836, opened 2021-04-29
Paginate api

aapomm:
### Summary

Currently, the API endpoints don’t paginate their data. We need the pagination behavior for the DataTables server-side fetching.

**Proposed solution**

Add an optional param: `page` that fetches 25 entries at a time.

### How to test

1. Create a project with a bunch of data.
2. Take note of a user’s api token.
3. Run the following command:

```
curl -k -H 'Authorization: Token token=<token>\n' http://localhost:3000/api/issues\?page\=1
```

4. Confirm that you get only upto 25 entries.
5. Run the command again but this time with `page=2`
6. Test the command in different endpoints: nodes, notes, evidence.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


engines/dradis-api/app/controllers/dradis/ce/api/v1/evidence_controller.rb
```diff
@@ -8,6 +8,7 @@ class EvidenceController < Dradis::CE::API::APIController
 
       def index
         @evidence = @node.evidence.all.order('updated_at desc')
```
etdsoft, line 10: Seeing `.all.` is missing from Issues and Nodes, but pressent in Evidence and Notes, what gives?

Also, it is my understanding that `.all.`, in recent versions of Rails, has turned into a scope and it no longer transforms the relation into an array. Can you confirm that?

aapomm: > Also, it is my understanding that .all., in recent versions of Rails, has turned into a scope and it no longer transforms the relation into an array.

At the moment it returns `AssociationRelation` vs `CollectionProxy` when not using `.all`. In this use case, I believe keeping it as the `CollectionProxy` makes more sense, so I'll remove the `.all` call.

etdsoft (approved):

PR #838, opened 2021-04-30
Datatables replace itemtables

sean-yeoh:
### Summary

Provide a general description of the code changes in your pull
request... were there any bugs you had fixed? If so, mention them. If
these bugs have open GitHub issues, be sure to tag them here as well,
to keep the conversation linked together.

### Other Information

If there's anything else that's important and relevant to your pull
request, mention that information here. This could include
benchmarks, or other information.

Thanks for contributing to Dradis!

### Copyright assignment

Collaboration is difficult with commercial closed source but we want
to keep as much of the OSS ethos as possible available to users
who want to fix it themselves.

In order to unambiguously own and sell Dradis Framework commercial
products, we must have the copyright associated with the entire
codebase. Any code you create which is merged must be owned by us.
That's not us trying to be a jerks, that's just the way it works.

Please review the [CONTRIBUTING.md](https://github.com/dradis/dradis-ce/blob/master/CONTRIBUTING.md)
file for the details.

You can delete this section, but the following sentence needs to
remain in the PR's description:

> I assign all rights, including copyright, to any future Dradis
> work by myself to Security Roots.

### Check List

- [ ] Added a CHANGELOG entry

PR #842, opened 2021-05-06
Datatables toggle column visibility

sean-yeoh:
### Summary
This PR shows default columns for DataTables and add ability to hide/show columns.
**Note:** Do note merge until https://github.com/dradis/dradis-ce/pull/838 is merged.

### Prerequisites
https://github.com/dradis/dradis-ce/pull/838

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -1,16 +1,49 @@
...
+      return;
+    }
+
     this.dataTable = null;
+    this.tableHeaders = Array.from(this.$table[0].querySelectorAll('thead th, thead td'));
     this.init();
   }
 
   init() {
+    var colvisColumnIndexes = this.tableHeaders.reduce(function(indexes, column, index) {
```
aapomm, line 15: What's the reason for using `reduce` here? According to the MDN docs:

> The reduce() method executes a reducer function (that you provide) on each element of the array, resulting in a single output value.

By my understanding, it looks like we're doing some sort of `map` here to output an array not a single output value. Can we use this instead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map ? Or `forEach`?

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -21,8 +54,20 @@ class DradisDatatable {
   unbindDataTable() {
     var that = this;
 
+    that.hideColumns();
```
aapomm, line 57: Is this related to the datatable being unbind or should this be called in under the `behaviors()` method?

app/assets/javascripts/shared/dradis_datatable.js
sean-yeoh: I wanted to iterate over an array and filter it based on the `data-colvis="false"` attribute and return the array.

But I think I can probably use standard way.

app/assets/javascripts/shared/dradis_datatable.js
sean-yeoh: > Is this related to the datatable being unbind or should this be called in under the behaviors() method?

Eh, you're right. Thanks for pointing it out.

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -2,30 +2,68 @@ class DradisDatatable {
   constructor(tableElement) {
     this.$table = $(tableElement);
     this.dataTable = null;
+    this.tableHeaders = Array.from(this.$table[0].querySelectorAll('thead th, thead td'));
```
etdsoft, line 5: We don't need `td` and `th`, if there's an instance of `thead > td` we need to fix the HTML source to turn that into a `th`

etdsoft, line 5: We don't need `td` and `th`, if there's an instance of `thead > td` we need to fix the HTML source to turn that into a `th`

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -2,30 +2,68 @@ class DradisDatatable {
...
     this.init();
   }
 
   init() {
-    var dataTable = this.$table.DataTable({
+    // Remove dropdown option for <th> columns that has data-colvis="false" in colvis button
```
etdsoft, line 10: comment explains what the code do (the code does that already) can we explain the why instead?

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -2,30 +2,68 @@ class DradisDatatable {
...
+    var colvisColumnIndexes = [];
```
etdsoft, line 11: `colvis` seems to be a name we've chosen, correct? If so, I'd say we could probably choose a more descriptive name that doesn't require the reader to think so hard about what is going on? `column-visible`?

etdsoft (changes requested):

app/assets/images/check.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 20 20'><path fill='none' stroke='#fff' stroke-linecap='round' stroke-linejoin='round' stroke-width='3' d='M6 10l3 3l6-6'/></svg>
```
etdsoft, line 1: let's give this an extra line at the end.

app/assets/javascripts/shared/dradis_datatable.js
sean-yeoh: Changed from `'thead th, thead td'` to `thead th`.

app/assets/javascripts/shared/dradis_datatable.js
sean-yeoh: It removes the option from the dropdown.

image

app/assets/javascripts/shared/dradis_datatable.js
sean-yeoh: Thanks, updated to something more readable.

PR #843, opened 2021-05-06
Datatables checkbox

aapomm:
### Summary

The current implementation of the Datatable, is still based on the old behavior of the checkboxes. This doesn’t work with the built-in events that we can hook on when implementing the Issues merge, multi delete, etc.

**Proposed solution**

Implement this: https://datatables.net/extensions/select/examples/initialisation/checkbox.html

### How to test

1. Open the issues page in a project.
2. Confirm that you can select multiple rows.
3. Confirm that the Select all button works as expected.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -2,35 +2,129 @@ class DradisDatatable {
...
+        "<'row'<'col-lg-12'tr>>" +
+        "<'dataTables_footer_content'ip>",
       initComplete: function (settings) {
         settings.oInstance.wrap("<div class='table-wrapper'></div>");
       },
-      pageLength: 25
+      lengthChange: false,
+      pageLength: 25,
+      select: {
+        selector: 'td:first-child',
```
sean-yeoh, line 60: Shouldn't this be `td.select-checkbox`?

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -2,35 +2,129 @@ class DradisDatatable {
...
+    // Remove default datatable button listener to make the checkbox "checking"
+    // work, before adding our own click handler.
+    $selectAllBtn.off('click.dtb').click( function (){
+      if (that.areAllSelected()) {
+        that.dataTable.rows().deselect();
+      }
+      else {
+        that.dataTable.rows().select();
+      }
+    });
```
sean-yeoh, line 122: I'm still kinda confused with this, why is this needed?

aapomm: @sean-yeoh we're not using the Datatables' built-in selectAll button since it doesn't fit our use case. That means we have to implement the selectAll/deselectAll functionality ourselves.

app/assets/javascripts/shared/dradis_datatable.js
aapomm: Both works, but I like the `select-checkbox` more for the specificity.

etdsoft (approved):

PR #850, opened 2021-05-18
Datatables state save

sean-yeoh:
### Summary
This PR adds state saving to DataTables.

### Prerequisites
https://github.com/dradis/dradis-ce/pull/851

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/assets/javascripts/shared/dradis_datatable.js
+++ b/app/assets/javascripts/shared/dradis_datatable.js
@@ -5,6 +5,8 @@ class DradisDatatable {
     this.tableHeaders = Array.from(this.$table[0].querySelectorAll('thead th'));
     this.$paths = this.$table.closest('[data-behavior~=datatable-paths]');
     this.itemName = this.$table.data('item-name');
+    this.localStorageKey = this.$table.data('local-storage-key');
+    this.initialPageLoad = false;
     this.init();
   }
 
@@ -65,24 +67,9 @@ class DradisDatatable {
       lengthChange: false,
       pageLength: 25,
       stateSave: true,
-      stateSaveCallback: function(settings, data) {
-        var dataTableDomId = settings.oInstance[0].id;
-        var localStorageKey = `Datatables_${dataTableDomId}_${window.location.pathname}`;
-
-        // Check if the table has been loaded before so that we can skip
-        // the manually hidden columns on page load.
-        var initialLoad = false;
-        if (localStorage.getItem(localStorageKey) === null) {
-          initialLoad = true;
-        }
-
-        data.initialLoad = initialLoad;
-        localStorage.setItem(localStorageKey, JSON.stringify(data));
-      },
+      stateSaveCallback: this.setupLocalStorage.bind(this),
       stateLoadCallback: function(settings) {
-        var dataTableDomId = settings.oInstance[0].id;
-        var localStorageKey = `Datatables_${dataTableDomId}_${window.location.pathname}`;
-        return JSON.parse(localStorage.getItem(localStorageKey));
+        return JSON.parse(localStorage.getItem(that.localStorageKey));
       },
       select: {
         selector: 'td.select-checkbox',
@@ -222,11 +209,7 @@ class DradisDatatable {
   }
 
   hideColumns() {
-    var dataTableDomId = this.$table.attr('id');
-    var localStorageKey = `Datatables_${dataTableDomId}_${window.location.pathname}`;
-    var data = JSON.parse(localStorage.getItem(localStorageKey));
-
-    if (data && data.initialLoad) {
+    if (this.initialPageLoad) {
       // Hide columns that has data-hide-on-load="true" on initial page load
       var that = this;
       that.tableHeaders.forEach(function(column, index) {
@@ -286,6 +269,14 @@ class DradisDatatable {
         that.dataTable.rows().select();
       }
     });
+  }
+
+  setupLocalStorage(settings, data) {
+    if (localStorage.getItem(this.localStorageKey) === null) {
+      this.initialPageLoad = true;
+    }
+
+    localStorage.setItem(this.localStorageKey, JSON.stringify(data));
   }
 
   areAllSelected() {
--- a/app/assets/javascripts/shared/dradis_datatable.js
+++ b/app/assets/javascripts/shared/dradis_datatable.js
@@ -6,7 +6,6 @@ class DradisDatatable {
     var defaultColumns = this.$table.data('default-columns') || [];
     this.defaultColumns = defaultColumns.concat(['Select', 'Actions']);
     this.itemName = this.$table.data('item-name');
-    this.localStorageKey = this.$table.data('local-storage-key');
     this.tableHeaders = Array.from(this.$table[0].querySelectorAll('thead th'));
     this.init();
   }
@@ -46,7 +45,7 @@ class DradisDatatable {
         },
         buttons: [
           {
-            available: function(){
+            available: function() {
               return that.$table.find('[data-behavior~=select-checkbox]').length;
             },
             attr: {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -64,6 +64,26 @@ class DradisDatatable {
       },
       lengthChange: false,
       pageLength: 25,
+      stateSave: true,
+      stateSaveCallback: function(settings, data) {
+        var dataTableDomId = settings.oInstance[0].id;
+        var localStorageKey = `Datatables_${dataTableDomId}_${window.location.pathname}`;
```
aapomm, line 70: This looks like a variable we can save as a class variable, then accessible with `this.localStorageKey`

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -64,6 +64,26 @@ class DradisDatatable {
       },
       lengthChange: false,
       pageLength: 25,
+      stateSave: true,
+      stateSaveCallback: function(settings, data) {
+        var dataTableDomId = settings.oInstance[0].id;
```
aapomm, line 69: If this is the same as `var dataTableDomId = this.$table.attr('id');`, then I think we can save this as a class variable since it will never change.

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -64,6 +64,26 @@ class DradisDatatable {
       },
       lengthChange: false,
       pageLength: 25,
+      stateSave: true,
+      stateSaveCallback: function(settings, data) {
```
aapomm, line 68: Let's defined this function as a class function to keep the dataTable definition simple

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -64,6 +64,26 @@ class DradisDatatable {
...
+        var dataTableDomId = settings.oInstance[0].id;
+        var localStorageKey = `Datatables_${dataTableDomId}_${window.location.pathname}`;
+
+        // Check if the table has been loaded before so that we can skip
+        // the manually hidden columns on page load.
+        var initialLoad = false;
```
aapomm, line 74: Do we really need this boolean? Can we just check if there is an "initial load" if the localstorage is empty or not? We access the localstorage in the `hideColumns` method anyway.

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -64,6 +64,26 @@ class DradisDatatable {
...
+        if (localStorage.getItem(localStorageKey) === null) {
+          initialLoad = true;
+        }
+
+        data.initialLoad = initialLoad;
+        localStorage.setItem(localStorageKey, JSON.stringify(data));
+      },
+      stateLoadCallback: function(settings) {
```
aapomm, line 82: If you'll move the variables to the class, we can simplify this function to a single line, which I don't think requires it be a class method as it'll be simple enough.

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -202,14 +222,20 @@ class DradisDatatable {
...
+    var localStorageKey = `Datatables_${dataTableDomId}_${window.location.pathname}`;
+    var data = JSON.parse(localStorage.getItem(localStorageKey));
+
+    if (data && data.initialLoad) {
+      // Hide columns that has data-hide-on-load="true" on initial page load
+      var that = this;
+      that.tableHeaders.forEach(function(column, index) {
+        if (column.dataset.hideOnLoad == 'true') {
+          var dataTableColumn = that.dataTable.column(index);
+          dataTableColumn.visible(false);
```
aapomm, line 235:
```suggestion
          that.dataTable.column(index).visible(false);
```

app/assets/javascripts/shared/dradis_datatable.js
sean-yeoh: Refactored it into method instead

app/assets/javascripts/shared/dradis_datatable.js
sean-yeoh: Moved to the class's attribute.

app/assets/javascripts/shared/dradis_datatable.js
sean-yeoh: Opted to use `data-local-storage-key="key"` instead as it provides more flexibility

app/assets/javascripts/shared/dradis_datatable.js
sean-yeoh: Opted to use data-local-storage-key="key" and saved it as a the object's attribute

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -4,7 +4,9 @@ class DradisDatatable {
     this.$paths = this.$table.closest('[data-behavior~=datatable-paths]');
     this.dataTable = null;
     this.defaultColumns = this.$table.data('default-columns') || [];
+    this.initialPageLoad = false;
```
etdsoft, line 7: Why are we going the inverse path? Shouldn't the default initializer behavior to be: initialPageLoad = true; then we check local storage and flip that to false when checking that the key is really present?

Also, it seems that this behavior is closely related to `setupDefaultColumns`, shouldn't that method be only called from within the `setupLocalStorage` method (when the key isn't found), instead of it being called all the time?

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -302,6 +309,14 @@ class DradisDatatable {
     });
   }
 
+  setupLocalStorage(settings, data) {
+    if (localStorage.getItem(this.localStorageKey) === null) {
+      this.initialPageLoad = true;
+    }
```
etdsoft, line 315: as mentioned in the init code, this is the reverse logic.

this method should have an if {} else {} structure, if it's the initial page load, then do the default columns thing, otherwise, load whatever was already selected.

Bonus points, those two actions (the default columns and the show/hide from storage) are actually the same action, with different inputs (the list of columns to show).

etdsoft (changes requested):

app/assets/javascripts/shared/dradis_datatable.js
```diff
@@ -20,6 +22,18 @@ class DradisDatatable {
...
+    var hiddenColumnIndexes = [];
+    if (localStorage.getItem(this.localStorageKey) === null) {
+      this.tableHeaders.forEach(function(column, index) {
+        var columnName = column.textContent.trim();
+
+        if (!that.defaultColumns.includes(columnName)) {
+          hiddenColumnIndexes.push(index);
+        }
+      });
+    }
```
etdsoft, line 35: Same anti pattern:

1. comment starts talking about the columns you will *show*
2. variable name flips that to track the *hidden* columns
3. when there's no local storage you loop through a list of columns you want to *show* (the default list)
4. but check that the current column is *not* on that list, and then you add it to the list you want to *hide*

Since what you're passing as "default columns" is the show list, and what you're storing in local storage is the "show list" too, I think this code can be made clearer.

etdsoft (approved):
Happy with the improvements, I found the logic flow of show/hide method to be a bit awkward though...

app/assets/javascripts/shared/dradis_datatable.js
sean-yeoh: Yeah it is a bit weird, but it is done this way because there is no way to tell datatables to show only specific columns as all columns are shown by default. So I had to do it reversely by hiding certain columns, instead of _only_ showing certain columns.

The data in local storage is handled by datatables itself. So the flow currently is:
1. Script checks if local storage key exists
1.a) If yes, datatables load the previous state

2. If no:
2.a) script specifies columns to hide
2.b) Datatables automatically save the state to local storage

 An example of the data that datatables saves to local storage is shown below. How they hide/show columns is by passing in the `visible: true` or `visible: false` property to the elements in the columns array.

```javascript
{
    "columns": [
        {
            "search": {
                "caseInsensitive": true,
                "regex": false,
                "search": "",
                "smart": true
            },
            "visible": true
        },
        {
            "search": {
                "caseInsensitive": true,
                "regex": false,
                "search": "",
                "smart": true
            },
            "visible": true
        },
        {
            "search": {
                "caseInsensitive": true,
                "regex": false,
                "search": "",
                "smart": true
            },
            "visible": true
        },
        {
            "search": {
                "caseInsensitive": true,
                "regex": false,
                "search": "",
                "smart": true
            },
            "visible": true
        },
        {
            "search": {
                "caseInsensitive": true,
                "regex": false,
                "search": "",
                "smart": true
            },
            "visible": false
        },
        {
            "search": {
                "caseInsensitive": true,
                "regex": false,
                "search": "",
                "smart": true
            },
            "visible": false
        },
        {
            "search": {
                "caseInsensitive": true,
                "regex": false,
                "search": "",
                "smart": true
            },
            "visible": true
        },
        {
            "search": {
                "caseInsensitive": true,
                "regex": false,
                "search": "",
                "smart": true
            },
            "visible": true
        },
        {
            "search": {
                "caseInsensitive": true,
                "regex": false,
                "search": "",
                "smart": true
            },
            "visible": true
        },
        {
            "search": {
                "caseInsensitive": true,
                "regex": false,
                "search": "",
                "smart": true
            },
            "visible": true
        }
    ],
    "length": 25,
    "order": [
        [
            0,
            "asc"
        ]
    ],
    "search": {
        "caseInsensitive": true,
        "regex": false,
        "search": "",
        "smart": true
    },
    "start": 0,
    "time": 1621489207448
}
```

PR #854, opened 2021-05-20
Add initial implementation of LiquidFilter

etdsoft:
### Summary

Adds a new filter to the rendering pipeline that allows us to have Liquid markup in user-provided content.

### Check List

- [ ] Added a CHANGELOG entry


app/controllers/concerns/liquid_enabled_resource.rb
```diff
@@ -0,0 +1,39 @@
...
+    helper_method :liquid_assigns
+  end
+
+  def liquid_assigns
+    @liquid_assigns ||= default_liquid_assigns.merge!(liquid_resource_assigns)
+  end
+
+  # To be overwritten by each controller
+  def liquid_resource_assigns
+    raise 'Controller included LiquidEnabledResource without defining a liquid_resource_assigns'
```
aapomm, line 14: What's the reason for enforcing this overwrite vs just using a `{}` as a default? That way, we won't have to define it on every controller that doesn't use this yet.

app/controllers/concerns/liquid_enabled_resource.rb
```diff
@@ -0,0 +1,39 @@
...
+    else
+      {}
+    end
+  end
+
+  def project_assigns
+    result = {}
+
+    # This is required because we may be in Markup#preview that's passing
+    # :project_id for Tilyum rendered editors
```
aapomm, line 31:
```suggestion
    # :project_id for Tylium rendered editors
```

app/helpers/application_helper.rb
```diff
@@ -1,21 +1,28 @@
...
+      HTML::Pipeline::AutolinkFilter,
+      HTML::Pipeline::Dradis::CodeHighlightFilter
+    ]
+
+    if options[:liquid]
+      context[:liquid_assigns] = liquid_assigns
+      pipeline_filters.insert(1, HTML::Pipeline::Dradis::LiquidFilter)
+    end
+
+    if options[:filters] && options[:filters].any?
```
aapomm, line 19: Can be simplified to

```suggestion
    if options[:filters].present?
```

app/controllers/concerns/liquid_enabled_resource.rb
```diff
@@ -0,0 +1,39 @@
...
+    project = Project.find(params[:project_id])
+    # authorize! :use, project
```
aapomm, line 33: I'm assuming this is a Pro-only code. If so, can we go with this for parity:

```suggestion
    if defined?(Dradis::Pro)
      authorize! :use, project
    end
```

app/controllers/concerns/liquid_enabled_resource.rb
etdsoft: You want the controller to have a way to define the local assigns, and you want the concern to include that when defined. Either you default to {} in the concern and do a `defined?` call to see if the controller defines a new collection, or you do something like this.

We can also take the HasFields approach:
https://github.com/dradis/dradis-ce/blob/main/app/models/concerns/has_fields.rb#L4-L9

app/controllers/concerns/liquid_enabled_resource.rb
etdsoft: I dislike those types of checks, but it's a good point re parity, I wonder what would it take (if anything) to have the `authorize!` call just work in CE.

app/controllers/concerns/liquid_enabled_resource.rb
brianp: I'm with @aapomm on this. I don't like the need for a bunch of mostly empty methods added in every controller. Applying `{}` as the default and overriding it as needed feels better. I don't think that requires the use of a `defined?` check. It's just method overriding.

Even that might be improved by a strict usage instead of method overrides. Something more along the [lines of permissions](https://github.com/dradis/dradis-plugins/blob/main/app/controllers/concerns/dradis/plugins/persistent_permissions.rb#L36-L39) so the assigns get defined as the requirement for the module is added. Although this is just a psudo example and I think instance variables would be hard to access like this.

```
include LiquidEnabledResource

liquid_assigns proc { { 'issue' => IssueDrop.new(@issue) } }
```

app/helpers/markup_helper.rb
```diff
@@ -1,10 +1,13 @@
...
+    preview_params[:project_id] = params[:project_id] if params[:project_id]
+
     {
-      form_url: form_fields_path(format: :js),
-      help_url: help_markup_path,
-      preview_url: preview_markup_path,
-      source_url: source_fields_path
+      form_url: main_app.form_fields_path(format: :js),
+      help_url: main_app.help_markup_path,
+      preview_url: main_app.preview_markup_path(preview_params),
```
brianp, line 9: What was the catalyst for this change. We're rendering somewhere outside of a project with the full editor?

app/helpers/application_helper.rb
```diff
@@ -1,21 +1,28 @@
...
+    context = {}
+    pipeline_filters = [
+      HTML::Pipeline::Dradis::FieldableFilter,
+      HTML::Pipeline::Dradis::TextileFilter,
+      HTML::Pipeline::SanitizationFilter,
+      HTML::Pipeline::AutolinkFilter,
+      HTML::Pipeline::Dradis::CodeHighlightFilter
+    ]
+
+    if options[:liquid]
```
brianp, line 14: This would just become: 

```suggestion
    if args.include? :liquid
```

app/helpers/application_helper.rb
```diff
@@ -1,21 +1,28 @@
 module ApplicationHelper # :nodoc:
-  def markup(text, filters = [])
+  def markup(text, options={})
```
brianp, line 2: Could we do something like

```suggestion
  def markup(text, *args)
```

Which would make the usage a bit cleaner. `markup(@issue.text, :liquid)` instead of `markup(@issue.text, liquid: true)`

etdsoft: started there but abandoned, because the options can have nested values, in our case we're immediately using filters: [filter1, filter2]

app/helpers/markup_helper.rb
etdsoft: Most Tylium controllers are aware of the project context via ProjectScoped, but MarkupController that renders the Preview isn't. We need for a way to let Preview know about the project context we're in.

app/controllers/concerns/liquid_enabled_resource.rb
etdsoft: fixed in 7ff6745b

app/controllers/concerns/liquid_enabled_resource.rb
brianp: >  wonder what would it take (if anything) to have the authorize! call just work in CE.

The Pro CE ability files are already different by a large margin. We could likely add a `can :manage, Project` in CE which would solve this?

brianp: Resolved in [a new PR](https://github.com/dradis/dradis-ce/pull/873/).

PR #859, opened 2021-05-24
Handle attachment newlines

aapomm:
### Summary

Uploading a file with newlines causes an error in the node.

**Proposed Solution**
Remove newlines and extra spaces with .squish before saving the attachments.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,9 +4,7 @@
   * [feature ...]
   * Upgraded gems: capybara, rails, rdoc, rubocop, nokogiri, papertrail, puma,
   * Bugs fixed:
-    - [bug fixed #1]
-    - [bug fixed ...]
-    - Bug tracker items: #N, #M, #O, ...
+    - Fix uploading attachments with newlines
   * New integrations:
     - [new integration #1]
     - [new integration ...]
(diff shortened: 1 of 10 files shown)


app/controllers/attachments_controller.rb
```diff
@@ -17,7 +17,7 @@ def create
     uploaded_file = params.fetch(:attachment_file, params.fetch(:files, []).first)
 
     attachment_name = NamingService.name_file(
-      original_filename: uploaded_file.original_filename,
+      original_filename: uploaded_file.original_filename.squish,
```
sean-yeoh, line 20: Aaron, can we do the fix in the NamingService class instead?

CHANGELOG
```diff
@@ -4,9 +4,7 @@
   * [feature ...]
   * Upgraded gems: capybara, rails, rdoc, rubocop, nokogiri, papertrail, puma,
   * Bugs fixed:
-    - [bug fixed #1]
-    - [bug fixed ...]
-    - Bug tracker items: #N, #M, #O, ...
+    - Fix uploading attachments with newlines
```
etdsoft, line 7: I'm not sure this describes the issue at hand as it makes it sound it's about the content of the attachment

PR #860, opened 2021-05-24
Fix/datatables buttons

MattBudz:
### Summary

- hide all actions buttons after successfully bulk-deleting 
- disable all action buttons while performing an action

### Check List

### DIFF
--- a/app/assets/javascripts/shared/datatable/main.js
+++ b/app/assets/javascripts/shared/datatable/main.js
@@ -143,10 +143,12 @@ class DradisDatatable {
     }
   }
 
-  toggleLoadingState(rows, isLoading, buttonName) {
-    var button = this.dataTable.buttons(buttonName + ':name');
+  toggleLoadingState(rows, isLoading) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/shared/datatable/main.js
```diff
@@ -135,10 +135,12 @@ class DradisDatatable {
...
 
-  toggleLoadingState(rows, isLoading, buttonName) {
-    var button = this.dataTable.buttons(buttonName + ':name');
+  toggleLoadingState(rows, isLoading) {
+    var buttons = this.dataTable.buttons('button');
 
-    $(button[0].node).toggleClass('disabled', isLoading);
+    for (var i = 1; i < buttons.length - 1; i++) {
+      $(buttons[i].node).toggleClass('disabled', isLoading);
+    }
```
aapomm, line 143: Can this be used here instead: https://datatables.net/reference/api/buttons().nodes() ? Something like: `buttons.nodes().toggleClass('disabled', isLoading);`

PR #862, opened 2021-05-28
Reinitialize Active Storage

brianp:
### Summary

Active Storage was setup during a previous upgrade but never used. This PR brings it back to life.

### Check List

- [] ~Added a CHANGELOG entry~

### DIFF
--- a/db/migrate/20210316202859_add_service_name_to_active_storage_blobs.active_storage.rb
+++ b/db/migrate/20210316202859_add_service_name_to_active_storage_blobs.active_storage.rb
@@ -1,0 +1,18 @@
+# This migration comes from active_storage (originally 20190112182829)
+class AddServiceNameToActiveStorageBlobs < ActiveRecord::Migration[6.0]
+  def up
+    unless column_exists?(:active_storage_blobs, :service_name)
+      add_column :active_storage_blobs, :service_name, :string
+
+      if configured_service = ActiveStorage::Blob.service.name
+        ActiveStorage::Blob.unscoped.update_all(service_name: configured_service)
+      end
+
+      change_column :active_storage_blobs, :service_name, :string, null: false
+    end
+  end
+
+  def down
+    remove_column :active_storage_blobs, :service_name
+  end
+end
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -89,7 +89,8 @@ RSpec.configure do |config|
       FactoryBot.lint
     ensure
       DatabaseCleaner.clean_with(:truncation)
-      Attachment.all.each(&:delete)
+      FileUtils.rm_rf(Attachment.pwd + '/*')
+      FileUtils.rm_rf(Rails.root.join('tmp/storage'))
     end
   end
 
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -89,7 +89,7 @@ RSpec.configure do |config|
       FactoryBot.lint
     ensure
       DatabaseCleaner.clean_with(:truncation)
-      FileUtils.rm_rf(Attachment.pwd + '/*')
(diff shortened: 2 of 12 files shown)


db/migrate/20210316202859_add_service_name_to_active_storage_blobs.active_storage.rb
```diff
@@ -0,0 +1,18 @@
+# This migration comes from active_storage (originally 20190112182829)
+class AddServiceNameToActiveStorageBlobs < ActiveRecord::Migration[6.0]
+  def up
+    unless column_exists?(:active_storage_blobs, :service_name)
```
sean-yeoh, line 4: What's the `service_name` column for?

spec/rails_helper.rb
```diff
@@ -90,6 +90,7 @@
     ensure
       DatabaseCleaner.clean_with(:truncation)
       Attachment.all.each(&:delete)
+      FileUtils.rm_rf(Rails.root.join("tmp/storage"))
```
sean-yeoh, line 93:
```suggestion
      FileUtils.rm_rf(Rails.root.join('tmp/storage'))
```

db/migrate/20210316202859_add_service_name_to_active_storage_blobs.active_storage.rb
brianp: If we were using AWS, or some other third party file storage I believe the name of the service used to store that particular blob will be added to that column. Not sure what ours will contain. Maybe blank, maybe "filestore".

sean-yeoh: Ahh okay, move along then.

db/migrate/20210316202859_add_service_name_to_active_storage_blobs.active_storage.rb
```diff
@@ -0,0 +1,18 @@
...
+      add_column :active_storage_blobs, :service_name, :string
+
+      if configured_service = ActiveStorage::Blob.service.name
+        ActiveStorage::Blob.unscoped.update_all(service_name: configured_service)
+      end
```
etdsoft, line 9: this seems like a weird migration:

1. create a brand new column
2. update all recrods with a new default
3. update column definition to disallow null columns

Can't 1-2 be combined into a single `add_column` with `default: configured_service` call? It seems that we're defining a service for all ./config/environments/ possible.

Or perhaps you can combine 2-3 with a call to [change_column_null](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_column_null) providing a 4th argument of `configured_service`)

spec/rails_helper.rb
```diff
@@ -90,6 +90,7 @@
     ensure
       DatabaseCleaner.clean_with(:truncation)
       Attachment.all.each(&:delete)
```
etdsoft, line 92: I'm wondering if we could save some time with a `.rm_rf` call on `Attachment.pwd` here?

db/migrate/20210316202859_add_service_name_to_active_storage_blobs.active_storage.rb
etdsoft: Why do you need a guard against the column existing?

etdsoft (approved):

db/migrate/20210316202859_add_service_name_to_active_storage_blobs.active_storage.rb
brianp: Same here. These migrations were generated by the rails upgrade command so I never questioned the contents.

spec/rails_helper.rb
brianp: Seems like a good recommendation for an aside PR.

PR #873, opened 2021-07-21
Use an actual integer and allow authorization checks on project

brianp:
### Summary

We were using hacks because `authorize! :use, project` was failing with weird errors. This PR fixes the error which was the initialization of Project with string id `"1"` instead of `1`. With this change we can safely call authorize on project in CE, and have parity between CE and Pro.

### Check List


app/models/ability.rb
```diff
@@ -4,6 +4,7 @@ class Ability
   def initialize(user)
     user ||= User.new # guest user (not logged in)
 
+    can :manage, Project
```
etdsoft, line 7: I appreciate the versatility, but would go for parity in this case

brianp: Updated

PR #880, opened 2021-08-05
Resource sidebar toggle

MattBudz:
### Summary

This PR adds the ability to collapse the resource sidebar (comments, subscriptions, validation, etc)

### Check List

- [x] Added a CHANGELOG entry


app/assets/javascripts/tylium/modules/resource_sidebar.js
```diff
@@ -0,0 +1,24 @@
...
+
+    if (savedState == 'true') {
+      toggleResourceSidebar();
+    }
+
+    $resourceSidebarToggle.on('click', function() {
+      toggleResourceSidebar();
+    });
+  };
+});
```
aapomm, line 24: Add a newline

app/assets/javascripts/tylium/modules/resource_sidebar.js
```diff
@@ -0,0 +1,24 @@
...
+    const $resourceSidebar = $('[data-behavior~=resource-sidebar]'),
+          $resourceSidebarToggle = $('[data-behavior~=resource-sidebar-toggle]'),
+          localStorageKey = 'project.ce.resource-sidebar-collapsed',
+          savedState = localStorage.getItem(localStorageKey);
+
+    function toggleResourceSidebar() {
+      $resourceSidebar.toggleClass('collapsed');
+      $resourceSidebarToggle.find('i').toggleClass('fa-chevron-right fa-chevron-left');
+      localStorage.setItem(localStorageKey, $resourceSidebar.hasClass('collapsed'));
+      Turbolinks.clearCache();
```
aapomm, line 13: What's the reason for clearing the cache here? It seems weird to lose the turbolinks speed when toggling the resource sidebar.

app/assets/javascripts/tylium/modules/resource_sidebar.js
```diff
@@ -0,0 +1,24 @@
+document.addEventListener('turbolinks:load', function() {
+
+  if ($('[data-behavior~=resource-sidebar]').length) {
+    const $resourceSidebar = $('[data-behavior~=resource-sidebar]'),
+          $resourceSidebarToggle = $('[data-behavior~=resource-sidebar-toggle]'),
+          localStorageKey = 'project.ce.resource-sidebar-collapsed',
```
aapomm, line 6: Let's have this key in an HTML data attribute, similar to the datatable storage keys for consistency.

app/assets/javascripts/tylium/modules/resource_sidebar.js
```diff
@@ -0,0 +1,24 @@
...
+          savedState = localStorage.getItem(localStorageKey);
+
+    function toggleResourceSidebar() {
+      $resourceSidebar.toggleClass('collapsed');
+      $resourceSidebarToggle.find('i').toggleClass('fa-chevron-right fa-chevron-left');
+      localStorage.setItem(localStorageKey, $resourceSidebar.hasClass('collapsed'));
+      Turbolinks.clearCache();
+    }
+
+    if (savedState == 'true') {
```
aapomm, line 16: Let's use a more semantic check:
```suggestion
    if (savedState == 'collapsed') {
```

app/views/cards/show.html.erb
```diff
@@ -32,103 +32,74 @@
...
-                }
-              )
-            }
-          %>
+          </div>
         </div>
       </div>
     </div>
 
+    <%= render partial: 'shared/resource_sidebar', locals: { resource: @card, content: ['comments', 'subscriptions'] } %>
```
aapomm, line 103: What's the reason for passing `content` here? It seems like all calls to this partial are passing the same `content` values. Is there a case where there is a comments feed without subscriptions or vice versa?

app/assets/javascripts/tylium/modules/resource_sidebar.js
MattBudz: This is the same approach as the main sidebar, if we don't clear the cache there is a weird jump in the view where the sidebar reverts to the default state for a second and then back to what the user had it set to.

app/views/cards/show.html.erb
MattBudz: This is for pro where we will pass the validations section to some views.

app/assets/javascripts/tylium/modules/resource_sidebar.js
MattBudz: fixed with a slightly different approach

PR #885, opened 2021-08-17
import css manifests from engines

MattBudz:
### Summary

This PR import's css manifests from engines if any exist

### Check List

### DIFF
--- a/app/assets/stylesheets/tylium/engines.css.erb
+++ b/app/assets/stylesheets/tylium/engines.css.erb
@@ -1,0 +1,11 @@
+<%
+Dradis::Plugins::with_feature(:addon).sort_by(&:plugin_description).each do |plugin|
+  begin
+    plugin_path = ActiveSupport::Inflector.underscore(ActiveSupport::Inflector.deconstantize(plugin.name))
+    %>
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/tylium/engines.css.erb
```diff
@@ -0,0 +1,11 @@
...
+Dradis::Plugins::with_feature(:addon).sort_by(&:plugin_description).each do |plugin|
+  begin
+    plugin_path = ActiveSupport::Inflector.underscore(ActiveSupport::Inflector.deconstantize(plugin.name))
+    %>
+    <%= require_asset "#{plugin_path}/manifests/tylium" %>
+    <%
+  rescue Sprockets::FileNotFound;
+  end
+end
+%>
```
brianp, line 11: What's the chance any of our plugins have css that was added, never worked, and was never removed. Of which will now be included and could do something weird?

MattBudz: I have gone through each of the addon's and engines. None of them have a tylium manifest (other than dradispro-word, which was already being imported), so this won't be a problem.

PR #893, opened 2021-09-02
Create issue for new evidence is no issues exist

MattBudz:
### Summary

Create an issue when creating new evidence is no issues exist

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 [v#.#.#] ([month] [YYYY])
   - Evidence:
-    - Create an issue for new evidence if no issues exist
+    - Create a new issue (optionally) when creating new evidence
   - Upgraded gems:
     - rails
   - Bugs fixes:
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 [v#.#.#] ([month] [YYYY])
-  - [entity]:
-    - [future tense verb] [feature]
+  - Evidence:
+    - Create an issue for new evidence is no issues exist
   - Upgraded gems:
     - rails
   - Bugs fixes:
--- a/app/controllers/evidence_controller.rb
+++ b/app/controllers/evidence_controller.rb
@@ -27,16 +27,6 @@ class EvidenceController < NestedNodeResourceController
   def create
     @evidence.author = current_user.email
 
-    unless evidence_params[:issue_id]
-      issue = Issue.create(
-        text: "#[Title]#\nNew issue",
-        node: current_project.issue_library,
-        author: current_user.email
-      )
-
-      @evidence.issue_id = issue.id
-    end
-
     respond_to do |format|
       if @evidence.save
         track_created(@evidence)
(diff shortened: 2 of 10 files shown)


CHANGELOG
```diff
@@ -1,6 +1,6 @@
 [v#.#.#] ([month] [YYYY])
-  - [entity]:
-    - [future tense verb] [feature]
+  - Evidence:
+    - Create an issue for new evidence is no issues exist
```
sean-yeoh, line 3:
```suggestion
    - Create an issue for new evidence if no issues exist
```
@MattBudz Small typo here

app/controllers/evidence_controller.rb
```diff
@@ -154,17 +154,18 @@ def set_or_initialize_evidence
...
-  # If the user selects "Add new issue" in the Evidence editor, we create an empty skeleton
-  def create_issue
-    Issue.create do |issue|
-      issue.text = "#[Title]#\nNew issue auto-created for node [#{@node.label}]."
-      issue.node = current_project.issue_library
-      issue.author = current_user.email
+  def evidence_params
+    evidence_params = params.require(:evidence).permit(:author, :content, :issue_id, :node_id)
+
+    if evidence_params[:issue_id].blank?
```
etdsoft, line 160: @sean-yeoh this behaviour is distinct from `evidence_params` and should be extracted to a different method. In fact, `evidence_params` shouldn't really concern itself with the :issue_id contents. You can do this in the action, or in a filter, but it doesn't seem to fit here.

In fact, I still think it belongs in a class method in Issue.

```
evidence.issue = Issue.autogenerate_from(evidence_params) unless evidence_params[:issue_id]
```

(or `.create_from`, etc.)

or whatever the right logic is.

etdsoft (changes requested):

app/controllers/evidence_controller.rb
```diff
@@ -140,6 +148,11 @@ def destroy
 
   private
 
+  def autogenerate_issue
+    @evidence.issue = Issue.autogenerate_from(@evidence)
```
etdsoft, line 152: you only need `@evidence` to pass the node: in `Issue.create`, which is just to tie the Issue to the project, and the controller already has a reference to the project.

So from a caller/controller perspective `autogenerate_from(evidence)` makes sense, but from a receiver/model perspective, not so much. It would make more sense to pass a reference to the project. Perhaps `autogenerate_for(current_project)`? Or maybe there's a way to go through `current_project` so we skip the call to the issues_library node, something like:

```
current_project.issues.autogenerate!
```
if that sets the right node_id:, then we're golden

app/controllers/evidence_controller.rb
```diff
@@ -140,6 +148,11 @@ def destroy
...
+    track_created(@evidence.issue)
```
etdsoft, line 153: 👏 good catch

app/controllers/evidence_controller.rb
sean-yeoh: We need the author too. We can also put it in `models/projects.rb` but it's weird.
```ruby
def autogenerate_issue_with_author(author)
  issues.create(
    author: author,
    node: issue_library
  )
end
```

PR #901, opened 2021-09-23
Fix DataTables resetting columns when columns are added or removed

sean-yeoh:
### Summary
This PR fixes datatable resetting its saved state when a new column is added or removed.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry


spec/support/datatables_shared_examples.rb
```diff
@@ -143,3 +141,67 @@
...
+      within '.dt-button-collection' do
+        click_link 'Created'
+        click_link 'Updated'
+      end
+    end
+  end
+
+  let(:update_resource_with_new_content) do
+    attributes = Hash[resource_attribute, new_content]
+    resource.update(attributes)
```
aapomm, line 163: Would something like this work instead:

```suggestion
    resource.update_attribute(resource_attribute, new_content)
```

spec/support/datatables_shared_examples.rb
```diff
@@ -143,3 +141,67 @@
     end
   end
 end
+
+# let(:new_content) { "#[Title]#\nTitle\n\n#[New Field]#\nNew Field Value" }
+# let(:old_content) { "#[Title]#\nTitle" }
+# let(:resource) { Issue.take }
+# let(:resource_attribute) { 'text' }
```
aapomm, line 148: I think `content_attribute` make more sense. Also, let's pass symbols instead.

```suggestion
# let(:resource_attribute) { :text }
```

app/assets/javascripts/shared/datatable/main.js
```diff
@@ -119,10 +119,18 @@ class DradisDatatable {
...
       },
       stateLoadCallback: function(settings) {
-        return JSON.parse(localStorage.getItem(that.localStorageKey));
+        if (localStorage.getItem(that.localStorageKey) === null) {
+          // We have to explicitly return null here because DataTables expects
+          // this function to return null or an object
+          return null;
+        }
+
+        var data = JSON.parse(localStorage.getItem(that.localStorageKey));
```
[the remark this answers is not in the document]
sean-yeoh: Gotcha, thanks

app/assets/javascripts/shared/datatable/main.js
```diff
@@ -119,10 +119,19 @@ class DradisDatatable {
...
       stateSaveCallback: function(settings, data) {
+        var data = that.saveHeadersInData(data);
         localStorage.setItem(that.localStorageKey, JSON.stringify(data));
       },
       stateLoadCallback: function(settings) {
-        return JSON.parse(localStorage.getItem(that.localStorageKey));
+        var localStorageData = JSON.parse(localStorage.getItem(that.localStorageKey));
+
+        if (localStorageData !== null) {
+          return that.rebuildLocalStorageColumns(localStorageData);
```
etdsoft, line 129:
```suggestion
          return that.rebuildColumnsFromLocalStorage(localStorageData);
```

app/assets/javascripts/shared/datatable/main.js
```diff
@@ -169,11 +178,47 @@ class DradisDatatable {
...
 
     document.addEventListener('turbolinks:before-cache', function() {
       that.dataTable.destroy();
     });
   }
+
+  // This method remove old columns from the data stored in localStorage
+  // (indirectly, because we are iterating over table headers on the page)
+  // and add new columns to it
+  rebuildLocalStorageColumns(localStorageData) {
```
etdsoft, line 203:
```suggestion
  rebuildColumnsFromLocalStorage(localStorageData) {
```

app/assets/javascripts/shared/datatable/main.js
```diff
@@ -169,11 +178,47 @@ class DradisDatatable {
...
+        return column.header == th.textContent;
+      })
+
+      if (column) {
+        newColumns.push(column);
+      } else {
+        // Add new column with default visibility = false
+        newColumns.push({ visible: false });
+      }
+    })
```
etdsoft, line 219: there's something about this loop within a loop to look for a single item that bothers me...

In pseudocode:
```js
var columnData = { visible: false };

localStorageData.columns.find(function(column) {
  if (column.header == th.textContent) {
    columnData = column;
    break;
  }
});
newColumns.push(columnData);
```

sean-yeoh: It's not looking for _just a_ single item.

It's looking for a column that matches the current iterator's text value, if found, we re-use the column stored in localStorage as it might contain the column's state.

Javascript's `Array.prototype.find()` returns a single value if found, and null otherwise, so we don't have to break it.

PR #903, opened 2021-10-05
Remove `category` selection from `Notes` form UI

MattBudz:
### Summary

Remove `category` selection from `Notes` form UI

### Check List

- [x] Added a CHANGELOG entry


app/views/notes/_form.html.erb
```diff
@@ -6,7 +6,7 @@
     }
   } do |f| %>
 
-  <%= f.association :category, collection: Category.all, label_method: :name, prompt: 'Assign note category' %>
+  <%= f.hidden_field :category, value: 1 %>
```
aapomm, line 9: Can we use the default category id here instead?
```suggestion
  <%= f.hidden_field :category, value: Category.default.id %>
```

PR #915, opened 2021-11-02
Validate evidence issue

sean-yeoh:
### Summary
This PR doesn't do much at all

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry


app/models/evidence.rb
```diff
@@ -20,6 +20,8 @@ class Evidence < ApplicationRecord
   validates :issue, presence: true, associated: true
   validates :node, presence: true, associated: true
 
+  validate :issue_cannot_be_on_another_project
```
brianp, line 23: I don't love the naming. How about something more positive

`validate :is_scoped_to_project`

app/models/evidence.rb
```diff
@@ -34,4 +36,14 @@ def local_fields
       'Title' => issue.try(:title)
     }
   end
+
+  private
+
+  def issue_cannot_be_on_another_project
+    return unless node && issue
+
+    if node.project.id != issue.project.id
```
brianp, line 45: we can just use `if project.id != issue.project.id` as evidence has a delegate method for project to the node.project.

PR #919, opened 2021-11-05
Add `View History` link to dots menu

MattBudz:
### Summary

Add `View History` link to dots menu

### Check List

- [x] Added a CHANGELOG entry


app/controllers/revisions_controller.rb
```diff
@@ -7,15 +7,18 @@ class RevisionsController < AuthenticatedController
   before_action :load_record, except: [ :trash, :recover ]
 
   def index
-    redirect_to action: :show, id: @record.versions.where(event: 'update').last.try(:id) || 0
+    redirect_to action: :show, id: @record.versions.last.try(:id) || 0
```
etdsoft, line 10: This changes, forces the `if` in L19, and when you're #show'ing a Revision that's not `update` then the `_empty` partial gets rendered giving the _false_ idea that there are no revisions for the record, there are (create, delete, restore), only that there are no _update_ revisions to show the diff.

I think the original behaviour was preferable, if you change the ID in the URL to some other random crap, you shouldn't get anywhere.

app/controllers/revisions_controller.rb
```diff
@@ -7,15 +7,18 @@ class RevisionsController < AuthenticatedController
...
   end
 
   def show
+    @record_class = @record.class.name.downcase
```
etdsoft, line 14: Is this actually used somewhere? Couldn't find it.

app/controllers/revisions_controller.rb
```diff
@@ -37,6 +40,11 @@ def recover
 
   private
 
+  def load_issues
```
etdsoft, line 43: This controller is working too hard as part of #show to render a custom sidebar depending on the reckon.

First I thought let's move the sidebar rendering across Tylium to an Ajax fetch post render. However for most controllers, that'd be overkill (perhaps not!).

But in this case, where the sidebar is potentially completely different in each controller, I think it does make sense.

Let the #show view retrieve the revision, and render the #show partial and have a fetch that in a separate request loads the sidebar contents.

app/views/evidence/_breadcrumbs.html.erb
```diff
@@ -1,6 +1,5 @@
 <ol class="breadcrumb">
   <li class="breadcrumb-item"><a href="javascript:void(0)" data-behavior="sidebar-toggle">Nodes</a></li>
-  <li class="breadcrumb-item" data-node-id="<%= @node.id %>"><%= link_to @node.label, main_app.project_node_path(@node.project, @node) %></li>
-  <li class="breadcrumb-item"><%= link_to 'Evidence', main_app.project_node_path(@node.project, @node, tab: 'evidence-tab') %></li>
+  <li class="breadcrumb-item" data-node-id="<%= @node.id %>"><%= link_to @node.label, main_app.project_node_path(@node.project, @node, tab: 'evidence-tab') %></li>
```
etdsoft, line 3: This PR does more than what it says in the tin, breadcrumb changes, including adding breadcrumbs to revisions is a different concern, split into a separate PR.

app/views/revisions/show.html.erb
```diff
@@ -1,23 +1,35 @@
 <% content_for :title, "#{@record.class.model_name} ##{@record.id} - comparing revisions" %>
 
-<% if @node %>
-  <%= render "nodes/sidebar" %>
+<% content_for :breadcrumbs do %>
+  <%= render partial: 'breadcrumbs' %>
+<% end %>
```
etdsoft, line 5: out of scope for this PR.

app/views/revisions/show.html.erb
```diff
@@ -1,23 +1,35 @@
...
+
+<% content_for :sidebar do %>
+  <% if @issue %>
+    <%= render 'issues/sidebar' %>
+  <% elsif @list %>
+    <%= render 'cards/sidebar' %>
+  <% elsif @node %>
+    <%= render 'nodes/sidebar' %>
+  <% end %>
 <% end %>
```
etdsoft, line 15: We can possibly turn this into an Ajax call, in a completely separate PR. I think the Ajax PR should probably be done/merged before the rest (breadcrumbs, and Activity tab => View History).

app/views/revisions/show.html.erb
```diff
@@ -1,23 +1,35 @@
...
   <div class="col-12 col-xl-8 col-xxl-9 p-0">
     <div class="content-container pb-3 mr-xl-3">
-      <div class="diff">
-        <h4 class="header-underline">
-          <%= @record.class.model_name %>
-          #<%= @record.id %> <span class="text-capitalize"> - view changes at <%= @diffed_revision.updated_at %></span>
-        </h4>
-        <pre class='js-diff-body'><%= @diffed_revision.diff.format_as(:color) %></pre>
-      </div>
+      <% if @diffed_revision %>
```
etdsoft, line 20: With this change, we've lost all of our front-end caching. We need to bring this back here, or we'll be taking a perf hit.

app/views/revisions/show.html.erb
```diff
@@ -1,23 +1,35 @@
...
+        </div>
+      <% else %>
+        <h4 class="header-underline">Revision Changes</h4>
+        <%= render 'empty' %>
+      <% end %>
     </div>
 
     <div id="<%= dom_id(@record) %>_revisions" class="content-container mr-xl-3 revisions-table note-text-inner">
-      <h4 class="header-underline">Revision History</h4>
+      <h4 class="header-underline"><%= @record.class.name %> History</h4>
```
etdsoft, line 32: Oh, perhaps this is where you intended to use that controller variable?

PR #929, opened 2021-11-18
Fix select all filtering

aapomm:
### Spec
Currently, the select all button does not behave intuitively with filtering. For example, if a user filters the table and selects all the records using the select all button, all the visible and filtered-out records will be selected. This means that a user may unknowingly delete a record.

**Proposed solution**
Update the datatable's checkbox behavior to handle a datatable that is in a filtered state.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


CHANGELOG
```diff
@@ -19,6 +19,7 @@
       - Ensure boards don't nest when the instance has been inactive
     - Tables
       - Prevent columns state from resetting
+      - Prevent the select all button from selecting the filtered rows
```
MattBudz, line 22:
```suggestion
      - Prevent the select all button from selecting filtered out rows when a filter is been applied
```

app/assets/javascripts/shared/datatable/checkbox.js
```diff
@@ -3,30 +3,49 @@ DradisDatatable.prototype.setupCheckboxListeners = function() {
...
+    that.updateSelectAllBtnState();
+  });
 }
 
-DradisDatatable.prototype.areAllSelected = function() {
+
+DradisDatatable.prototype.areAllSelected = function(filter = {}) {
   return(
-    this.dataTable.rows({selected:true}).count() == this.dataTable.rows().count()
+    this.dataTable.rows({selected: true}).count() == this.dataTable.rows(filter).count()
```
MattBudz, line 38:
```suggestion
    this.dataTable.rows({selected: true}).count() == this.dataTable.rows(filter).count();
```

PR #930, opened 2021-11-18
Show changes for issues evidence tab

MattBudz:
### Summary

- Display records in a DataTable
- Add multi-delete to the table

### Check List

- [x] Added a CHANGELOG entry


app/controllers/concerns/multiple_destroy.rb
```diff
@@ -11,7 +11,7 @@ def multiple_destroy
       job_params = {
         author_email: current_user.email,
         ids: params[:ids],
-        klass: params[:controller].singularize.capitalize,
+        klass: params[:controller].split('/')[-1].singularize.capitalize,
```
etdsoft, line 14: I thought we must have done this before, so I opened /controllers/concerns/ looking for instances, and found:

https://github.com/dradis/dradis-ce/blob/be46be1fff57c45c87a42a1956c3ded4191bf335/app/controllers/concerns/mentioned.rb#L15

so it seems that `params[:controller].split('/')[-1]` can be turned into [controller_name](https://api.rubyonrails.org/classes/ActionController/Metal.html#method-c-controller_name)?

app/controllers/issues/evidence_controller.rb
```diff
@@ -0,0 +1,4 @@
+class Issues::EvidenceController < AuthenticatedController
```
etdsoft, line 1: we discussed elsewhere that if we have an issues/evidence controller, now the rendering of the table does belong to this controller's #index as a :fetch post Issues#show is fully rendered.

That's a separate PR though, so you can make that first and the the data tables. It feels a better order than the reverse that would have you creating new code in a known-bad location to later move the code to the good location.

app/controllers/issues_controller.rb
```diff
@@ -22,6 +22,8 @@ def index
   def show
     @activities = @issue.commentable_activities.latest
 
+    @evidence_columns = ['Node'] | all_evidence_columns | ['Created by', 'Created',  'Updated']
```
etdsoft, line 25: This doesn't belong to the IssuesController any more.

app/controllers/issues_controller.rb
```diff
@@ -121,6 +123,13 @@ def import
   end
 
   private
+  def all_evidence_columns
+    @issue.evidence
+          .map { |evidence| evidence.fields.keys - ['Title', 'Label']  }
+          .flatten
+          .uniq
+  end
```
etdsoft, line 131: This is quacking like a setter (set_columns?) that needs to go into Issues::EvidenceController

From there we can also make the call of doing SQL eager loading to avoid N+1 in the query.

app/views/issues/_add_evidence.html.erb
```diff
@@ -1,4 +1,4 @@
-<div id="js-add-evidence-container">
+<div class="col-12 add-evidence-container" data-behavior="add-evidence-container">
```
etdsoft, line 1: 👏

spec/features/issues_spec.rb
```diff
@@ -323,112 +323,6 @@
           include_examples 'deleted item is listed in Trash', :issue
           include_examples 'recover deleted item', :issue
         end
-
-        describe 'add evidence', js: true do
```
etdsoft, line 327: We need an equivalent feature spec that confirms you can add Evidence from Issues#show.

There's quite a bit here actually: filtering of nodes, assignment of parent node, assignment of author, etc. We cannot just scrap all this desired behaviour from the test suite.

PR #931, opened 2021-11-25
Show changes issues evidence form

MattBudz:
### Summary

Move issues/_add_evidence form to isses/evidence/new

### Check List

- [x] Added a CHANGELOG entry


app/assets/javascripts/tylium/behaviors.js.coffee
```diff
@@ -193,3 +193,6 @@ document.addEventListener "turbolinks:load", ->
     history.pushState({}, '', this.href);
     $(this.hash)[0].scrollIntoView behavior: 'smooth'
     return
+
+  # Init Bootstrap tooltips
+  $('[data-toggle~=tooltip]').tooltip();
```
brianp, line 198: We're already using tooltips why wasn't this needed before?

app/controllers/issues/evidence_controller.rb
```diff
@@ -1,4 +1,15 @@
-class Issues::EvidenceController < AuthenticatedController
-  include MultipleDestroy
-  include ProjectScoped
+class Issues::EvidenceController < IssuesController
+  def new
+    @auto_save_key =  if params[:template]
+                        "project-#{current_project.id}-issue-#{params[:issue_id]}-evidence-#{params[:template]}"
+                      else
+                        "project-#{current_project.id}-issue-#{params[:issue_id]}-evidence"
+                      end
```
brianp, line 7: This would be better suited for a before filter. You don't even need to assign the filter itself. If the method name is `set_auto_save_key` it will automatically run because IssuesController calls the before filter.

app/views/issues/_sidebar.html.erb
```diff
@@ -9,12 +9,12 @@
 
 <%= render 'shared/sidebar_collection',
   category_id: nil,
-  collection: @issues,
+  collection: @issues.sort,
```
brianp, line 12: Is this an intended behaviour change?

spec/features/issues_spec.rb
```diff
@@ -274,7 +274,7 @@
 
           it 'presents the list of hosts affected by a given issue'  do
             expect(find('.secondary-sidebar-content')).to have_content('10.0.0.1')
-            expect(find('.secondary-sidebar-content')).to have_content('10.0.0.2', count: 5)
+            expect(find('.secondary-sidebar-content')).to have_content('10.0.0.2', count: 3)
```
brianp, line 277: But why did the count change if nothing else changed? 🧐

app/views/layouts/tylium/_sidebar.html.erb
```diff
@@ -8,7 +8,7 @@
         <i class="fa fa-tachometer sidebar-link-icon"></i> <span class="sidebar-link-label">Dashboard</span>
       <% end %>
     <% end %>
-    <%= content_tag :li, :class => controller_path == 'issues' || controller_path == 'issues/merge' ? 'active' : '' do %>
+    <%= content_tag :li, :class => controller_path.include?('issues') ? 'active' : '' do %>
```
brianp, line 11: Why isn't the issues/merge path relevant anymore?

app/assets/javascripts/tylium/behaviors.js.coffee
MattBudz: It used to live in `app/assets/javascripts/tylium/modules/issues/evidence.js.coffee` and was moved here since it shouldn't be only for `issues#show`.

app/views/issues/_sidebar.html.erb
MattBudz: Without this, I noticed the issues in the sidebar were not rendering in the same order as `issues#show`.

spec/features/issues_spec.rb
MattBudz: This actually brings it back to the way it is in `main`. This was a temporary change in the base branch to pass specs before the changes from this PR are merged in.

app/views/layouts/tylium/_sidebar.html.erb
MattBudz: It's not that it's no longer relevant. ` controller_path.include?('issues')` picks up `issues` in `issues/merge`. 

This change was done to get the correct nav-item to be active for all issues related views without adding a `... || controller_path == 'issues/evidence' || controller_path == 'issues/whatever_else_in_the_future'`

PR #933, opened 2021-11-29
Show changes dots menu

MattBudz:
### Summary

This PR moves the `...` (more actions) menu closer to the content affected by the actions of the menu

### Check List

- [x] Added a CHANGELOG entry


app/assets/stylesheets/tylium/modules/_dots-menu.scss
```diff
@@ -1,55 +1,48 @@
...
+    font-weight: 400 !important;
+    
+    &:hover {
+      color: $linkColor !important;
+      background-color: initial !important;
+    }
 
-  .breadcrumb {
-    width: calc(100% - 35px);
+    &:not(:first-child) {
```
sean-yeoh, line 21: What's this change for?

app/views/cards/show.html.erb
```diff
@@ -48,7 +27,28 @@
           <div class="tab-pane active" id="info-tab">
             <% cache ['card-information-tab', @card] do %>
               <div class="note-text-inner">
-                <h4 class="mb-4 header-underline"><%= @card.name %></h4>
+                <h4 class="mb-4 header-underline">
+                  <span class="text-truncate" title="<%= @card.name %>"><%= @card.name %></span>
```
sean-yeoh, line 31: Is `text-truncate` a new change? Why are we truncating instead of using `word-wrap: break-word;` if it's overflowing?

CHANGELOG
```diff
@@ -1,6 +1,6 @@
 [v#.#.#] ([month] [YYYY])
-  - [entity]:
-    - [future tense verb] [feature]
+  - Tylium:
+    - Move '...' (more actions) menu closer to the content affected by the actions of the menu
```
sean-yeoh, line 3:
```suggestion
    - Move the position of dots menu
```

I think the language here is a bit confusing, can you simplify it?

app/assets/stylesheets/tylium/modules/_dots-menu.scss
MattBudz: This gives header items in the dropdown menu some extra styling if the header is not at the top of the menu (we don't want double borders)

app/views/cards/show.html.erb
MattBudz: we are limiting the header to a single line to have consistent headers with the dots menu always in the same position. If we wrap the words instead, the header can be infinitely tall creating an ugly view.

MattBudz: Also, if the text is truncated, the user can still see the full version in a tooltip when they hover over the header.

PR #934, opened 2021-11-30
Specs - use a fixture for shared pwd management

etdsoft:
### Summary

Instead of calling `Configuration.create(name: 'admin:password', value: 'rspec_pass')` in a bunch of specs, just extract that to a fixture.

### Check List

- [x] Added a CHANGELOG entry

More of an internal refactor than a user-facing change.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -24,7 +24,7 @@ Rails.application.routes.draw do
 
   resources :notifications, only: [:index, :update]
 
-  resources :projects, only: [:show] do
+  resources :projects, only: [:index, :show] do
     resources :activities, only: [:index] do
       collection do
         get :poll, constraints: { format: /js/ }
--- a/spec/features/sessions_spec.rb
+++ b/spec/features/sessions_spec.rb
@@ -3,13 +3,8 @@ require 'rails_helper'
 describe 'Sessions' do
   subject { page }
 
-  before do
+  let(:user) do
     create(
-      :configuration,
-      name: 'admin:password',
-      value: ::BCrypt::Password.create('rspec_pass')
-    )
-    @user = create(
       :user,
       :author,
       password_hash: ::BCrypt::Password.create('rspec_pass')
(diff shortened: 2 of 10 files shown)


config/routes.rb
```diff
@@ -24,7 +24,7 @@
 
   resources :notifications, only: [:index, :update]
 
-  resources :projects, only: [:show] do
+  resources :projects, only: [:index, :show] do
```
brianp, line 27: What's this for?

spec/features/sessions_spec.rb
```diff
@@ -3,24 +3,20 @@
...
       password_hash: ::BCrypt::Password.create('rspec_pass')
     )
   end
 
-  let(:password) { 'rspec_pass' }
-
-  let(:login) do
+  # This needs to be a helper and not a let() block, because let is memoized
+  # and reused.
+  def login
```
brianp, line 17: You're not wrong but maybe memoized was the intent. 

Running this code is costly, and redundant if it's done before hundreds of other specs. If we want a login helper I'd go the route of faking the session as logged in instead of performing log in actions.

Unless you can firmly say this has no negative affect on the total run time I'd say I'm strongly against.

brianp (changes requested):

config/routes.rb
etdsoft: just for Pro parity down here:

https://github.com/dradis/dradis-ce/pull/934/files/b81de648c088ab4b371ade481eae126e9fa70831#diff-e5e534608b42f749939e20a14f9379d20abb07e787c1533fcac6cdfa15a16068R61

Turns out Projects#index was already implemented in CE:

https://github.com/dradis/dradis-ce/blob/main/app/controllers/projects_controller.rb#L11

spec/features/sessions_spec.rb
etdsoft: Agreed on all grounds, but again pro-parity (we need multiple users there, and can't be memoized).

I thought about using [`login_as_user` from ControllerMacros](https://github.com/dradis/dradis-ce/blob/b81de648c088ab4b371ade481eae126e9fa70831/spec/support/controller_macros.rb#L7), but then I realised this is a feature spec, so it's prolly the one place where filling in the form makes sense.

PR #937, opened 2021-12-02
CleanOpenVASSettings migration naming issue

etdsoft:
### Summary

Before this PR:

if OpenVAS is loaded (which includes the inflector), it works. if it’s not, it
fails with:

    NameError: uninitialized constant CleanOpenvasSettings
    Did you mean?  CleanOpenVASSettings

However, if you change it to`CleanOpenvasSettings`, then:

if OpenVAS is loaded, it fails.

    NameError: uninitialized constant CleanOpenVASSettings
    Did you mean?  CleanOpenvasSettings

if OpenVAS isn’t loaded, it works.

### Check List
- [x] Added a CHANGELOG entry

This is an internal refactor that isn't user-facing or requiring a CHANGELOG entry.

### DIFF
--- a/db/migrate/20131002215400_clean_openvas_settings.rb
+++ b/db/migrate/20131002215400_clean_openvas_settings.rb
@@ -1,4 +1,4 @@
-class CleanOpenVASSettings < ActiveRecord::Migration[5.1]
+class CleanLegacyConfig < ActiveRecord::Migration[5.1]
   def up
     # This removes old settings that may exist in versions of Pro prior to v1.9
     %w{openvas:node_label}.each do |name|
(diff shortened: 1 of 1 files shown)


db/migrate/20131002215400_clean_openvas_settings.rb
```diff
@@ -1,4 +1,4 @@
-class CleanOpenVASSettings < ActiveRecord::Migration[5.1]
+class CleanLegacyConfig < ActiveRecord::Migration[5.1]
```
sean-yeoh, line 1: why not change the file name to `db/migrate/20131002215400_clean_legacy_config.rb` if we're already changing the class name?

etdsoft: Mainly because it doesn't describe the problem enough. With this trickery the filename remains descriptive, and Rails find the right class through the trickery in L14-15

PR #938, opened 2021-12-09
Have dradis:setup:welcome use dradis:setup:kit internally

etdsoft:
### Summary

Before this PR the `dradis:setup:welcome` task would manually copy templates and initialize project content. This PR removes that duplication and relies on `dradis:setup:kit` to do the heavy lifting.

### Check List

- [x] Added a CHANGELOG entry

Internal refactor, no user-facing changes.

### DIFF
--- a/lib/tasks/thor/setup.rb
+++ b/lib/tasks/thor/setup.rb
@@ -70,41 +70,56 @@ class DradisTasks < Thor
 
     desc 'welcome', 'adds initial content to the repo for demonstration purposes'
     def welcome
-      # TODO: once dradis:setup:kit is available, replace this entire task with
-      # two steps:
-      #   1. prepare Welcome kit from template files
-      #   2. invoke 'dradis:setup:kit', [], [welcome_kit.path]
+      kit_file = prepare_kit
 
-      # --------------------------------------------------------- Note template
-      if NoteTemplate.pwd.exist?
-        say 'Note templates folder already exists. Skipping.'
-      else
-        template 'note.txt', NoteTemplate.pwd.join('basic_fields.txt')
+      # Before we import the Kit we need at least 1 user
+      User.create!(email: 'adama@dradisframework.com').id
+      invoke 'dradis:setup:kit', [], file: kit_file.path
+    end
+
+    private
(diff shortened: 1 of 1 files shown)


lib/tasks/thor/setup.rb
```diff
@@ -70,41 +70,56 @@ def kit
...
-      if NoteTemplate.pwd.exist?
-        say 'Note templates folder already exists. Skipping.'
-      else
-        template 'note.txt', NoteTemplate.pwd.join('basic_fields.txt')
-      end
+      kit_file = prepare_kit
+
+      # Before we import the Kit we need at least 1 user
+      User.create!(email: 'adama@dradisframework.com').id
+      invoke 'dradis:setup:kit', [], file: kit_file.path
```
brianp, line 77: I think this is the only yellow flag I see. Overall I like the idea of what's happening. It makes sense, it's an improvement. The flag drop comes when a task invokes another task. I'm not sure if this is standard in Thor but I know it's considered bad form in rake. Alternatively most tasks would be backed by a class or service object that performs the work which also makes testing easier. The task would just be an entry point to invoking the service object. 
If that were the case then instead of invoking another task, this task/service class could call the additional needed classes directly. Heck if one of them knew simply how to operate on the file structure we could end up skipping the zip completely when invoked from a task. Just point it at the template files and send it to work.

That said, this is still an improvement. So I wouldn't hold up the change for the more perfect edition.

etdsoft: Not sure if you saw the other Thor task? [It just invokes the service object](https://github.com/dradis/dradis-ce/blob/b098d1f89d0679fe9abb30db99112b81553bbd31/lib/tasks/thor/setup.rb#L65-L69) (well, in this case is a job:

```ruby
      KitImportJob.perform_now(file: options[:file], logger: default_logger)
```

My thinking was: if what we're doing is enqueing that job, and the `dradis:setup:kit` task already knows how to do it, let's invoke it instead of duplicating the job setup/enqueing in case that has to change so we don't have to keep track of two pieces of identical code...

catch 22?

PR #942, opened 2021-12-16
Add Deploy to Digital Ocean button

etdsoft:
### Summary

This PR adds a Deploy to DO button to the repo. Hopefully this will make it easier for teams wanting to check CE out to spin up a server and give it a go.

It's still work in progress as there are some decisions we need to make, mainly in terms of `Gemfile.plugins` and `bin/setup` changes.

### Check List

- [ ] Added a CHANGELOG entry


config/environments/production.rb
```diff
@@ -24,7 +24,8 @@
   config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
   # Compress JavaScripts and CSS.
-  config.assets.js_compressor = :uglifier
+  # config.assets.js_compressor = :uglifier
+  config.assets.js_compressor = Uglifier.new(harmony: true)
```
brianp, line 28: What's this about?

etdsoft: Some sort of JS transpiler issue without which deployment fails in managed environments (I thought that I could get away without it after I got past the `yarn` hurdle, but tested it this morning and failed, so it has to stay):
Screenshot 2021-12-17 at 13 11 12

```
Uglifier::Error: Unexpected token: name (LocalAutoSave). To use ES6 syntax, harmony mode must be enabled with Uglifier.new(:harmony => true).
```

etdsoft: TIL, we change this in Pro 3 years go. So this change brings back parity.

Gemfile.plugins.template
```diff
@@ -5,34 +5,12 @@
 #   $ bundle
 
 # ----------------------------------------------------------------- Calculators
-gem 'dradis-calculator_cvss', '~> 4.1.0'
```
sean-yeoh, line 8: Why are the plugins being removed from the template?

PR #955, opened 2022-01-10
Show changes issues evidence datatable

MattBudz:
### Summary

Display an issue's associated evidence records in a DataTable.

### Check List

- [x] Added a CHANGELOG entry


app/controllers/issues_controller.rb
```diff
@@ -121,6 +123,13 @@ def import
   end
 
   private
+  def all_evidence_columns
+    @issue.evidence
+          .map { |evidence| evidence.fields.keys - ['Title', 'Label']  }
```
etdsoft, line 128: can you move the `- []` part of the operation out of the loop? That way instead of it being applied to each iteration we just apply it once, after `.uniq`.

app/views/issues/show.html.erb
```diff
@@ -35,8 +35,8 @@
 <!-- Main content -->
 <div class="row m-0">
   <ul class="tabs-container nav nav-tabs main-tabs w-100">
-    <li class="nav-item"><a href="#info-tab" data-toggle="tab" class="nav-link active"><%= colored_icon_for_model(@issue, 'fa-bug') %> Information</a></li>
-    <li class="nav-item"><a href="#evidence-tab" class="nav-link" data-toggle="tab"><i class="fa fa-laptop"></i> Evidence <span class="badge bg-primary text-white"><%= @issue.evidence.count %></span></a></li>
+    <li class="nav-item"><a href="#info-tab" data-toggle="tab" class="nav-link <%= class_names(active: !params[:tab]) %>"><%= colored_icon_for_model(@issue, 'fa-bug') %> Information</a></li>
```
etdsoft, line 38: I'm not sure any of the references to `params[:tab]` belong to this PR.

Afaict, it's only used in this view, and only to render HTML, can you please remove?

spec/features/issues_spec.rb
```diff
@@ -323,112 +323,6 @@
           include_examples 'deleted item is listed in Trash', :issue
           include_examples 'recover deleted item', :issue
         end
-
-        describe 'add evidence', js: true do
```
etdsoft, line 327: you're removing significant amounts of test coverage without replacing it with other coverage.

Are these specs already elsewhere?

CHANGELOG
```diff
@@ -1,3 +1,31 @@
+[v#.#.#] ([month] [YYYY])
+  - Issues:
+    - Display evidence in a table
```
etdsoft, line 3: Single item lists can be compressed into 1 line for the time being.

app/views/issues/show.html.erb
MattBudz: This is setting us up for the bulk delete pr and adding new evidence pr, so that those actions return the user to the evidence tab.

spec/features/issues_spec.rb
MattBudz: The specs are bring migrated/updated in the add evidence PR. New specs are also being added in the bulk delete PR. Both PR's will be merged into this one so we will not be without test coverage once this is ready to merge to develop

CHANGELOG
MattBudz: There are 2 other PR's that will be merging into this one with their own entires into the changelog and this makes the merge smoother.

app/controllers/issues_controller.rb
MattBudz: makes sense, pushed the change.

PR #956, opened 2022-01-10
Show changes issues evidence bulk delete

MattBudz:
### Summary

add bulk-delete to issue level evidence table

### Check List

- [x] Added a CHANGELOG entry


app/controllers/issues/evidence_controller.rb
```diff
@@ -0,0 +1,2 @@
+class Issues::EvidenceController < IssuesController
+end
```
brianp, line 2: What actions actually get used via this inheritance. The bulk destroy only?

Doesn't seem like we really need the inheritance if we can include a single module in a new controller. Less surface area, even if the rest isn't exposed _it's there_.

MattBudz: We are using Bulk Delete, Project Scoped, and a couple of methods. This will all come together when we merge all the PR's down to the feature branch. There is a lot more to this controller, but for this PR this is all we have.

MattBudz: We will be (testing) inheriting from AuthenticatedController after merging to the feature branch

PR #957, opened 2022-01-11
load issue evidence tab content with ajax

MattBudz:
### Summary

Clean up and improve the performance of `issues#show` by loading the evidence tab content with ajax

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/controllers/issues/evidence_controller.rb
+++ b/app/controllers/issues/evidence_controller.rb
@@ -1,5 +1,5 @@
 class Issues::EvidenceController < IssuesController
-  before_action :set_issues, only: [:index]
+  before_action :set_issue, only: [:index]
   before_action :set_affected_nodes, only: [:index]
 
   def index
@@ -17,8 +17,7 @@ class Issues::EvidenceController < IssuesController
           .uniq - ['Title', 'Label']
   end
 
-  def set_issues
-    @issues = Issue.where(node_id: current_project.issue_library.id)
-    @issue = @issues.find(params[:issue_id])
+  def set_issue
+    @issue = Issue.where(node_id: current_project.issue_library.id).find(params[:issue_id])
   end
 end
--- a/app/controllers/issues/evidence_controller.rb
+++ b/app/controllers/issues/evidence_controller.rb
@@ -1,0 +1,24 @@
+class Issues::EvidenceController < IssuesController
+  before_action :set_issues, only: [:index]
+  before_action :set_affected_nodes, only: [:index]
+
+  def index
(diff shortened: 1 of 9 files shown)


app/controllers/issues/evidence_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def all_evidence_columns
+    @issue.evidence
+          .map { |evidence| evidence.fields.keys }
+          .flatten
+          .uniq - ['Title', 'Label']
+  end
+
+  def set_issues
+    @issues = Issue.where(node_id: current_project.issue_library.id)
```
etdsoft, line 21: Why do you need to load this collection?

app/views/issues/_evidence.html.erb
```diff
@@ -1,24 +1,8 @@
...
-  </div>
-<% end %>
+<%= content_tag :div,
+  spinner_tag,
+  data: {
+    behavior: 'fetch',
+    path: project_issue_evidence_index_path(issue_id: @issue.id)
+  }
+%>
+
```
etdsoft, line 8: Do we need a partial any more? It seems that this chunk can go into show.html.erb 

Aesthetically, once there, I think I'd move the `content_tag` to a new line nested under `<%=`

app/views/issues/evidence/index.html.erb
```diff
@@ -0,0 +1,23 @@
+<h4 class="header-underline">Evidence information</h4>
+<% if @issue.affected.any? %>
+  <% cache ['issue-evidence-tab', @issue, @issue.evidence] do %>
+    <%= render 'issues/evidence/table' %>
+  <% end %>
+<% else %>
```
etdsoft, line 6: this is inconsistent (I believe) with Nodes#show>Evidence:empty

https://github.com/dradis/dradis-ce/blob/develop/app/views/nodes/show/_evidence.html.erb

app/views/layouts/tylium.html.erb
```diff
@@ -82,7 +82,7 @@
       <%= render 'boards/modals/new_board_form' unless request.xhr? %>
     <% end %>
 
-    <% if controller_name == 'evidence'%>
+    <% if controller_path == 'evidence' %>
```
etdsoft, line 85: 👏

app/controllers/issues/evidence_controller.rb
MattBudz: good question, removed.

app/views/issues/evidence/index.html.erb
MattBudz: updated

PR #959, opened 2022-01-18
Remove ../../shared/ from the code

etdsoft:
### Summary

Before this change, several parts of the code (mainly around the Configuration class) would assume that a "../../shared" folder existed. While this is true in Pro / production, it would cause initialization issues in CE or on a fresh clone, if the user didn't have access to write on ../../ (e.g. if you're running from `/home/etd/dradis-ce/` or even `/tmp/dradis-ce`).

This PR removes the `../../shared/` hard-coding and replaces it with the creation of a `./templates` folder by `./bin/setup`

### Check List

- [x] Added a CHANGELOG entry


app/models/configuration.rb
```diff
@@ -41,28 +41,35 @@ def self.signups_enabled?
...
 
   def self.paths_templates_projects
-    create_with(value: Rails.root.join('../../shared/templates/projects/').to_s)
+    create_with(value: paths_templates.join('projects').to_s)
       .find_or_create_by(name: 'admin:paths:templates:projects').value
   end
 
   def self.paths_templates_reports
-    create_with(value: Rails.root.join('templates', 'reports'))
+    create_with(value: paths_templates.join('reports'))
```
aapomm, line 72: This is missing the `to_s` call but it doesn't seem to be needed when I test it in the console. I suggest we skip it entirely but this needs to be consistent with the rest.

app/models/concerns/file_backed_model.rb
```diff
@@ -51,16 +51,11 @@ def from_file(filename)
     # Returns a Pathname to location configured in the database through the
     # :setting_name setting
```
aapomm, line 52: Due to the refactor, this comment is not relevant anymore and can be removed

app/models/configuration.rb
etdsoft: Addressed, you could have saved me a commit w/ a suggestion ;-)

app/models/concerns/file_backed_model.rb
etdsoft: fixed.

PR #960, opened 2022-01-18
Add breadcrumbs to Revision History view

MattBudz:
### Summary

This PR adds breadcrumbs to the Revision History view to improve navigation.

### Check List

- [x] Added a CHANGELOG entry


app/controllers/revisions_controller.rb
```diff
@@ -11,6 +11,7 @@ def index
   end
 
   def show
+    @record_class = @record.class.name.downcase
```
aapomm, line 14:
```suggestion
    @class_name = @record.class.name.downcase
```
For naming consistency with: https://github.com/dradis/dradis-ce/blob/develop/app/views/nodes/items_table/_table.html.erb#L1

app/views/revisions/_breadcrumbs.html.erb
```diff
@@ -0,0 +1,21 @@
+<nav>
+  <ol class="breadcrumb">
+    <% if @record_class == 'card' %>
```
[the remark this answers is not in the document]
MattBudz: I think we definitely could but we should discuss this on BC first so we don't corner ourselves with flexibility, kind of how we did at first with the shared empty state partial. 

I'm going to close this as out of scope but it's a convo worth having.

app/controllers/revisions_controller.rb
```diff
@@ -11,6 +11,7 @@ def index
   end
 
   def show
+    @class_name = @record.class.name.downcase
```
etdsoft, line 14: it seems to me you don't need this variable.

You can replace the view "if else if else" with a case structure:

```ruby
<% case @record %>
<% when Card %>
<% when Issue %>
<% end %>

File: app/views/revisions/_breadcrumbs.html.erb [hunk 4]
```diff
@@ -0,0 +1,21 @@
+<nav>
+  <ol class="breadcrumb">
+    <% if @class_name == 'card' %>
+      <li class="breadcrumb-item"><%= link_to 'Methodologies', project_boards_path(current_project) %></li>
+      <li class="breadcrumb-item"><%= link_to "#{@record.board.name} - #{@record.list.name}", project_board_path(current_project, @record.board) %></li>
+      <li class="breadcrumb-item"><%= link_to @record.name, project_board_list_card_path(current_project, @board, @list, @record) %></li>
+    <% elsif @class_name == 'evidence' %>
+      <li class="breadcrumb-item"><a href="javascript:void(0)" data-behavior="sidebar-toggle">Nodes</a></li>
+      <li class="breadcrumb-item"><%= link_to @record.node.label, project_node_path(current_project, @record.node, tab: "#{@class_name.pluralize(2)}-tab") %></li>
```
etdsoft, line 9: no value in interpolating this variable, you're already in the Evidence section, so hard-code: `tab: 'evidence-tab'`

PR #962, opened 2022-01-20
Cache Column Title keys

brianp:
### Summary

As records go up in quantity the dynamic key generation takes it's toll.

Any record type that allows the use of Dradis Fields is inherently
dynamic. This makes drawing the table complicated because we need to
know all possible values for table headers. Focusing on issues as an
example: Currently we load every issue into memory examine the fields
and build the list of known columns.

I ran some bench marking on this query and the amount of records that
need to be scanned are positively correlated to the amount of time it
takes to process the query.

We do lots of caching on the page already but those very cache keys use
this expensive query as their key. Caching won't help if the key is the
problem.

This commit CACHES the expensive query used to later make the keys. I
heard you like caches so I cached your cache key.

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 [v#.#.#] ([month] [YYYY])
   - Add Setup Wizard
   - Editor: Support fields with the same name in the Fields View
+  - Inceased page loading performance on Issues, Evidence, and Notes for
+    projects with *a lot* of issues, evidence, or notes
   - Tylium:
     - Import CSS manifests from addons
     - Move '...' (more actions) menu closer to the content affected by the actions of the menu
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,7 @@
 [v#.#.#] ([month] [YYYY])
   - Add Setup Wizard
   - Editor: Support fields with the same name in the Fields View
-  - Inceased page loading performance on Issues, Evidence, and Notes for
+  - Increased table loading performance on Issues, Evidence, and Notes for
     projects with *a lot* of issues, evidence, or notes
   - Tylium:
     - Import CSS manifests from addons
--- a/app/controllers/concerns/cached_columns.rb
+++ b/app/controllers/concerns/cached_columns.rb
@@ -2,7 +2,7 @@ module CachedColumns
   CACHE_KEY = "%{identifier}/%{record_type}/column-keys/%{tail}".freeze
 
   # Takes an ActiveRecord::Relation so we can make one more query off it
-  def cached_collection_column_keys(collection, extra_columns, identifier: @current_project.id)
+  def cached_collection_column_keys(collection, extra_columns, identifier = "projects-#{@current_project.id}")
     last_updated_record = collection.order(updated_at: :desc).first
     # Exit early if the collection is empty.
     return extra_columns unless last_updated_record
--- a/app/controllers/concerns/cached_columns.rb
+++ b/app/controllers/concerns/cached_columns.rb
@@ -1,0 +1,18 @@
+module CachedColumns
+  CACHE_KEY = "projects/%{project_id}/%{record_type}/column-keys/%{tail}".freeze
+
+  # Takes an ActiveRecord::Relation so we can make one more query off it
+  def cached_collection_column_keys(collection, extra_columns)
+    last_updated_record = collection.order(updated_at: :desc).first
+    last_updated = last_updated_record.updated_at.to_i
+
+    key_opts = { project_id: current_project.id, record_type: last_updated_record.class.name }
+
+    Rails.cache.fetch(CACHE_KEY % key_opts.merge(tail: last_updated)) do
+      # Delete any old key before we write the new one
+      Rails.cache.delete_matched(CACHE_KEY % key_opts.merge(tail: '*'))
+
+      collection.map(&:fields).map(&:keys).uniq.flatten | extra_columns
+    end
+  end
+end
--- a/app/controllers/concerns/cached_columns.rb
+++ b/app/controllers/concerns/cached_columns.rb
@@ -2,10 +2,10 @@ module CachedColumns
   CACHE_KEY = "%{identifier}/%{record_type}/column-keys/%{tail}".freeze
 
   # Takes an ActiveRecord::Relation so we can make one more query off it
-  def cached_collection_column_keys(collection, extra_columns, identifier = "projects-#{@current_project.id}")
+  def cached_collection_column_keys(collection, identifier = "projects-#{@current_project.id}")
     last_updated_record = collection.order(updated_at: :desc).first
     # Exit early if the collection is empty.
-    return extra_columns unless last_updated_record
+    return [] unless last_updated_record
 
     last_updated = last_updated_record.updated_at.to_i
 
@@ -15,7 +15,7 @@ module CachedColumns
       # Delete any old key before we write the new one
       Rails.cache.delete_matched(CACHE_KEY % key_opts.merge(tail: '*'))
 
-      collection.map(&:fields).map(&:keys).uniq.flatten | extra_columns
+      collection.map(&:fields).map(&:keys).uniq.flatten
     end
   end
 end
--- a/app/controllers/concerns/cached_columns.rb
+++ b/app/controllers/concerns/cached_columns.rb
@@ -6,6 +6,9 @@ module CachedColumns
     last_updated_record = collection.order(updated_at: :desc).first
     last_updated = last_updated_record.updated_at.to_i
 
+    # DradisPro IssueLibrary support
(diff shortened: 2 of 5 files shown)


CHANGELOG
```diff
@@ -1,6 +1,8 @@
 [v#.#.#] ([month] [YYYY])
   - Add Setup Wizard
   - Editor: Support fields with the same name in the Fields View
+  - Inceased page loading performance on Issues, Evidence, and Notes for
```
etdsoft, line 4:
```suggestion
  - Increased table loading performance on Issues, Evidence, and Notes for
```

app/controllers/concerns/cached_columns.rb
```diff
@@ -0,0 +1,21 @@
+module CachedColumns
```
etdsoft, line 1: I hear you, but don't love it. I think "Field" something somewhere should be part of this concern. Right now "CachedColumns" could mean a lot of things.

Is it worth adding a description as well?

Thinking about the behaviour / concern, it seems this provides some sort of functionality for actions with record tables?

app/controllers/issues_controller.rb
```diff
@@ -15,9 +16,10 @@ class IssuesController < AuthenticatedController
   before_action :set_or_initialize_tags, except: [:destroy]
   before_action :set_auto_save_key, only: [:new, :create, :edit, :update]
 
+  EXTRA_COLUMNS = ['Title', 'Tags', 'Affected', 'Created', 'Created by', 'Updated'].freeze
+
   def index
-    @columns = @issues.map(&:fields).map(&:keys).uniq.flatten |
-      ['Title', 'Tags', 'Affected', 'Created', 'Created by', 'Updated']
+    @columns = cached_collection_column_keys(@unsorted_issues) | EXTRA_COLUMNS
```
etdsoft, line 22: I get that the first time around you don't have anything, but cache hitting / building every single time you go back to the page seems a bit much?

Should be also do the calculation post #update or #create to try to minimise the need to build on #index?

app/controllers/concerns/cached_columns.rb
```diff
@@ -0,0 +1,21 @@
...
+  CACHE_KEY = "%{identifier}/%{record_type}/column-keys/%{tail}".freeze
+
+  # Takes an ActiveRecord::Relation so we can make one more query off it
+  def cached_collection_column_keys(collection, identifier = "projects-#{@current_project.id}")
```
etdsoft, line 5: This method name is working too hard, and still not being clear enough about what to expect from it. It doesn't return a key, it does work on a collection, but are columns really something we deal with?

app/controllers/issues_controller.rb
```diff
@@ -1,5 +1,6 @@
 class IssuesController < AuthenticatedController
   include ActivityTracking
+  include CachedColumns
```
etdsoft, line 3: should the concern be smarter? and hook on after_actions perhaps for #update or #create?

app/controllers/nodes_controller.rb
```diff
@@ -1,18 +1,20 @@
 # This controller exposes the REST operations required to manage the Node
 # resource.
 class NodesController < NestedNodeResourceController
+  include CachedColumns
```
etdsoft, line 4: similar concerns to IssueController

app/controllers/issues_controller.rb
brianp: > but cache hitting every single time you go back to the page seems a bit much?

Isn't that the definition and purpose of caching? We look for a cache hit and act with it. That's pretty SOP.

>  building every single time you go back

Only gets built on cache miss. So still winning.

> Should be also do the calculation post #update or #create to try to minimise the need to build on #index?
Answer on pacifica.

app/controllers/issues_controller.rb
brianp: There's risk involved with that. Comment coming in Pacifica.

app/controllers/concerns/cached_columns.rb
brianp: `cached_collection_column_keys` -> `collection_field_names`

app/controllers/concerns/cached_columns.rb
brianp: `CachedColumns` -> `DataTableFieldHeaderCalculatorExtrodinare`

i kid...

`CachedColumns` -> `DynamicFieldNamesCacher`

CHANGELOG
```diff
@@ -1,6 +1,8 @@
 [v#.#.#] ([month] [YYYY])
   - Add Setup Wizard
   - Editor: Support fields with the same name in the Fields View
+  - Increased table loading performance on Issues, Evidence, and Notes for
+    projects with *a lot* of issues, evidence, or notes
```
etdsoft, line 5: why here and not under the Tylium list (L6)?

app/controllers/concerns/dynamic_field_names_cacher.rb
```diff
@@ -0,0 +1,21 @@
+module DynamicFieldTitlesCacher
+  CACHE_KEY = "%{identifier}/%{record_type}/column-keys/%{tail}".freeze
+
+  # Takes an ActiveRecord::Relation so we can make one more query off it
+  def collection_field_names(collection, identifier = "projects-#{@current_project.id}")
```
etdsoft, line 5: I don't know about working so hard as part of the argument? what about `identifier=nil` and then later `identifier ||= "projects-" to keep the signature a bit cleaner?

app/controllers/concerns/dynamic_field_names_cacher.rb
```diff
@@ -0,0 +1,21 @@
...
+    last_updated_record = collection.order(updated_at: :desc).first
+    # Exit early if the collection is empty.
+    return [] unless last_updated_record
```
etdsoft, line 8: I'd make this L1 of the method, and replace with `collection.empty?` to make it clear it has nothing to do with the :last_updated value. Presumably the `collection` has already been loaded, so it won't add an extra SQL to just check for empty? (or any?)

app/controllers/concerns/dynamic_field_names_cacher.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    last_updated = last_updated_record.updated_at.to_i
+
+    key_opts = { identifier: identifier, record_type: last_updated_record.class.name }
+
+    Rails.cache.fetch(CACHE_KEY % key_opts.merge(tail: last_updated)) do
+      # Delete any old key before we write the new one
+      Rails.cache.delete_matched(CACHE_KEY % key_opts.merge(tail: '*'))
+
+      collection.map(&:fields).map(&:keys).uniq.flatten
```
etdsoft, line 18: Another thought, do you really need a new query on L6 (pre changes) to SELECT and ORDER BY?

The collection is already loaded, so just a `map(&:updated_at).sort` may save you a call. I don't think that query is really used.

One issue in Issues#index is that the `collection` is carrying a LEFT OUTER JOIN with it, so I'm thinking reducing the # of queries is probably good (🤷).

app/controllers/issues_controller.rb
```diff
@@ -15,9 +16,10 @@ class IssuesController < AuthenticatedController
   before_action :set_or_initialize_tags, except: [:destroy]
   before_action :set_auto_save_key, only: [:new, :create, :edit, :update]
 
+  EXTRA_COLUMNS = ['Title', 'Tags', 'Affected', 'Created', 'Created by', 'Updated'].freeze
+
   def index
-    @columns = @issues.map(&:fields).map(&:keys).uniq.flatten |
-      ['Title', 'Tags', 'Affected', 'Created', 'Created by', 'Updated']
+    @columns = collection_field_names(@unsorted_issues) | EXTRA_COLUMNS
```
etdsoft, line 22: This change clashes with @aapomm's RTP column change. I think you two need to sync on what the fallback strategy is.

app/controllers/nodes_controller.rb
```diff
@@ -1,18 +1,20 @@
 # This controller exposes the REST operations required to manage the Node
 # resource.
 class NodesController < NestedNodeResourceController
+  include DynamicFieldNamesCacher
   include NodesSidebar
 
   skip_before_action :find_or_initialize_node, only: [ :sort, :create_multiple ]
   before_action :initialize_nodes_sidebar, except: [ :sort, :create_multiple ]
 
+  EXTRA_COLUMNS = ['Title', 'Created', 'Created by', 'Updated'].freeze
```
etdsoft, line 10: In the Pro PR this constant is duplicated.

etdsoft (commented):
Would like to see again please

brianp:
Closing in favour of https://github.com/dradis/dradis-ce/pull/973

PR #964, opened 2022-01-21
Increase evidence content length size

brianp:
### Summary

Some users run into errors where the evidence of the scan files far
exceeds that of our defined content length. Increase the allowed content
length for evidence. There's no reason for us not to if users need the
space. We're also not doing it for _all_ fields because it hasn't proven
necessary in places like issues or notes yet.

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry


engines/dradis-api/spec/requests/dradis/ce/api/v1/evidence_spec.rb
```diff
@@ -229,7 +229,7 @@
         end
 
         context "with params for an invalid evidence" do
-          let(:params) { { evidence: { content: "a"*65536 } } } # too long
+          let(:params) { { evidence: { issue_id: '' } } }
```
brianp, line 232: This change is here because to meet the new maximum to cause an error would require a ~4gb txt file saved in the repo, or 38 seconds to generate that many a's.

config/initializers/db_constants.rb
```diff
@@ -1,2 +1,7 @@
 DB_MAX_STRING_LENGTH = 255
-DB_MAX_TEXT_LENGTH   = 65535
+
+# The max length we set most fields as
+DB_MAX_TEXT_LENGTH = 65535
+
+# The max length supported by the DB
+DB_REAL_MAX_LENGTH = 4294967295
```
etdsoft, line 7: Why is `DB_MAX` enough in some places but `DB_REAL_MAX` in others?

brianp:
Closing as the scope ballooned.

PR #965, opened 2022-01-25
Add sidebar to `revisions#show`

MattBudz:
### Summary

This PR add's a sidebar to the `revisions#show` view for easier navigation.
The sidebar is loaded via ajax to not hinder the performance of the view.

### Check List

- [x] Added a CHANGELOG entry


app/controllers/revisions_controller.rb
```diff
@@ -52,28 +60,29 @@ def load_node
...
-                Issue.find(params[:issue_id])
-              elsif params[:note_id]
-                @node.notes.find(params[:note_id])
-              else
-                raise 'Unable to identify record type'
-              end
+  def set_variables_from_params
+    if params[:card_id]
+      load_list
+      @path = sidebar_project_board_list_card_revisions_path
```
aapomm, line 66: I don't think we need an extra "sidebar" route for every type of item. A single one will suffice  like: `sidebar_project_revisions_path(params)`. That way, we can pass the same `@path` value for each type.

config/routes.rb
```diff
@@ -32,7 +32,11 @@
         member { post :move }
         resources :cards, except: [:index] do
           member { post :move }
-          resources :revisions, only: [:index, :show]
+          resources :revisions, only: [:index, :show] do
+            collection do
```
aapomm, line 36: See comment above about using a single route vs a sidebar route for every type of item.

app/views/revisions/sidebar.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if @issue %>
```
aapomm, line 1: I imagine we have the `@record` here. If so, we can probably simplify this file to:

```
<%= render "#{@record.class.name}/sidebar" %>
```

MattBudz: That won't work because we need to render `nodes/sidebar` for notes and evidence

app/controllers/revisions_controller.rb
```diff
@@ -54,28 +62,37 @@ def load_node
...
-              elsif params[:note_id]
-                @node.notes.find(params[:note_id])
-              else
-                raise 'Unable to identify record type'
-              end
-  rescue ActiveRecord::RecordNotFound
-    flash[:error] = 'Record not found'
-    redirect_to :back
+  def set_revision_params
+    @revision_params = { :"#{@record.class.name.downcase}_id" => @record.id }
```
aapomm, line 66:
```suggestion
    @revision_params = { "#{@record.class.name.downcase}_id" => @record.id }
```

not sure if we need this to be a symbol vs a string

app/controllers/revisions_controller.rb
```diff
@@ -54,28 +62,37 @@ def load_node
...
+    elsif params[:note_id]
+      load_node
+      @record = @node.notes.find(params[:note_id])
+    else
+      raise 'Unable to identify record type'
     end
+
+    set_revision_params
+
+  rescue ActiveRecord::RecordNotFound
```
aapomm, line 94: I'm not fond of using errors as control flow and we may be better off using if else here instead. Instead of using `find` we can use `find_by_id` so that it returns nil if the record is missing instead of throwing an error. Then, we can check for the existence of `@record` here instead.

MattBudz: I removed this rescue block completely since there is no valid case for a record to not be found.

spec/features/revisions_spec.rb
```diff
@@ -6,7 +6,7 @@
   describe 'when the record has 2 revisions' do
     let(:record) do
       with_versioning do
-        create(:issue)
+        create(:issue, node_id: current_project.issue_library.id)
```
etdsoft, line 9: Why is this needed? The Issues factory already accounts for the right Node association.

config/routes.rb
```diff
@@ -51,7 +51,7 @@
 
       resources :evidence, concerns: :multiple_destroy, controller: 'issues/evidence', only: [:index, :new]
       resources :nodes, only: [:show], controller: 'issues/nodes'
-      resources :revisions, only: [:index, :show]
+      resources :revisions, only: [:index, :show]
```
etdsoft, line 54: extra space here, L78, L82

@aapomm this could have been a peer-review suggestion?

app/controllers/revisions_controller.rb
```diff
@@ -54,28 +62,33 @@ def load_node
...
+      @revision_params[:board_id] = @record.board.id
+      @revision_params[:list_id] = @record.list.id
+    end
   end
 
-  def load_sidebar
-    if params[:evidence_id] || params[:note_id]
-      load_node
-    elsif params[:card_id]
+  def set_variables_from_params
```
etdsoft, line 75: this is the wrong name for this helper. It's setting the `@record` variable yes (so `set_record`, but it's also loading collections (e.g. `load_list`, `load_issues`).

It's doing too much. What about extracting the loading of Sidebar collections (which you don't need any more until the ajax call to a controller concern as well?

You duplicate two `if else if` trees, but at least it's easier to follow.

app/controllers/revisions_controller.rb
```diff
@@ -54,28 +62,33 @@ def load_node
...
+      load_issues
+      @record = @issue
+    elsif params[:note_id]
+      load_node
+      @record = @node.notes.find(params[:note_id])
+    else
+      raise 'Unable to identify record type'
     end
+
+    set_revision_params
```
etdsoft, line 92: more over, we're also calling another helper! Should this last one be called from a before_action or the action itself?

There's definitely too much going on here.

app/views/revisions/show.html.erb
```diff
@@ -1,7 +1,18 @@
...
+  <%= content_tag(
+    :div,
+    content_tag(
+      :div,
+      spinner_tag,
+      class: 'd-flex align-items-center justify-content-center h-100'
+    ),
+    data: {
+      behavior: 'fetch',
+      path: project_revisions_sidebar_path(@revision_params)
```
etdsoft, line 13: once extracted (see below) this will have to be a resource-dependent route, so probably set in the controller (for easier logic)

app/views/layouts/nested_node_resource.html.erb
```diff
@@ -1,5 +1,7 @@
 <% content_for :content do %>
-  <%= render "nodes/sidebar" %>
+  <% content_for :sidebar do %>
```
etdsoft, line 2: does `content_for` need to be nested inside the other `content_for`? Seems a bit untidy

PR #966, opened 2022-01-26
Move Revision History to dots menu

MattBudz:
### Summary

Recent activity and revisions history show similar information. This PR consolidates the two and add's a `View History` link to the dots-menu.

### Check List

- [x] Added a CHANGELOG entry


spec/support/activity_shared_examples.rb
```diff
@@ -54,32 +54,3 @@
     expect { submit_form }.not_to have_enqueued_job(ActivityTrackingJob)
   end
 end
-
-# Define the following let variables before using these examples:
-#
-#   create_activities : a block which creates the activities AND IS CALLED
-#                       BEFORE THE PAGE LOADS
-#   trackable: the model which the 'show' page is about
-shared_examples 'a page with an activity feed' do
```
aapomm, line 63: I'm realizing that we still have this feed but it's now in the revisions view. In that case we still need to have a spec for that.

spec/features/revisions_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'rails_helper'
+
+describe 'Revisions#show:' do
+  subject { page }
+
+  describe 'when the record has 2 revisions' do
+
```
aapomm, line 7:
```suggestion
```

spec/features/revisions_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    before do
+      login_to_project_as_user
+
+      with_versioning do
+        record.text = 'updated text'
+        record.save
+      end
+    end
+    
+    it 'the 2 revisions are listed in the history table' do
```
aapomm, line 23:
```suggestion
    it 'lists the revisions in the history table' do
```

to make it make sense with the `it` block

spec/support/activity_shared_examples.rb
MattBudz: added specs

PR #973, opened 2022-01-28
[Feature] Datatables column changes

aapomm:
### Summary

Merged feature branches:

- `rtp-default-fields`

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


app/controllers/issues/evidence_controller.rb
```diff
@@ -0,0 +1,19 @@
+class Issues::EvidenceController < IssuesController
+  before_action :set_issue, only: [:index]
+  before_action :set_affected_nodes, only: [:index]
+
+  EXTRA_COLUMNS = ['Created by', 'Created', 'Updated'].freeze
+  SKIP_COLUMNS = ['Title', 'Label'].freeze
+
+  def index
+    @evidence_columns = ['Node'] | (collection_field_names(@issue.evidence) - SKIP_COLUMNS) | EXTRA_COLUMNS
```
etdsoft, line 9: Not a fan of `SKIP_COLUMNS`, Title and Label are "pseudofields" that come from:

https://github.com/dradis/dradis-ce/blob/develop/app/models/evidence.rb#L33

For all intents and purposes, they're real fields, so I don't think we need to skip them. I'd definitely hide them by default, but I don't think we should obliterate them.

Also, Label and Node are prolly equivalent?

app/controllers/nodes_controller.rb
```diff
@@ -1,18 +1,20 @@
...
 
   # GET /nodes/<id>
   def show
     @activities       = @node.nested_activities.latest
-    @note_columns     = @sorted_notes.map(&:fields).map(&:keys).uniq.flatten \
-                      | ['Title', 'Created', 'Created by', 'Updated']
-    @evidence_columns = @sorted_evidence.map(&:fields).map(&:keys).uniq.flatten \
-                      | ['Title', 'Created', 'Created by', 'Updated']
+    @note_columns     = collection_field_names(@node.notes) | EXTRA_COLUMNS
+    @evidence_columns = collection_field_names(@node.evidence) | @rtp_default_evidence_fields | EXTRA_COLUMNS
```
etdsoft, line 17: I think this is not a bad pattern, but it has to be consistent with `Issues#index` and `Issues::Evidence#index`

What I don't like is that we're `|` the return from a method, with a local variable, with a constant.

app/models/project.rb
```diff
@@ -116,4 +116,8 @@ def plugin_uploads_node
   def recovered
     @recovered ||= nodes.find_or_create_by(label: 'Recovered')
   end
+
+  def report_template_properties
+    nil
+  end
```
etdsoft, line 122: Feels like #weaksauce but perhaps is #weaksauce we have to live with.

app/views/issues/evidence/_table.html.erb
```diff
@@ -0,0 +1,53 @@
...
+    <% @affected_nodes.each do |node| %>
+      <% node.evidence.where(issue_id: @issue.id).each do |evidence| %>
+        <tr id="evidence-<%= evidence.id %>">
+          <td class="select-checkbox" data-behavior="select-checkbox"></td>
+          <% @evidence_columns.each do |column| %>
+            <%
+              sort, display =
+              case column
+              when 'Node'
+                [evidence.node.label, link_to(evidence.node.label, [current_project, node, evidence])]
```
etdsoft, line 25: `Label` quacks better than this `Node` field

app/views/issues/show.html.erb
```diff
@@ -15,8 +15,8 @@
 <!-- Main content -->
 <div class="row m-0">
   <ul class="tabs-container nav nav-tabs main-tabs w-100">
-    <li class="nav-item"><a href="#info-tab" data-toggle="tab" class="nav-link active"><%= colored_icon_for_model(@issue, 'fa-bug') %> Information</a></li>
-    <li class="nav-item"><a href="#evidence-tab" class="nav-link" data-toggle="tab"><i class="fa fa-laptop"></i> Evidence <span class="badge bg-primary text-white"><%= @issue.evidence.count %></span></a></li>
+    <li class="nav-item"><a href="#info-tab" data-toggle="tab" class="nav-link <%= class_names(active: !params[:tab]) %>"><%= colored_icon_for_model(@issue, 'fa-bug') %> Information</a></li>
+    <li class="nav-item"><a href="#evidence-tab" class="nav-link <%= class_names(active: params[:tab] == 'evidence-tab') %>" data-toggle="tab"><i class="fa fa-laptop"></i> Evidence <span class="badge bg-primary text-white" data-behavior="evidence-count"><%= @issue.evidence.count %></span></a></li>
```
etdsoft, line 19: `class_names` #ftw

app/views/layouts/tylium.html.erb
```diff
@@ -82,7 +82,7 @@
       <%= render 'boards/modals/new_board_form' unless request.xhr? %>
     <% end %>
 
-    <% if controller_name == 'evidence'%>
+    <% if controller_path == 'evidence' %>
```
etdsoft, line 85: 👏

app/views/nodes/items_table/_table.html.erb
```diff
@@ -2,9 +2,10 @@
   <div class="table-wrapper">
     <table class="table table-striped"
       data-behavior="dradis-datatable"
-      data-default-columns='["Title", "Created", "Updated"]'
-      data-item-name="<%= items.first.class.name.underscore %>"
-      data-local-storage-key="project.ce.<%= dom_id(current_project) %>.<%= items.first.class.name.underscore.pluralize %>_datatable">
+      data-default-columns=<%= @default_columns.fetch(class_name.to_sym, ['Title', 'Created', 'Updated', 'Validation']).to_json %>
```
etdsoft, line 5: smell

app/views/nodes/items_table/_table.html.erb
```diff
@@ -2,9 +2,10 @@
...
+      data-item-name="<%= class_name %>"
```
etdsoft, line 6: I do think `class_name` is better than dragging the `items.first...` list throughout, but perhaps, we can define it ourselves at the beginning of the partial via `<% class_name = items.first.class.name.underscore %> than have the caller pass botht he `items` collection, and also spell out the `class_name` for the collection's items?

spec/features/issues_spec.rb
```diff
@@ -322,112 +320,6 @@
           include_examples 'deleted item is listed in Trash', :issue
           include_examples 'recover deleted item', :issue
         end
-
-        describe 'add evidence', js: true do
```
etdsoft, line 326: Where did all of this go?

MattBudz: New specs are [here](https://github.com/dradis/dradis-ce/pull/973/files#diff-ccbb31fd05b3f301b632274f365f30c66c50ccf8f425687875bd638821d7b1f6)

And [here](https://github.com/dradis/dradis-ce/pull/973/files#diff-47f3c1aa36f81a81a36d2c830ee2190892338ca03f5318dc7447c26759aa9bb7)]

app/controllers/issues/evidence_controller.rb
brianp: I think we can ditch the skip addition and unify this like the rest.

app/controllers/nodes_controller.rb
brianp: > What I don't like is that we're | the return from a method, with a local variable, with a constant.

I hear that. We can improve it. I've got an idea for a more unified solution.

app/views/issues/evidence/_table.html.erb
MattBudz: I thought `Node` would be more clear to users when they are looking at the table but now that I'm looking at this again I have pushed a change to replace `Node` with `Affected` for consistency with `issues#index`

app/views/nodes/items_table/_table.html.erb
aapomm: Fixed ✅

app/views/nodes/items_table/_table.html.erb
aapomm: Updated to use a hash with default values for simplicity ✅

app/models/project.rb
aapomm: The alternative to this is explicit edition checks with `defined?`. While that is what we do with some parts of the app, this is way simpler and something I think we can live with.

app/controllers/issues/evidence_controller.rb
brianp: Updated.

app/controllers/nodes_controller.rb
brianp: Updated.

app/controllers/issues/evidence_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    default_field_names = ['Affected', 'Title', 'Created', 'Updated'].freeze
+    extra_field_names = ['Created by'].freeze
+
+    dynamic_fields = dynamic_field_names(@issue.evidence)
+
+    rtp = current_project.report_template_properties
+    rtp_default_fields = rtp ? rtp.issue_fields.default.field_names : []
+
+    @default_columns = rtp_default_fields.presence || default_field_names
+    @evidence_columns = default_field_names | rtp_default_fields | dynamic_fields | extra_field_names
```
etdsoft, line 33: My initial instinct was to ask for a rename here for `@all_columns` which would make more sense (we're in the context of EvidenceController after all) but then I thought: not worth it.

Now I see we're going `@all_columns` in IssuesController, so have the consistency excuse to make the request.

app/controllers/issues/evidence_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  def new
+    @nodes_for_add_evidence = current_project.nodes.user_nodes.order(:label)
+
+    @template_content = template_content if params[:template]
+  end
+
+  private
+
+  def set_columns
+    default_field_names = ['Affected', 'Title', 'Created', 'Updated'].freeze
```
etdsoft, line 24: We can move 'Created' and 'Updated' of the default-show list (consistent with Issues) and use the already available Evidence#local_field 'Label' instead of 'Affected'

app/controllers/issues/evidence_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    extra_field_names = ['Created by'].freeze
```
etdsoft, line 25:
```suggestion
    default_field_names = ['Label', 'Title'].freeze
    extra_field_names = ['Created', 'Created by', 'Updated'].freeze
```

app/controllers/nodes_controller.rb
```diff
@@ -114,6 +113,26 @@ def tree
   end
 
   private
+
+  def set_evidence_default_columns
```
brianp, line 117:
```suggestion
  def set_columns
```

app/controllers/nodes_controller.rb
```diff
@@ -1,18 +1,17 @@
 # This controller exposes the REST operations required to manage the Node
 # resource.
 class NodesController < NestedNodeResourceController
+  include DynamicFieldNamesCacher
   include NodesSidebar
 
   skip_before_action :find_or_initialize_node, only: [ :sort, :create_multiple ]
   before_action :initialize_nodes_sidebar, except: [ :sort, :create_multiple ]
+  before_action :set_evidence_default_columns, only: :show
+  before_action :set_columns, only: :index
```
etdsoft, line 10: I think this is a typo? We don't even have an #index action

app/controllers/nodes_controller.rb
```diff
@@ -1,18 +1,17 @@
 # This controller exposes the REST operations required to manage the Node
 # resource.
 class NodesController < NestedNodeResourceController
+  include DynamicFieldNamesCacher
   include NodesSidebar
 
   skip_before_action :find_or_initialize_node, only: [ :sort, :create_multiple ]
   before_action :initialize_nodes_sidebar, except: [ :sort, :create_multiple ]
+  before_action :set_evidence_default_columns, only: :show
```
etdsoft, line 9: We use `:set_columns` in other controllers. Arguably you could end up here with `set_evidence_columns` and `set_note_columns` at some point, but I don't think we need the `_default_` part

app/controllers/nodes_controller.rb
```diff
@@ -114,6 +113,26 @@ def tree
   end
 
   private
+
+  def set_evidence_default_columns
+    default_field_names = ['Title', 'Created', 'Updated'].freeze
+    extra_field_names = ['Created by'].freeze
```
etdsoft, line 119: Let's make this consistent with `Issues::Evidence#index` table: Label and Title, moving Created and Updated down to `extra_`

app/controllers/nodes_controller.rb
```diff
@@ -114,6 +113,26 @@ def tree
...
+
+    note_dynamic_fields = dynamic_field_names(@node.notes)
```
etdsoft, line 121: OK, if we're doing this, then the method needs to also lose `evidence_`

app/controllers/nodes_controller.rb
```diff
@@ -114,6 +113,26 @@ def tree
...
+    rtp = current_project.report_template_properties
+    rtp_default_evidence_fields = rtp ? rtp.evidence_fields.default.field_names : []
+
+    @note_columns     = default_field_names | note_dynamic_fields | extra_field_names
+    @evidence_columns = default_field_names | rtp_default_evidence_fields | evidence_dynamic_fields | extra_field_names
+
+    @default_columns = {
+      evidence: rtp_default_evidence_fields.presence || default_field_names,
+      note: default_field_names
+    }
```
etdsoft, line 133: I don't love that we keep separate `@note_columns` and `@evidence_columns` but merge them into a single `@default_columns`, I think`@default_note_columns` and `@default_evidence_columns` makes a bit more sense?

app/views/issues/_sidebar.html.erb
```diff
@@ -9,12 +9,12 @@
 
 <%= render 'shared/sidebar_collection',
   category_id: nil,
-  collection: @issues,
+  collection: @issues.sort,
```
etdsoft, line 12: This is the controller's task, not the view's.

app/controllers/issues/evidence_controller.rb
brianp: I'm all for "all_columns"

app/controllers/issues/evidence_controller.rb
brianp: Done

app/controllers/nodes_controller.rb
brianp: I think it's some bad copy/paste and weird expectations. The method should be `set_columns, only: :show` but it was "working" because the method is named wrong, and already has a before action. I'll make this consistent with `set_columns`

app/controllers/nodes_controller.rb
brianp: Done

app/controllers/nodes_controller.rb
brianp: Done

app/controllers/issues_controller.rb
```diff
@@ -122,17 +119,40 @@ def import
...
+  def set_affected_nodes
+    @affected_nodes = Node.joins(:evidence)
+                          .select('nodes.id, label, type_id, count(evidence.id) as evidence_count, nodes.updated_at')
+                          .where('evidence.issue_id = ?', @issue.id)
+                          .group('nodes.id')
+                          .sort_by { |node, _| node.label }
+  end
+
+  def set_columns
+    default_field_names = ['Title', 'Tags', 'Affected', 'Validation'].freeze
```
brianp, line 131:
```suggestion
    default_field_names = ['Title', 'Tags', 'Affected'].freeze
```

app/views/issues/_sidebar.html.erb
MattBudz: fixed

PR #974, opened 2022-01-28
Reduce local auto save shared examples specs

sean-yeoh:
### Summary
We're running redundant specs for the local auto save feature, which wastes time in the product workflow.

### Proposed solution
Run the JS functionality for local auto save once (Cache different inputs, clearing cache).

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


spec/features/local_auto_save_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+  let(:list) { create(:list, board: board) }
+
+  let!(:users) do
+    user1 = create(:user, :author)
+    user2 = create(:user, :author)
+
+    [
+      user1,
+      user2
+    ]
```
aapomm, line 18: use `create_list` instead?

spec/features/local_auto_save_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+
+  describe 'custom tag input' do
+    let!(:tags) do
+      [
+        current_project.tags.create(name: '!9467bd_critical'),
+        current_project.tags.create(name: '!d62728_high')
+      ]
+    end
+
+    it 'will be cached' do
```
aapomm, line 29: I don't think the tense is consistent with the other assertions.

```suggestion
    it 'caches the content' do
```

spec/features/local_auto_save_spec.rb
sean-yeoh: Sounds good

PR #976, opened 2022-01-31
Add Deploy to Heroku button

etdsoft:
### Summary

Alternative approach to https://github.com/dradis/dradis-ce/pull/943 and https://github.com/dradis/dradis-ce/pull/975

Here we don't have a custom `database.yml.heroku` (we use `sed` to replace the adapter) and make `pg` part of the bundle. It adds another development dependency (easily satisfied in Ubuntu) to simplify parity between Heroku and non-Heroku bundles.

### Check List

- [x] Added a CHANGELOG entry

Deploy to Heroku button added to README.

### DIFF
--- a/app.json
+++ b/app.json
@@ -26,7 +26,7 @@
   "name": "Dradis Framework - Community Edition",
   "repository": "https://github.com/dradis/dradis-ce/tree/deploy-to-heroku",
   "scripts": {
-    "postdeploy": "bin/heroku"
+    "postdeploy": "rails /bin/heroku"
   },
   "website": "http://drad.is"
 }
--- a/app.json
+++ b/app.json
@@ -26,8 +26,9 @@
   "name": "Dradis Framework - Community Edition",
   "repository": "https://github.com/dradis/dradis-ce/tree/deploy-to-heroku",
   "scripts": {
-    "heroku-postbuild": "sed -i 's/adapter: sqlite3/adapter: postgresql/' config/database.yml.template",
-    "postdeploy": "bin/setup"
+    "heroku-prebuild": "sed -i 's/adapter: sqlite3/adapter: postgresql/' config/database.yml.template",
+    "heroku-postbuild": "cp config/database.yml.template config/database.yml",
+    "postdeploy": "bundle exec rake db:prepare"
   },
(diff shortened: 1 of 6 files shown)


bin/heroku
```diff
@@ -0,0 +1,28 @@
...
+# path to your application root.
+APP_ROOT = File.expand_path('..', __dir__)
+
+def system!(*args)
+  system(*args) || abort("\n== Command #{args} failed ==")
+end
+
+FileUtils.chdir APP_ROOT do
+  puts "\n== Copying sample files =="
+  # FileUtils.cp 'config/database.yml.template', 'config/database.yml'
```
aapomm, line 13: Remove unused code

bin/heroku
```diff
@@ -0,0 +1,28 @@
...
+  FileUtils.cp 'config/secrets.yml.template', 'config/secrets.yml'
+  FileUtils.cp 'config/smtp.yml.template', 'config/smtp.yml'
+
+  puts "\n== Replacing DB adapter =="
+  # system! "sed -i 's/adapter: sqlite3/adapter: postgresql/' config/database.yml"
```
aapomm, line 18: ditto

app.json
```diff
@@ -0,0 +1,32 @@
...
+    {
+      "url": "heroku/nodejs"
+    },
+    {
+      "url": "heroku/ruby"
+    }
+  ],
+  "description": "The original open-source reporting and collaboration tool trusted by InfoSec professionals around the world.",
+  "env": {
+    "HEROKU": {
```
aapomm, line 15: Where are we using this ENV variable?

app.json
```diff
@@ -0,0 +1,32 @@
+{
+  "buildpacks": [
+    {
+      "url": "heroku-community/apt"
+    },
+    {
+      "url": "heroku/nodejs"
```
aapomm, line 7: Do we need this buildpack? If so, what for?

etdsoft: We do, assets:precompile would fail if the `yarn` binary isn't installed, this buildpack does the trick.

app.json
etdsoft: Nowhere! 👍

Removing.

PR #980, opened 2022-02-01
Setup Wizard: allow users to choose a Kit to start with

etdsoft:
### Summary

Before this PR, users could use ./bin/setup to start with a Welcome project to show them around.

This PR makes that part of the initial Setup Wizard, and users will be able to use the browser to choose between a Welcome kit and an empty slate.

Apart from the main goal of this PR, we had to go down a few rabbit holes, so it ended up in the double-digits and doing more than it says on the tin. However, most of this changes were already reviewed as part of:

https://github.com/dradis/dradis-ce/pull/981
https://github.com/dradis/dradis-ce/pull/982
https://github.com/dradis/dradis-ce/pull/983

To recap:
  - There's a new Step 2: Kits in the Setup Wizard
  - KitImportJob has been refactored to take either a Kit .zip file or a folder containing a Kit
  - Thor dradis:setup:welcome has been refactored to account for the previous.
  - Thor dradis:reset has been refactored and cleaned after 4~6 years to use better practices
  - Welcome kit files in lib/tasks/templates/ have been moved under lib/tasks/templates/welcome/
  - And a few specs were cleaned up along the way
  - bin/setup no longer needs to load a Welcome all the time

### Check List

- [x] Added a CHANGELOG entry


app/jobs/kit_import_job.rb
```diff
@@ -37,19 +38,31 @@ def perform(file:, logger:, user_id: nil)
...
     logger.info { 'Assigning RTP to project...' }
 
     @project.update_attribute :report_template_properties_id, @word_rtp.id if @word_rtp
   end
 
+  def copy_kit_to_working_dir(source)
+    if File.file?(source)
+      file = source
+      FileUtils.cp file, working_dir
+      unzip(file)
```
aapomm, line 57:
```suggestion
      FileUtils.cp source, working_dir
      unzip(source)
```

app/views/layouts/setup.html.erb
```diff
@@ -0,0 +1,31 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <title><%= content_for?(:title) ? yield(:title) : "Dradis Community Edition" %></title>
```
aapomm, line 4:
```suggestion
    <title><%= content_for?(:title) ? yield(:title) : 'Dradis Community Edition' %></title>
```

app/jobs/kit_import_job.rb
```diff
@@ -9,25 +9,26 @@ class KitImportJob < ApplicationJob
...
-    FileUtils.cp file, temporary_dir
-    unzip
+    copy_kit_to_working_dir(file_or_folder)
+
     import_methodology_templates
     import_note_templates
     import_plugin_templates
     import_project_package
     import_project_templates
     import_report_template_files
```
aapomm, line 30: I'm getting some errors around this area since the kit included here does not have a project template and an RTP. I think we should add a guard to check for the existence of the relevant directories.

PR #981, opened 2022-02-02
KitImportJob - do not delete kit file

etdsoft:
The Job didn't create the file it's receiving as an argument, so it
should concern itself with deleting it either.

This change is required in preparation to allow for the Job to receive
either a Kit file or a folder containing the Kit's contents.

### Check List

- [x] Added a CHANGELOG entry

Internal refactor, doesn't modify any user-facing features.

### DIFF
--- a/app/jobs/kit_import_job.rb
+++ b/app/jobs/kit_import_job.rb
@@ -38,7 +38,6 @@ class KitImportJob < ApplicationJob
   ensure
     logger.info('Worker process completed.')
     FileUtils.remove_entry temporary_dir
-    File.unlink(file) if File.exists?(file)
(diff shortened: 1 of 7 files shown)


app/jobs/kit_import_job.rb
```diff
@@ -38,7 +38,6 @@ def perform(file:, logger:, user_id: nil)
   ensure
     logger.info('Worker process completed.')
     FileUtils.remove_entry temporary_dir
-    File.unlink(file) if File.exists?(file)
```
brianp, line 41: 👍🏻

brianp (approved):

PR #984, opened 2022-02-04
Tags and Node links

brianp:
## Issue table:
Tags weren't included in the set of columns to select from.
Affected wasn't included in the set of columns to select from.

They were both included as default, but if the default didn't render they were missed entirely. 

Nodes were not linked for easy access.

## Evidence table:
No longer used Affected field for node links. Needed updating to use the Label column for node links. 

### Check List

### DIFF
--- a/app/controllers/issues_controller.rb
+++ b/app/controllers/issues_controller.rb
@@ -137,7 +137,7 @@ class IssuesController < AuthenticatedController
     rtp_default_fields = rtp ? rtp.issue_fields.default.field_names : []
 
     @default_columns = rtp_default_fields.presence || default_field_names
-    @all_columns = ['Title'] | rtp_default_fields | dynamic_fields | extra_field_names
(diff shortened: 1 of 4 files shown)


app/controllers/issues_controller.rb
```diff
@@ -137,7 +137,7 @@ def set_columns
     rtp_default_fields = rtp ? rtp.issue_fields.default.field_names : []
 
     @default_columns = rtp_default_fields.presence || default_field_names
-    @all_columns = ['Title'] | rtp_default_fields | dynamic_fields | extra_field_names
+    @all_columns = ['Title'] | rtp_default_fields | default_field_names | dynamic_fields | extra_field_names
```
brianp, line 140: Previously we removed this under the assumption "everything would get brought in via the dynamic fields" but it turns out we were wrong. Tags and Affected were considered default fields, but not included as extra fields.

PR #991, opened 2022-02-15
Consistent resource headers

MattBudz:
### Summary

There are some areas of the app that inconsistently display resource titles. Some are the actual titles while others are just some hard-coded copy. This PR improves the consistency within projects by displaying the resource title as the header (where possible). 

### Check List

- [x] Added a CHANGELOG entry


app/helpers/issues_helper.rb
```diff
@@ -59,16 +59,26 @@ def issue_tags(issue)
...
+                  [
+                    colored_icon_for_model(issue, 'fa-bug'),
+                    h(tag.display_name)
+                  ]
+                else
+                  [
+                    content_tag(:i, nil, class: 'fa fa-bug'),
+                    'Untagged'
+                  ]
+                end
```
aapomm, line 74:
```suggestion
      tag = issue.tags.first
      content =
        if tag
          [
            colored_icon_for_model(issue, 'fa-bug'),
            h(tag.display_name)
          ]
        else
          [
            content_tag(:i, nil, class: 'fa fa-bug'),
            'Untagged'
          ]
        end
```

PR #1001, opened 2022-03-22
Reduce datatable shared examples specs

sean-yeoh:
### Summary
We're running redundant specs for the datatables feature, which wastes time in the product workflow.

### Proposed solution
Run the JS functionality for datatable once (multi tagging, mutli delete, filter, hide/show columns)

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


spec/features/issue_pages/datatable_spec.rb
```diff
@@ -1,8 +1,41 @@
...
+        text: "#[Title]#\nIssue1\n\n#[Risk]#\nHigh\n\n#[Description]#\nn/a",
+        node: current_project.issue_library
+      ),
+      create(:issue, node: current_project.issue_library)
+    ]
+  end
+
+  let(:tags) do
+    Tag::DEFAULT_TAGS.map do |tag|
+      if defined?(Dradis::Pro)
```
aapomm, line 20: I feel like we can improve this. Is it possible to update the factory instead and have it accept a project association even in CE? Seeing as `current_project` exists in both editions, the ideal code here would be: `create(:tag, name: tag, project: current_project)`. I do see that this piece of code is used elsewhere in the test suite so I'm fine with moving forward with this.

sean-yeoh: > Is it possible to update the factory instead and have it accept a project association even in CE?
Yeah we can use transient attributes https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#transient-attributes

PR #1003, opened 2022-03-23
Fix datatables resetting intermittently

sean-yeoh:
### Summary
Currently, DataTables reset its state after 2 hours.

### Proposed solution
Add stateDuration: 0 to when initializing DataTables https://datatables.net/reference/option/stateDuration

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/assets/javascripts/shared/datatable/main.js
+++ b/app/assets/javascripts/shared/datatable/main.js
@@ -200,14 +200,14 @@ class DradisDatatable {
       //
       // To prevent a reset from happening, we just have to ensure that the number of columns on the page
       // matches the length of columns array in the saved state object.
-      stateLoadCallback: function(_settings) {
+      stateLoadCallback: function(_settings, callback) {
         var localStorageData = JSON.parse(localStorage.getItem(that.localStorageKey));
 
-        if (localStorageData !== null) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/shared/datatable/main.js
```diff
@@ -200,14 +200,14 @@ class DradisDatatable {
...
         var localStorageData = JSON.parse(localStorage.getItem(that.localStorageKey));
 
-        if (localStorageData !== null) {
-          return that.rebuildSavedStateColumnsFromLocalStorage(localStorageData);
-        } else {
+        if (!localStorageData) {
           return null;
         }
+
+        callback(that.rebuildSavedStateColumnsFromLocalStorage(localStorageData));
```
aapomm, line 210: would something like this work instead:
```suggestion
        if (localStorageData) {
          callback(that.rebuildSavedStateColumnsFromLocalStorage(localStorageData));
        }
```

sean-yeoh: We have to return null explicitly. I tried removing it previously but the datatable broke when initializing.

image

> If the data is loaded synchronously the return value should be the loaded state (or null if no data was loaded).

https://datatables.net/reference/option/stateLoadCallback

PR #1004, opened 2022-03-25
Login view refresh

MattBudz:
### Summary

A refresh of the login view.

### Check List

- [x] Added a CHANGELOG entry


app/assets/javascripts/shared/sessions.js
```diff
@@ -0,0 +1,31 @@
...
+    setTimeout(function() {
+      setInterval(function() {
+        if (charCount <= randomString.length) {
+          $('[data-behavior~=mission-text]').text(randomString.substring(0, charCount));
+          charCount += 1;
+        }
+      }, 20);
+    }, 500);
+
+    $('[data-behavior~=logo-icons]').addClass('expand');
```
sean-yeoh, line 22: Since this is adding the `expand` class on page load, is there any difference adding `expand` to the markup directly https://github.com/dradis/dradis-ce/pull/1004/files#diff-f43fe075643e681b2c01c2f853bb0c4299d135b47fcbd4da96890d521c49e3ebR22?

MattBudz: If we add the class directly to the element then it will instantly show the icons without doing the animation.

PR #1005, opened 2022-03-30
Gateway comments view hooks

MattBudz:
### Summary

Add split button to comment feed for improved Pro parity. 

### Check List


app/assets/javascripts/tylium/behaviors.js.coffee
```diff
@@ -63,7 +63,7 @@ document.addEventListener "turbolinks:load", ->
   $('.modal').on 'shown.bs.modal', ->
     $(this).find('input:text:visible:first').focus()
 
-  $('.js-try-pro').on 'click', ->
+  $('body').on 'click', '.js-try-pro', ->
```
sean-yeoh, line 66: @MattBudz Can you add a comment here on why this change is needed for a smoother review process? Thanks

MattBudz: We need this change in order for the upsell modal to appear. The trigger for the modal is the link in the split button, but the split button is added to the DOM via ajax. Before this change, the eventListener would bind to `.js-try-pro` on page load so it wouldn't work for the split button since the split button doesn't exist on page load.

PR #1009, opened 2022-04-05
Gateway comments notify

aapomm:
### Summary

- Create notifications in comment for each passed recipient id 

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -39,17 +39,22 @@ class Comment < ApplicationRecord
   def notify(action:, actor:, recipients:)
     case action.to_s
     when 'create'
-      create_notifications(action: :mention, actor: actor, recipients: recipients)
+      if recipients
+        create_notifications(action: :mention, actor: actor, recipients: recipients)
+      end
 
       subscribe_mentioned()
-      create_notifications(action: :mention, actor: actor,  recipients: mentions)
+      create_notifications(action: :mention, actor: actor, recipients: mentions)
 
       # We're finding subscribers that have not been mention here
       # using ActiveRecord because create_notifications expect recipients
       # to be an ActiveRecord::Relation.
       subscribers = User.includes(:subscriptions).where(
-        subscriptions: { subscribable_id: commentable.id }
-      ).where.not(id: [user.id] + mentions.pluck(:id))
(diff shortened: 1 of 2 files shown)


app/models/comment.rb
```diff
@@ -39,6 +39,8 @@ def create_subscription
   def notify(action:, actor:, recipients:)
     case action.to_s
     when 'create'
+      create_notifications(action: :mention, actor: actor, recipients: recipients)
```
sean-yeoh, line 42: Will this create duplicated notifications since it isn't excluded like on line: 52 for mentions?

aapomm: Not currently since line 52 checks for users with subscriptions and we don't have that implemented yet.

sean-yeoh: Since we don't have a check in https://github.com/secroots/dradis-pro/blob/develop/app/models/concerns/notifiable.rb#L8, wouldn't this raise an error if `recipients` is `nil`?

aapomm: Looks like I'm mistaken and the comment builds a subscription on creation. I added the recipients to the line with exclusions to prevent doubling of the notifications.

app/models/comment.rb
```diff
@@ -39,15 +39,22 @@ def create_subscription
   def notify(action:, actor:, recipients:)
     case action.to_s
     when 'create'
+      if recipients
+        create_notifications(action: :mention, actor: actor, recipients: recipients)
```
etdsoft, line 43: Why do we need this extra call? It seems we're duplicating what we used to do with a single line in L43 (pre) and L47 (now).

etdsoft (commented):
I'm not sure I understand what's going on here.

app/models/comment.rb
aapomm: L47 notifies the all the mentioned users in the comment while L58 mentions the subscribers. To implement the described feature, this extra line will have to be added. Note that the `recipients` argument is not used before but it was there because it's used by the other notifiables and the caller would always pass that argument. As for why this was done, I've updated the card to describe the reasoning.

PR #1033, opened 2022-06-16
Added multi tags for issues

marschatha:
### Summary

At the moment we have a Tag and a Tagging model. Tagging is a polymorphic association which allows us to tag any model in the app. However we're only tagging Issues for now.

Along the same lines an Issue can have many Tags (through Taggings) at the database level. However, at the UI level we're limiting the association to a single tag per Issue.

The 1 Tag per Issue rule was implemented with the expectation that tags will be used for a single purpose: associate an issue with a given risk rating (e.g. "High" vs "Medium" vs "Low"). However, users are finding the rule limiting and would like to be able to have multiple tags associated with each issue.

**Proposed solution**

Users should be allowed to associate multiple tags with an issue in Issues#new and Issues#edit. The current drop-down tag list should be replaced with a text area that lets users assign multiple tags at a time.

A tag is added by typing the name of the tag and hitting <Enter>, if a tag already exists in the project with the same name, that tag is added to the issue (via a new Tagging), otherwise a new Tag is created and then a new Tagging.

### How to test

Go to an issue and add multiple tags, you can also add multiple tags from the main test field now.
Tags that have been added can also be removed by clicking the `x` icon next to them. 
Multi tags are also compatible with local save feature.

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/assets/javascripts/shared/auto_save/local.js
+++ b/app/assets/javascripts/shared/auto_save/local.js
@@ -91,9 +91,6 @@ class LocalAutoSave {
           }
         } else if (element.type === 'file') {
           element.value = '';
-        } else if (element.name === 'issue[tag_list]') {
-          var $tagDropdownItem = $(`.js-taglink[data-tag='${item[1]}']`);
-          new SelectTagDropdown($tagDropdownItem)
         } else {
           element.value = item[1];
         }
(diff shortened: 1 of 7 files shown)


etdsoft:
Thanks for this @marschatha !

PR #1035, opened 2022-06-21
Tag crud

RiyadBen:
### Summary

The current state of the application uses a many-to-many relationship between Issues and Tags through Taggings but , it's not being used as intended  from the UI , as it only sends one tag at a time.
This can be very limiting to the user , thus , it's important to add into the views the possibility to add multiple tags as needed.

### Solution
The user should be allowed to enter many tags at a time , using tags input or multiselect dropdown.
In my solution i used a pill-like tags input.

### How to test
When creating a new issue or editing an existing one , the tags input is shown for the user , they can write any tag , press space or ',' and it will be automatically added to the tags list that will be sent , they can press `x` to remove an entry from the list and they can also use `backspace` to remove the most recent one.

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/assets/javascripts/tylium.js
+++ b/app/assets/javascripts/tylium.js
@@ -68,3 +68,5 @@
 //= require tylium/pages/projects/boards_summary
 //= require tylium/pages/projects/issues_chart
 //= require tylium/pages/projects/summary
+
+//= require tylium/pages/issues_tags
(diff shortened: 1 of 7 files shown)


etdsoft:
Thanks for sending this, closing for now.

PR #1037, opened 2022-06-28
Ruby 3.1.2

aapomm:
### Summary

Upgrade Dradis ruby to 3.1.2

Updated gems:
- unicorn (`unicorn-5.4.1 requires ruby version < 3.0, which is incompatible with the current version, ruby 3.1.2p20`)
- bootsnap (`realpath_cache.rb:17:in `dirname': no implicit conversion of nil into String (TypeError)`)
- net-* gems and matrix gem (https://rubyreferences.github.io/rubychanges/3.0.html#libraries-promoted-to-default-gems)
- acts_as_tree (`NoMethodError: undefined method `arity'`)
- thor (`Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated`)
- timecop (`no implicit conversion of Hash into Integer` when running the specs)
- factory_bot (`wrong number of arguments (given 2, expected 1)` when running the specs)
- bundler-audit (`Tried to load unspecified class: Date (Psych::DisallowedClass)`)
- paper_trail (`Psych::DisallowedClass:`)

Also, `YAML.load_file` method was changed to `YAML.unsafe_load_file`

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry


spec/features/node_merging_spec.rb
```diff
@@ -21,6 +21,8 @@
     click_link 'Merge'
   end
 
+  after { FileUtils.rm_rf(Dir.glob(Attachment.pwd + '*')) }
```
sean-yeoh, line 24: Why is this change needed?

aapomm: In it's original location, it wouldn't trigger if the `expect` fails. Putting it in the `after` block ensures it triggers even if the test fails.

etdsoft: it seems we'd have this "gotcha" with failing specs elsewhere in the specs/ folder, no?

etdsoft (commented):
I'm really worried about this:

> Also, YAML.load_file method was changed to YAML.unsafe_load_file

Instead of using a method that by the looks of it is unsafe, what's the recommended "safe" way to call `YAML.load` now?

spec/features/node_merging_spec.rb
aapomm: Looks like the issue is with the rails_helper and not this. I've removed this after callback and update the cleanup in the rails_helper

PR #1055, opened 2022-07-29
Fix heroku 1 click

aapomm:
### Spec
Errors for 1 click deploy:
```
app[web.1]: fatal: not a git repository (or any parent up to mount point /)
app[web.1]: Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
app[web.1]: NOTE: nokogumbo: Using Nokogiri::HTML5 provided by Nokogiri. See https://github.com/sparklemotion/nokogiri/issues/2205 for more information.
app[web.1]: Could not find server "". 
app[web.1]: Run `bin/rails server --help` for more options.
```
This seems to be caused by the Ruby 3 removing Webrick as a bundled gem. Previously, the deployed app uses Webrick by default when being deployed.

**Proposed solution**
The solution I went with is moving the Puma gem outside of the development group in the Gemfile allowing it to be installed and used on deploy.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


.do/deploy.template.yaml
```diff
@@ -10,7 +10,7 @@ spec:
       value: enabled
       scope: RUN_TIME
     git:
-      branch: develop
+      branch: fix-heroku-1-click
```
aapomm, line 13: This needs to be reverted to `develop` once merged/before merging.

config/cable.yml
```diff
@@ -5,6 +5,4 @@ test:
   adapter: test
 
 production:
-  adapter: redis
-  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
-  channel_prefix: dradispro_production
+  adapter: async
```
etdsoft, line 8: are we better off w/ 2 different files than with a `defined?(Dradis::Pro)` conditional?

Would :async blow up if we provide :url and :channel_preffix? If not, and we can getaway with a single line conditional for the adapter, we may be better off than keeping two different versions of the file in sync

config/initializers/_preprovision_database.rb
```diff
@@ -0,0 +1,21 @@
+# In App Platforms, assets:precompile is run before the DB is provisioned causing a
```
etdsoft, line 1: Isn't this something that belongs in dradis-plugins?

.do/deploy.template.yaml
```diff
@@ -9,8 +9,10 @@ spec:
     - key: RAILS_SERVE_STATIC_FILES
       value: enabled
       scope: RUN_TIME
+    - key: HEROKU_DEBUG_RAILS_RUNNER
+      value: '1'
```
etdsoft, line 13: should this be here?

PR #1058, opened 2022-08-03
Fix issue evidence form

aapomm:
### Spec
When creating an Evidence for an Issue via the Evidence tab in the Issue details and not selecting a node, the form submits correctly and a flash message "Evidence added for selected nodes." appears but no Evidence is created.

**Proposed solution**
There needs to be validation in both the frontend and backend. In the frontend, the form should not be submittable if no nodes are selected. In the backend, the EvidenceController#create_multiple action should be updated to handle if the node params are missing, and if so, re-render the form with an error message.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/controllers/evidence_controller.rb
+++ b/app/controllers/evidence_controller.rb
@@ -50,7 +50,7 @@ class EvidenceController < NestedNodeResourceController
     issue = current_project.issues.find(evidence_params[:issue_id])
 
     if node_params_empty?
-      redirect_to project_issue_path(current_project, evidence_params[:issue_id], tab: 'evidence-tab'), alert: 'Evidence cannot be created.'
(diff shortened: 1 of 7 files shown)


app/controllers/evidence_controller.rb
```diff
@@ -49,6 +49,11 @@ def create_multiple
     # validate Issue
     issue = current_project.issues.find(evidence_params[:issue_id])
 
+    if node_params_empty?
+      redirect_to project_issue_path(current_project, evidence_params[:issue_id], tab: 'evidence-tab'), alert: 'Evidence cannot be created.'
```
etdsoft, line 53: We know the error was around Node information, can we provide a better error here?

PR #1061, opened 2022-08-10
Holiday

MattBudz:
### Summary

Holiday fun

### Check List


app/helpers/holiday_helper.rb
```diff
@@ -0,0 +1,5 @@
+module HolidayHelper
+  def december?
```
aapomm, line 2: I'm not sure if we need this helper if we're only using it for just 2 instances. If we're planning to add more methods, then I can see it. Otherwise, I think it's unnecessary for now.

MattBudz: Fair, I'll move this to the `application_helper` since this method is used by the application layout. In CE we could've gotten away with having this as an instance variable in the sessions controller since only the login view uses the application layout. In Pro, we have multiple views that use this layout (login, password resets, new users sign up, new license, etc) so I think having it in the application helper makes sense.

PR #1067, opened 2022-09-06
dockerize dradis-ce

alexrupom:
### Summary

Dockerized dradis-ce 

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


bin/setup
```diff
@@ -52,7 +52,7 @@ FileUtils.chdir APP_ROOT do
   puts "\n== Removing old logs and tempfiles =="
   system! 'bin/rails log:clear tmp:clear'
 
-  puts "\n== Restarting application server =="
-  # system! 'bin/rails restart'
-  system! 'bin/rails server -p 8080'
+  # puts "\n== Restarting application server =="
+  # # system! 'bin/rails restart'
+  # system! 'bin/rails server -p 8080'
```
alexrupom, line 57: I have commented this part to reuse it in Dockerfile

sean-yeoh: What do you mean by "to re-use in Dockerfile"?

I don't think we should comment this out as it'll have implications to our 1 click Digital Ocean deploy https://github.com/dradis/dradis-ce/blob/develop/.do/deploy.template.yaml#L16

What about adding another bin file but for Docker? See: https://github.com/dradis/dradis-ce/blob/develop/bin/heroku

.dockerignore
```diff
@@ -0,0 +1,45 @@
...
+# RSpec testing
+/spec/.examples.txt
+
+# App-specific
+/attachments/
+/backup/
+/config/shared/
+/templates/
+/sample_files/
+/Gemfile.plugins
```
sean-yeoh, line 34: What's the purpose of this file and why is it ignoring Gemfile.plugins?

db/seeds.rb
```diff
@@ -1,4 +1,4 @@
 # Create a few default tags.
 Tag::DEFAULT_TAGS.each do |name|
-  Tag.create!(name: name)
+  Tag.find_or_create_by(name: name)
```
alexrupom, line 3: We are running `rails db:seed` when building the docker image. In case there is existing sqlite3 and tags are already created in that DB we will get error of uniqueness for tags. To avoid uniqueness error  we have changed `create!` to `find_or_create_by`. We can also remove 'bang(!)' to avoid error. But I think `find_or_create_by` is more appropriate in this scenario.

Taragos: But wouldn't running db:seed require access to the DB during the build process?
I think another option would be to construct a secondary-image or script that takes over the seeding-process and whatever is needed before the first start of the application.

PR #1069, opened 2022-09-12
Tag CRUD

alexrupom:
### Summary

When working on a project, we want to be able to CRUD the list of Tags that are available, so we can change Tag colors and names through the UI

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry


app/views/layouts/tylium.html.erb
```diff
@@ -102,6 +102,10 @@
       <%= render partial: 'styles_tylium/modal' %>
     <% end %>
 
+    <% if controller_name == 'tags' || controller_name == 'issues'%>
```
MattBudz, line 105:
```suggestion
    <% if controller_name == 'tags' || controller_name == 'issues' %>
```

app/views/tags/_table.html.erb
```diff
@@ -0,0 +1,56 @@
...
+                data: {
+                  sort: sort
+                }
+              %>
+            <% end  %>
+            <td class="column-actions">
+              <%= link_to edit_project_tag_path(current_project, tag), remote: true do %>
+                <i class="fa fa-pencil"></i> Edit
+              <% end %>
+              <%= link_to project_tag_path(current_project, tag), method: :delete, data: { confirm: "Are you sure?\n\nProceeding will delete this tag and any associated tagging." }, class: 'text-error-hover' do %>
```
MattBudz, line 47: let's make the confirmation message less scary and more clear

app/views/tags/edit.js.erb
```diff
@@ -0,0 +1,2 @@
+$('#modal-tag-form').modal('show');
```
MattBudz, line 1: Let's use a data-attribute for selecting the element rather than the ID. Please review the Frontend Development Guidelines on BC.

app/views/tags/edit.js.erb
```diff
@@ -0,0 +1,2 @@
...
+$('#modal-tag-content').html("<%= j (render partial: 'form', locals: { header: 'Edit tag', url: project_tag_path(current_project, @tag),  method: :patch }) %>");
```
MattBudz, line 2: ditto

app/views/tags/index.html.erb
```diff
@@ -0,0 +1,32 @@
...
+    <%= render 'table' %>
+  <% else %>
+  <div class="empty-state mt-3">
+    <div class="empty-state-icon">
+      <%= image_tag "icons/#{rand(9)}.svg" %>
+    </div>
+    <div class="empty-state-body">
+      <div class="container">
+        <h4 class="empty-state-title">There are no tags created yet.</h4>
+        <p class="empty-state-body-text">Create and use tags to manage your issues.</p>
```
MattBudz, line 23:
```suggestion
        <p class="empty-state-body-text">Use tags to flag and/or organize your issues.</p>
```

app/views/tags/index.html.erb
```diff
@@ -0,0 +1,32 @@
...
+          <i class="fa fa-plus fa-fw"></i> New Tag
+        <% end %>
+      </div>
+    </span>
+  </h4>
+
+  <% if @tags.any? %>
+    <%= render 'table' %>
+  <% else %>
+  <div class="empty-state mt-3">
```
MattBudz, line 16: indentation

app/views/tags/modals/_form.html.erb
```diff
@@ -0,0 +1,12 @@
+<div
+  id="modal-tag-form"
+  class="modal fade tags-modal"
+  tabindex="-1"
+  role="dialog"
+  aria-labelledby="modal-tag-header"
+  aria-hidden="true"
+>
+  <div class="modal-dialog">
+    <div id='modal-tag-content'class="modal-content"></div>
```
MattBudz, line 10:
```suggestion
    <div id="modal-tag-content" class="modal-content"></div>
```

app/views/tags/new.js.erb
```diff
@@ -0,0 +1,2 @@
+$('#modal-tag-form').modal('show');
```
MattBudz, line 1: same deal here, let's use a data-attribute to select the element

app/views/tags/new.js.erb
```diff
@@ -0,0 +1,2 @@
...
+$('#modal-tag-content').html("<%= j (render partial: 'form', locals: { header: 'Add new tag', url: project_tags_path(current_project),  method: :post }) %>");
```
MattBudz, line 2: ditto

MattBudz (changes requested):

app/models/tag.rb
```diff
@@ -22,20 +22,25 @@ class Tag < ApplicationRecord
   before_save :normalize_name
 
   # -- Validations ------------------------------------------------------------
-  validates :name, presence: true, uniqueness: { case_sensitive: false }
+  validates :name,
+  presence: true,
+  uniqueness: { case_sensitive: false } ,
+  format: {
+    with: /\A(!\h{6})_[a-zA-Z]+?\z/, message: 'is invalid: Special character and number are not permitted'
```
MattBudz, line 29:
```suggestion
    with: /\A(!\h{6})_[a-zA-Z]+?\z/, message: 'is invalid: Numbers and special characters are not permitted.'
```

app/views/tags/_table.html.erb
```diff
@@ -0,0 +1,58 @@
...
+                data: {
+                  sort: sort
+                }
+              %>
+            <% end  %>
+            <td class="column-actions">
+              <%= link_to edit_project_tag_path(current_project, tag), remote: true do %>
+                <i class="fa fa-pencil"></i> Edit
+              <% end %>
+              <%= link_to project_tag_path(current_project, tag), method: :delete, data: { confirm: "Are you sure?\n\nProceeding will untag any associated issues and delete this tag." }, class: 'text-error-hover' do %>
```
MattBudz, line 49:
```suggestion
              <%= link_to project_tag_path(current_project, tag), method: :delete, data: { confirm: "Are you sure?\n\nProceeding will delete this tag. Any issues with this tag will be untagged." }, class: 'text-error-hover' do %>
```

app/views/tags/edit.js.erb
```diff
@@ -0,0 +1,2 @@
+$('[data-behavior="modal-tag-form"]').modal('show');
+$('[data-behavior="modal-tag-content"]').html("<%= j (render partial: 'form', locals: { header: 'Edit tag', url: project_tag_path(current_project, @tag),  method: :patch }) %>");
```
MattBudz, line 2:
```suggestion
$('[data-behavior~=modal-tag-form]').modal('show');
$('[data-behavior~=modal-tag-content]').html("<%= j (render partial: 'form', locals: { header: 'Edit tag', url: project_tag_path(current_project, @tag),  method: :patch }) %>");
```

We like to use `~=` rather than `==` because we can stack additional behaviors in the view but only use the ones we need in JS. For example, a view could have `data-behavior="123 ABC"` but in the JS that only cares about `ABC` we can use `$('[data-behavior~=ABC]')`.

app/views/tags/new.js.erb
```diff
@@ -0,0 +1,2 @@
+$('[data-behavior="modal-tag-form"]').modal('show');
+$('[data-behavior="modal-tag-content"]').html("<%= j (render partial: 'form', locals: { header: 'Add new tag', url: project_tags_path(current_project),  method: :post }) %>");
```
MattBudz, line 2:
```suggestion
$('[data-behavior~=modal-tag-form]').modal('show');
$('[data-behavior~=modal-tag-content]').html("<%= j (render partial: 'form', locals: { header: 'Add new tag', url: project_tags_path(current_project),  method: :post }) %>");
```

app/views/layouts/tylium.html.erb
```diff
@@ -102,6 +102,10 @@
       <%= render partial: 'styles_tylium/modal' %>
     <% end %>
 
+    <% if controller_name == 'tags' || controller_name == 'issues' %>
```
MattBudz, line 105:
```suggestion
    <% if %w[tags issues].include?(controller_name) %>
```

Let's DRY this up

MattBudz (changes requested):

app/views/tags/edit.js.erb
```diff
@@ -0,0 +1,2 @@
+$('[data-behavior~=modal-tag-form]').modal('show');
```
etdsoft, line 1: should we guard this so this code doesn't run in every single view?

app/views/tags/new.js.erb
```diff
@@ -0,0 +1,2 @@
+$('[data-behavior~=modal-tag-form]').modal('show');
```
etdsoft, line 1: same for this code.

spec/factories/tags.rb
```diff
@@ -1,5 +1,5 @@
 FactoryBot.define do
   factory :tag do
-    sequence(:name){ |n| "tag-#{n}" }
+    name { "!#{Random.bytes(3).unpack1('H*')}_tag" }
```
etdsoft, line 3: this seems like a much more expensive way to build a name, I'm curious about the thinking behind it?

app/models/tag.rb
```diff
@@ -22,20 +22,25 @@ class Tag < ApplicationRecord
...
   # -- Class Methods ----------------------------------------------------------
 
-
   # -- Instance Methods -------------------------------------------------------
 
   # Returns a version of the tag's name suitable to present to the user:
   #  * If the tag name contains color details, they are stripped
   #  * The result is titleized
   def display_name()
+    return '' if self.name.nil?
```
etdsoft, line 42: This seems like a validation failure? Under what conditions would we be having a `#display_name` call, on a Tag that has a `nil` name?

app/models/tag.rb
```diff
@@ -51,7 +56,9 @@ def display_name()
   # Strips the tag's name and returns the color details if present
   # if no color information is found, returns a default value of #ccc
   def color()
-    name[/\A(!\h{6})_[[:word:]]+?\z/,1].try(:gsub, "!", "#") || "#555"
+    return '' if self.name.nil?
```
etdsoft, line 59: ditto here, how do we get to this situation, and can it be prevented w/ ActiveRecord validation?

app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+      redirect_to project_tags_path(current_project), alert: 'Tag deleted.'
+    else
+      redirect_to project_tags_path(current_project), alert: @tag.errors.full_messages.join('; ')
+    end
+  end
+
+  private
+
+  def tag_params
+    modified_params = params.require(:tag).permit(:display_name, :color)
```
etdsoft, line 46: There's a bit of inconsistency with the names of the Tag attributes:

We have Tag/Name/Display Name on the one hand, and Color code / Color on the other.

I think we should stick with a consistent set (e.g. Name/Color).

If Display_name is a "helper" method in the model, let's not use it as an HTTP param. and if we're calling it Name in the modal, let's call it Name in the table too.

app/views/tags/edit.js.erb
alexrupom: We have added a guard in [here](https://github.com/dradis/dradis-ce/pull/1069/files#diff-0ce3ccc401d82b8c2668f62ba39540adf2f26f57b33038c8f9ba27b0117501fe) which make sure that modal-tag-form is only rendered in tags and issues controller views. And edit.js.erb file is only rendered when we have a request in tags_controller#edit.

spec/factories/tags.rb
alexrupom: My previous implementation was `sequence(:name){ |n| "!00000#{n}"tag }` which was not generating the correct format of hex code and hence specs were failing. Therefore I have moved to `Random.bytes(3).unpack1('H*')` which will surely create a unique and ensures spec failure for uniqueness. I will be looking for an inexpensive solution.

app/models/tag.rb
alexrupom: We needed this condition for the tag form when there is a new record of the tag. In tag form when `#display_name` or `#color` fields try to retrieve its value from `Tag.new` have the name value of nil we get an error of `nil:NilClass`. In order to prevent that error from happening we added a nil guard. 

I have refactored my implementation, now we are checking if the @tag object is new_record? then using `#display_name` or `#color` to retrieve the values if that's not the case then we are setting empty string('') as the value for these fields.

app/models/tag.rb
alexrupom: Same reason as `#display_name`

I have refactored my implementation

app/controllers/tags_controller.rb
alexrupom: refactored

spec/factories/tags.rb
alexrupom: I have found an alternative solution `sequence(:name) { |n| "!#{"%06d" % n}_tag" }` to generate hex code, but we have a limitation for this we can generate hex code from `000000 to 999999` after that we will get failed specs because the hex code will exceed 7 digits and will get a validation error. But I do not think we have to worry about that limitation because we won't be generating up to 999999 tags in RSpec. Thoughts?

PR #1070, opened 2022-09-12
Lint changed files using rubocop for pull requests using GH Actions

caitmich:
### Summary

Instead of running `rubocop` as a step in the Codeship CI pipeline, we will use GitHub Actions to perform our rubocop check on changed files only. We will do this by comparing to `origin/develop`
If there are newly introduced rubocop warnings, the workflow run fails. If not, it passes and you are good to go! Code will not be merged without a passing rubocop step

❗️**Note:** This will need to be pushed along with a change to our current Codeship pipeline step to avoid having the pipeline fail 

Pro Sync: [https://github.com/secroots/dradis-pro/pull/1765](https://github.com/secroots/dradis-pro/pull/1765)

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry


caitmich:
When we push this, we will need to ensure that we remove the rubocop-ci step from the Codeship pipeline, as this will take care of that step, (and that step will continuously fail bc of the reference to origin/head)

.github/workflows/rubocop.yml
```diff
@@ -0,0 +1,20 @@
+name: Lint
+
+on:
+  pull_request:
+    branches:
+      - '*'
+
```
MattBudz, line 7: If we only do pull_request I believe the action will only run on the first commit to the branch and on merges. Adding `push` will also run in the cop on subsequent code sommits

.rubocop.yml
```diff
@@ -1,8 +1,9 @@
 AllCops:
-  TargetRubyVersion: 2.4
+  TargetRubyVersion: 3.0
```
MattBudz, line 2: if you remove this line then rubocop will use `.ruby-version` (probably cleaner as we won't have to update this every time we bump ruby).

.rubocop.yml
```diff
@@ -130,5 +127,4 @@ Layout/EndAlignment:
 
 # Use my_method(my_arg) not my_method( my_arg ) or my_method my_arg.
 Lint/RequireParentheses:
-  Enabled: true
-
+  Enabled: true
```
MattBudz, line 130: Missing EOF new line

.github/workflows/rubocop.yml
```diff
@@ -0,0 +1,20 @@
+name: Lint
+
+on:
+  pull_request:
+    branches:
+      - '*'
```
MattBudz, line 6:
```suggestion
on:
  pull_request:
    branches:
      - '*'
  push:
    branches:
      - '*'

```

If we only do pull_request I believe the action will only run on the first commit to the branch and on merges. Adding `push` will also run in the cop on subsequent code commits

.rubocop.yml
MattBudz, line 2:
```suggestion

```

if you remove this line then rubocop will use .ruby-version (probably cleaner as we won't have to update this every time we bump ruby).

.rubocop.yml
MattBudz, line 130:
```suggestion
  Enabled: true

```

Missing EOF new line

caitmich: Thanks, I have since updated my vscode settings so it will now do it automatically!

.rubocop.yml
caitmich: Made this change, but we have to update rubocop because the current `1.13.0` doesn't support ruby version 3.1.2

PR #1088, opened 2022-12-09
Validate move param

alexrupom:
### Summary

**Spec**
Users can hide methodology lists without removing the content (hiding it) by providing an invalid or empty “previous id”.
We are considering having server validation although this bug can not be generated from UI because in the past we had security reports that prioritized server validation. 

**Proposed solution**
Ensure that the prev_id value is checked and validated before processing it

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry


app/controllers/concerns/validate_move.rb
```diff
@@ -0,0 +1,15 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :validate_previous_id, only: :move
+  end
+
+  protected
+
+  def validate_previous_id
+    return true if params[:next_id].present? && params[:prev_id].blank?
```
aapomm, line 11: This doesn't work when moving a card to a list with no other cards since both `next_id` and `prev_id` is nil.

app/controllers/concerns/validate_move.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    head :unprocessable_entity unless controller_name.classify.constantize.exists? params[:prev_id]
```
aapomm, line 13: This line exceeds the [maximum line length](https://github.com/rubocop/ruby-style-guide#maximum-line-length). I suggest you configure your editor to show the 80 character limit.

It would also help the readability if you move `controller_name.classify.constantize` to a separate method.

app/controllers/cards_controller.rb
```diff
@@ -50,14 +54,9 @@ def update
...
-    List.move(
-      @card,
-      prev_item: @board.cards.find_by(id: params[:prev_id]),
-      next_item: @board.cards.find_by(id: params[:next_id])
-    )
-
-    if params[:new_list_id]
-      @card.list = @board.lists.find(params[:new_list_id])
+    List.move(@card, prev_item: @prev_item, next_item: @next_item)
+    if move_params[:new_list_id]
```
caitmich, line 58: This is a repeat of setting @parent below, we can DRY this up

alexrupom: We cannot have both operations together in a single 'if' condition because we need to perform setting @parent before 'List.move' and update the new_list of a card after 'List.move' is performed. Performing setting @parent before 'List.move' is required because validate_move_params method is dependent on that. Updating the new_list of a card after 'List.move' is required to prevent breaking the linked list, which will happen when we move the head card of one list to another list and update the new_list is performed before 'List.move'. That is why we need to maintain this specific order and cannot DRY this up more. Please let me know if you have any suggestions.

app/controllers/cards_controller.rb
```diff
@@ -10,6 +10,10 @@ class CardsController < AuthenticatedController
   before_action :set_or_initialize_card
   before_action :initialize_sidebar, only: [:show, :new, :edit]
   before_action :set_auto_save_key, only: [:new, :create, :edit, :update]
+  before_action :set_next_item_and_prev_item, only: :move
+  before_action :set_parent, only: :move
+  # Not at top because we need prev item and next item set first
```
MattBudz, line 15:
```suggestion

  # Not at top because we need prev item and next item set first
```

app/controllers/cards_controller.rb
```diff
@@ -10,6 +10,10 @@ class CardsController < AuthenticatedController
...
+  include ValidateMove
```
MattBudz, line 16:
```suggestion
  before_action :set_parent, only: :move
  
  # Not at top because we need prev item and next item set first
  include ValidateMove
```

app/controllers/cards_controller.rb
```diff
@@ -50,14 +54,9 @@ def update
...
-      @card,
-      prev_item: @board.cards.find_by(id: params[:prev_id]),
-      next_item: @board.cards.find_by(id: params[:next_id])
-    )
-
-    if params[:new_list_id]
-      @card.list = @board.lists.find(params[:new_list_id])
+    List.move(@card, prev_item: @prev_item, next_item: @next_item)
+    if new_list?
+      @card.list = @board.lists.find(move_params[:new_list_id])
```
MattBudz, line 59:
```suggestion
    List.move(@card, prev_item: @prev_item, next_item: @next_item)
    
    if new_list?
      @card.list = @board.lists.find(move_params[:new_list_id])
```

app/controllers/lists_controller.rb
```diff
@@ -4,6 +4,10 @@ class ListsController < AuthenticatedController
 
   before_action :set_current_board
   before_action :set_list, only: [:show, :edit, :update, :destroy, :move]
+  before_action :set_next_item_and_prev_item, only: :move
+  before_action :set_parent, only: :move
+  # Not at top because we need prev item and next item set first
+  include ValidateMove
```
MattBudz, line 10:
```suggestion
  before_action :set_parent, only: :move
  
  # Not at top because we need prev item and next item set first
  include ValidateMove
```

app/models/concerns/acts_as_linked_list.rb
```diff
@@ -36,28 +36,51 @@ def list_item_name(name)
...
+                                end
+        return false unless is_valid_new_position
 
         # Update previous position
         next_item.update_attribute(:previous_id, item.previous_id) if next_item
 
         # Update new position
-        new_previous = new_position[:prev_item]
-        new_next = new_position[:next_item]
+        new_next.update_attribute(:previous_id, item.id) if new_next
```
alexrupom, line 77: The move method was returning nil when new_next was not present therefore I have moved update new_next here and update new_previous after that to ensure we always return true or false.

alexrupom:
After I ran rubocop on the files which got warnings it seems like the code become unreadable.
this is how the code looks like after rubocop is run at app/models/concerns/acts_as_linked_list.rb
Screenshot 2022-12-22 at 5 28 27 PM

app/controllers/concerns/validate_move.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def valid_move_params?
+    if move_params[:prev_id] == nil
+      if @parent.items.empty?
+        move_params[:next_id] == nil
+      else
+        if @parent.first_item
+          @next_item == @parent.first_item
+        else
```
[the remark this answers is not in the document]
alexrupom: Aaaah just got another comment 😄 . I think I have addressed your review

app/controllers/cards_controller.rb
caitmich: Your change of adding `new_list` looks good to me

app/models/concerns/acts_as_linked_list.rb
```diff
@@ -36,28 +36,51 @@ def list_item_name(name)
...
+        # Validates new position
+        is_new_position_valid = if new_previous.present?
+                                  next_item_of_new_previous = new_previous.send("next_#{new_previous.class.name.downcase}")
+                                  if next_item_of_new_previous
+                                    new_next == next_item_of_new_previous
+                                  else
+                                    new_next.nil?
+                                  end
+                                else
+                                  if parent.items.empty?
```
caitmich, line 61: This still hasn't addressed [this comment](https://github.com/dradis/dradis-ce/pull/1088#discussion_r1053775107). If `parent.items.empty?` resolves to `true`, then `if parent.first_item` will also always resolve to false, no? I added a suggestion for a change in that comment linked above

alexrupom: Aaah makes sense. I will do that change

PR #1116, opened 2023-03-03
Add state to issue form and controller handling

aapomm:
### Spec
We need a way to see/set/edit the states for new and existing issues.

**Proposed solution**

- Update the issue form to allow setting the state
- Show the state in issue#show
- Add a state column in issue#index table

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/views/issues/_table.html.erb
+++ b/app/views/issues/_table.html.erb
@@ -1,5 +1,5 @@
 <% cache ['issues-table', @all_columns, issues.map(&:id), @issues.map(&:updated_at).map(&:to_i).sort.last, @tags] do %>
-    <table class="table table-striped mb-0"
+  <table class="table table-striped mb-0"
       data-behavior="dradis-datatable"
       data-default-columns="<%= @default_columns.to_json %>"
       data-item-name="issue"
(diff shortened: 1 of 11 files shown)


app/views/issues/_table.html.erb
```diff
@@ -8,65 +8,66 @@
...
+                  [
+                    issue.affected_count,
+                    issue.affected.map { |node| link_to(node.label, [current_project, node]) }.join("\n").html_safe
+                  ]
+                when 'Created'
+                  [issue.created_at.to_i, local_time_ago(issue.created_at)]
+                when 'Created by'
+                  [issue.author, issue.author]
+                when 'State'
+                  [issue.state, issue.state.humanize]
```
MattBudz, line 46: These are the only two lines that actually changed in this file (the rest is linting)

app/views/issues/edit.html.erb
```diff
@@ -10,12 +10,11 @@
...
 <% end %>
 
-<div class="content-container">
-  <div id="issues_editor">
-    <div class="note-text-inner">
-      <h4 class="header-underline">Edit issue</h4>
-      <div class="autofocus" autofocus="true" tabindex="1"></div> <!-- This div is needed to allow correct tabbing order in Chrome with page is loaded with turbolinks -->
-      <%= render 'form' %>
-    </div>
+<div class="content-container" id="issues_editor">
```
etdsoft, line 13: @MattBudz make sure this change doesn't have any JS implications before merging, I distrust the existence of `id="issues_editor"` and would like to make sure we understand why is it there, and whether it's safe to move it.

PR #1121, opened 2023-03-09
QA: show view

MattBudz:
### Summary

Add QA::Index#show

### Check List


app/assets/stylesheets/tylium/base.scss
```diff
@@ -216,6 +222,10 @@ strong {
   color: $primaryColor !important;
 }
 
+.top-5 {
+  top: 5rem !important;
+}
```
MattBudz, line 227: This is the only change in this file, the rest is linting

app/assets/stylesheets/tylium/modules.scss
```diff
@@ -199,12 +199,15 @@
...
 .nav-tabs {
-
   .nav-link:hover:not(.active) {
     border-color: transparent;
   }
 }
 
+.p-static {
+  position: static;
+}
```
MattBudz, line 209: This is the only change in this file, the rest is linting

app/views/issues/_form.html.erb
```diff
@@ -32,6 +32,9 @@
   <%= render 'tag_input', f: f %>
 
   <div class="form-actions">
+    <% if params[:redirect_to] %>
+      <%= hidden_field :issue, :redirect_to, value: params[:redirect_to] %>
```
aapomm, line 36:
```suggestion
      <%= hidden_field :redirect_to, value: params[:redirect_to] %>
```

No need to nest this under `:issue`. Note that we're passing all values under `:issue` as attributes for the issue record.

To illustrate, if nested under issue, the params will look like:
```
{
  issue:  {
    text: '...',
    redirect_to: 'somewhere'
  }
}
```
the controller will try to update the issue's `redirect_to` method. It needs to be:
```
{
  issue:  {
    text: '...'
  },
  redirect_to: 'somewhere'
}
```

app/views/shared/_sidebar_collection.html.erb
```diff
@@ -3,8 +3,8 @@
     <div class="options">
       <%= render 'shared/add_item_dropdown',
         name: name,
-        new_path: new_path,
-        category_id: category_id,
+        new_path: local_assigns[:new_path] ? new_path : nil,
```
aapomm, line 6: Does this work
```suggestion
        new_path: local_assigns[:new_path],
```

app/views/shared/_sidebar_collection.html.erb
```diff
@@ -3,8 +3,8 @@
...
+        category_id: local_assigns[:category_id] ? category_id : nil,
```
aapomm, line 7: Ditto

app/views/shared/_sidebar_collection.html.erb
```diff
@@ -14,7 +14,7 @@
 
 <div id="<%= name.downcase %>" class="note-list collapse show">
   <% if collection.any? %>
-    <%= render partial: collection.first.to_partial_path, collection: collection %>
+    <%= render partial: collection.first.to_partial_path, collection: collection, locals: { qa: local_assigns[:qa].present? } %>
```
aapomm, line 17:
```suggestion
    <%= render partial: collection.first.to_partial_path, collection: collection, locals: { qa: local_assigns[:qa] } %>
```

config/routes.rb
```diff
@@ -96,7 +96,11 @@
     resources :tags, except: [:show]
 
     namespace :qa do
-      resources :issues, only: [:index, :show, :edit]
+      resources :issues, only: [:index, :show] do
+        member do
+          patch :update_state
```
aapomm, line 101: Can we use the default `:update` route/action?

app/controllers/issues_controller.rb
```diff
@@ -183,7 +195,7 @@ def set_or_initialize_tags
   end
 
   def issue_params
-    params.require(:issue).permit(:state, :tag_list, :text)
+    params.require(:issue).permit(:state, :tag_list, :text, :redirect_to)
```
aapomm, line 198: No need for this if we're not nesting the `redirect_to` param under `:issue`

```suggestion
    params.require(:issue).permit(:state, :tag_list, :text)
```

app/controllers/issues_controller.rb
```diff
@@ -129,6 +133,14 @@ def set_affected_nodes
                           .sort_by { |node, _| node.label }
   end
 
+  def set_form_cancel_path
+    @form_cancel_path, = if params[:redirect_to]
+      params[:redirect_to]
+    else
+      @issue.new_record? ? project_issues_path(current_project) : project_issue_path(current_project, @issue)
+    end
```
aapomm, line 141:
```suggestion
    path = @issue.new_record? ? project_issues_path(current_project) : project_issue_path(current_project, @issue)
    @form_cancel_path, = params[:redirect_to] || path
```

app/controllers/issues_controller.rb
MattBudz: fixed

app/views/issues/_form.html.erb
MattBudz: got it, made the change

app/views/shared/_sidebar_collection.html.erb
MattBudz: yep, updated.

app/views/shared/_sidebar_collection.html.erb
MattBudz: fixed

app/views/shared/_sidebar_collection.html.erb
MattBudz: fixed

config/routes.rb
MattBudz: yup, changed to use `updated`

app/controllers/issues_controller.rb
MattBudz: implemented

app/controllers/issues_controller.rb
```diff
@@ -82,7 +83,14 @@ def update
         @modified = true
         check_for_edit_conflicts(@issue, updated_at_before_save)
         track_updated(@issue)
-        format.html { redirect_to project_issue_path(current_project, @issue), notice: 'Issue updated' }
+        format.html do
+          redirect_path = if params[:redirect_to] && URI.parse(params[:redirect_to]).relative?
```
aapomm, line 87: Instead of calling `params[:redirect_to] && URI.parse(params[:redirect_to]).relative?` multiple times, can we use a before_action cal instead? Something like:

```
def set_redirect_to
  @redirect_to =
    if URI.parse(params[:redirect_to]).relative?
      params[:redirect_to]
    else
      raise 'Something is fishy...'
    end
end
```

Then we can be sure `@redirect_to` is a local path and use it where we need to

app/controllers/qa/issues_controller.rb
```diff
@@ -1,8 +1,36 @@
...
+      redirect_to project_qa_issues_path(current_project), notice: 'State updated successfully.'
+    else
+      render :show, alert: @issue.errors.full_messages.join('; ')
+    end
+  end
+
+  private
+
+  def issue_params
+    params.permit(:state)
```
aapomm, line 26: As per Daniel's comment in the bulk update branch, we have to make sure that the state here is valid Draft/Ready For Review/Published.

app/models/issue.rb
```diff
@@ -12,6 +12,8 @@ class Issue < Note
 
   enum state: [:draft, :ready_for_review, :published]
 
+  attr_accessor :redirect_to
```
aapomm, line 15: This is a smell. We shouldn't need to add a virtual attribute

app/controllers/issues_controller.rb
MattBudz: This has actually changed in the #edit view PR and it's only called once in the controller. 

Leaving this as is here to prevent a merge conflict when the #edit view PR is merged into this PR.

app/controllers/qa/issues_controller.rb
MattBudz: should this live here or should we have a custom validation on the model?

app/controllers/issues_controller.rb
etdsoft: First I thought, Rails 7 cuts this by default and requires :allow_other_host if you want to go out of your way. That seems also to be available in our 6.1

https://github.com/rails/rails/blob/6-1-stable/actionpack/lib/action_controller/metal/redirecting.rb

But then I also remember `#redirect_back` couldn't we generally use that to avoid having to manually handle this?

app/controllers/issues_controller.rb
```diff
@@ -129,6 +137,16 @@ def set_affected_nodes
                           .sort_by { |node, _| node.label }
   end
 
+  def set_form_cancel_path
```
etdsoft, line 140: I was today-years-old when I learned about `link_to "foo", :back 

> Using a :back Symbol instead of an options hash will generate a link to the referrer (a JavaScript back link will be used in place of a referrer if none exists). If nil is passed as the name the value of the link itself will become the name.

https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to

MattBudz: I used this previously but I don't like how it only works if you navigated to a page with clicks. For example, if we use :back in the cancel link on #edit, and a user is linked directly to #edit without navigating to it from another view or if they refresh the page while on it, the #edit view cancel link will just link to #edit and the user can't actually "cancel".

app/controllers/issues_controller.rb
MattBudz: If we use `#redirect_back` the user will submit the form and land back in #edit, since #update will redirect_back to #edit.

I tested the redirect without `URI.parse(params[:redirect_to]).relative?` and it seems like rails will allow the redirect to another host 

Instead of using `params[:redirect_to]`, we could approach this with `session[:return_to]` instead but regardless, I don't see a reliable solution without passing in the path to `session[:return_to]` as we do now with `params[:redirect_to]`

Alternatively, we can check `request.referrer` in the controller and set the redirect path based on that. This should eliminate the need for `params[:redirect_to]`

Thoughts? cc @aapomm

PR #1123, opened 2023-03-13
QA: Edit view

MattBudz:
### Summary

Add a dedicated QA show view

### Check List


app/controllers/qa/issues_controller.rb
```diff
@@ -12,6 +12,16 @@ def index
 
   def show; end
 
+  def edit
+    @form_cancel_path = if params[:redirect_to] && URI.parse(params[:redirect_to]).relative?
+      params[:redirect_to]
+    else
+      project_issue_path(current_project, @issue)
```
aapomm, line 19: What's the use case for redirecting to issues#show from QA? I'm thinking `@form_cancel_path` will always be QA::issues#show, correct?

MattBudz: yep, I missed updating this when I moved it over from IssuesController.

Updated now to `@form_cancel_path = project_issue_qa_path(current_project, @issue)`

app/views/issues/_table.html.erb
```diff
@@ -61,7 +61,7 @@
               %>
             <% end %>
             <td class="column-actions">
-              <%= link_to edit_project_issue_path(current_project, issue) do %>
+              <%= link_to local_assigns[:qa] ? edit_project_qa_issue_path(current_project, issue) : edit_project_issue_path(current_project, issue) do %>
```
etdsoft, line 64: It seems to me this will break the HTML fragment cache /cc @aapomm 

It'd be better if the relevant JS code updates this value, wouldn't it? Perhaps adding both links to the view / cache, and then having JS choose which one is loaded?

MattBudz: replied in BC

PR #1126, opened 2023-03-16
QA: Import issues

MattBudz:
### Summary

Add's state selection to issues#import table

### Check List


app/assets/stylesheets/shared/datatables.scss
```diff
@@ -137,6 +134,7 @@ div.dataTables_wrapper {
     }
 
     tr {
+      &:focus-within,
```
MattBudz, line 137: this line is the only change in this file, the rest is linting

etdsoft (approved):

PR #1131, opened 2023-03-23
QA export all

aapomm:
### Spec
We need a way for users to export all records regardless if they are set to published or not

**Proposed solution**

- In the export view, split the export button to support exporting all records or just published records (default)
- Pass the selected scope to the exporter
- In the content service, return all/published records depending on the passed scope

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/controllers/export_controller.rb
+++ b/app/controllers/export_controller.rb
@@ -103,9 +103,9 @@ class ExportController < AuthenticatedController
   def validate_scope
     return if @exporter.to_s.include?('::Projects')
 
-    if params[:scope] == 'all' || params[:scope] == 'published'
-      @scope = params[:scope]
-    else
+    @scope = params[@exporter::Engine::plugin_name.to_s][:scope]
+
+    unless @scope == 'all' || @scope == 'published'
       redirect_to project_export_manager_path(current_project), alert: 'Something fishy is going on...'
     end
   end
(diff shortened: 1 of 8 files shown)


app/controllers/export_controller.rb
```diff
@@ -98,4 +96,14 @@ def validate_template
       end
     end
   end
+
+  def validate_scope
+    return if @exporter.to_s.include?('::Projects')
+
+    @scope = params[@exporter::Engine::plugin_name.to_s][:scope]
```
aapomm, line 103:
```suggestion
    @scope = params[@exporter::Engine::plugin_name.to_s].fetch(:scope, :published)
```

app/controllers/export_controller.rb
```diff
@@ -4,22 +4,20 @@ class ExportController < AuthenticatedController
...
-    # *Warning* can't store too much data here.
-    session[:export_manager] = {
-      project_id: current_project.id,
-      template: @template_file
-    }
-
     # FIXME: check the Routing guide to find a better way.
     action_path = "#{params[:route]}_path"
-    redirect_to eval(@exporter::Engine::engine_name).send(action_path)
+    redirect_to eval(@exporter::Engine::engine_name).send(
```
aapomm, line 15:
```suggestion
    redirect_to send(@exporter::Engine::engine_name).send(
```

PR #1132, opened 2023-03-23
QA index: fix cache

MattBudz:
### Summary

Prevent the `issues/_table` cache from breaking by removing the conditional server-side rendering of elements. Instead, use JS to show/hide the correct elements.

### Check List


app/views/issues/_table.html.erb
```diff
@@ -61,13 +61,14 @@
               %>
             <% end %>
             <td class="column-actions">
-              <%= link_to local_assigns[:qa] ? edit_project_qa_issue_path(current_project, issue) : edit_project_issue_path(current_project, issue) do %>
+              <%= link_to  edit_project_issue_path(current_project, issue) do %>
```
aapomm, line 64:
```suggestion
              <%= link_to edit_project_issue_path(current_project, issue) do %>
```

app/assets/javascripts/tylium/modules/qa_visibility.js
```diff
@@ -0,0 +1,12 @@
+document.addEventListener('turbolinks:load', function () {
```
aapomm, line 1: `qa-viewer` seems like it's only present in the QA page. If that's the case, can we make this a QA specific JS and follow the naming convention? Something like `tylium/pages/qa/table.js`? Also instead of `qa-viewer` does it make sense to use `body.qa` instead?

app/assets/javascripts/tylium/modules/qa_visibility.js
```diff
@@ -0,0 +1,12 @@
...
+  if ($('[data-behavior~=qa-viewer]').length) {
+    $('[data-qa-visible]').each(function () {
+      if ($(this).data('qa-visible') === true) {
+        $(this).removeClass('d-none');
```
aapomm, line 5: Would jquery's `.toggleClass` work here instead of `removeClass`/`addClass`?

app/assets/javascripts/tylium/modules/qa_visibility.js
```diff
@@ -0,0 +1,12 @@
...
+        $(this).prev().addClass('d-none');
```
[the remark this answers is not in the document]
MattBudz: `prev` is the previous sibling. The HTML looks like this:
```html
<td>
  <a href="normal-view">Title</a>
  <a href="qa-view" data-qa-visible="true">Title</a>
</td>
```

When we show the qa-view link, we hide the sibling. 
```js
if ($(this).data('qa-visible') === true) {
  $(this).removeClass('d-none');
  $(this).prev().addClass('d-none');
```

This was done in an effort to eliminate the need for specifying both true and false:
```html
<td>
  <a href="normal-view" data-qa-visible="false">Title</a>
  <a href="qa-view" data-qa-visible="true">Title</a>
</td>
```

but looking at it now, I think it's more readable and it reduces the JS down to just
```js
$(this).toggleClass('d-none', $(this).data('qa-visible') === true);
```

I have gone ahead and implemented this change.

PR #1133, opened 2023-03-23
QA: Bulk update states from #index

MattBudz:
### Summary

Add the ability to bulk update states from the normal #index view.

### Check List

### DIFF
--- a/app/assets/javascripts/shared/datatable/state.js
+++ b/app/assets/javascripts/shared/datatable/state.js
@@ -1,59 +1,84 @@
 $(document).on('preInit.dt', 'body.issues.index', function (e, settings) {
   var $paths = $('[data-behavior~=datatable-paths]');
-  if ($paths.data('table-state-url') == undefined) { return; }
+  if ($paths.data('table-state-url') == undefined) {
+    return;
+  }
 
-  var api = new $.fn.dataTable.Api( settings );
+  var api = new $.fn.dataTable.Api(settings),
+    isQA = $('[data-behavior~=qa-viewer]').length > 0;
+
   api.button().add(1, {
     attr: {
-      'data-behavior': 'table-action'
+      'data-behavior': 'table-action',
     },
     autoClose: true,
     className: 'd-none',
     extend: 'collection',
     name: 'stateBtn',
     text: '<i class="fa fa-adjust fa-fw"></i>State<i class="fa fa-caret-down fa-fw"></i>',
-    buttons: DradisDatatable.prototype.setupStateButtons.call(api)
+    buttons: DradisDatatable.prototype.setupStateButtons.call(api, isQA),
   });
 });
 
 $(document).on('init.dt', 'body.issues.index', function (e, settings) {
   var $paths = $('[data-behavior~=datatable-paths]');
-  if ($paths.data('table-state-url') == undefined) { return; }
+  if ($paths.data('table-state-url') == undefined) {
+    return;
+  }
 
-  var api = new $.fn.dataTable.Api( settings );
-  api.on('select.dt deselect.dt', function() {
-    var selectedCount = this.rows({selected:true}).count();
-    DradisDatatable.prototype.toggleStateBtn.call(api, selectedCount !== 0);
-  }.bind(api));
+  var api = new $.fn.dataTable.Api(settings);
+  api.on(
+    'select.dt deselect.dt',
+    function () {
+      var selectedCount = this.rows({ selected: true }).count();
+      DradisDatatable.prototype.toggleStateBtn.call(api, selectedCount !== 0);
+    }.bind(api)
+  );
 });
 
-DradisDatatable.prototype.setupStateButtons = function() {
-  var states = [['Draft', 'fa-pencil-square-o'], ['Published', 'fa-rocket']],
+DradisDatatable.prototype.setupStateButtons = function (isQA) {
+  var states = [
+      ['Draft', 'fa-pencil-square-o'],
+      ['Ready for review', 'fa-eye'],
+      ['Published', 'fa-rocket'],
+    ],
     stateButtons = [],
     api = this;
 
-  states.forEach(function(state){
+  if (isQA) {
+    states.splice(1, 1);
+  }
+
+  states.forEach(function (state) {
     stateButtons.push({
-      text: $(`<i class="fa ${state[1]} fa-fw"></i><span>${state[0]}</span>`),
-      action: DradisDatatable.prototype.updateRecordState.call(api, state[0].toLowerCase())
+      text: $(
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/shared/datatable/state.js
```diff
@@ -1,59 +1,84 @@
 $(document).on('preInit.dt', 'body.issues.index', function (e, settings) {
   var $paths = $('[data-behavior~=datatable-paths]');
-  if ($paths.data('table-state-url') == undefined) { return; }
+  if ($paths.data('table-state-url') == undefined) {
+    return;
+  }
+
+  var api = new $.fn.dataTable.Api(settings),
+    isQA = $('[data-behavior~=qa-viewer]').length > 0;
```
aapomm, line 8: Can we define the `isQA` variable in the `setupStateButtons` method instead of passing it?

app/assets/javascripts/shared/datatable/state.js
```diff
@@ -1,59 +1,84 @@
...
 
-  var api = new $.fn.dataTable.Api( settings );
   api.button().add(1, {
     attr: {
-      'data-behavior': 'table-action'
+      'data-behavior': 'table-action',
```
aapomm, line 12: Are [trailing commas](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas) the convention?

app/controllers/qa/issues_controller.rb
```diff
@@ -20,11 +20,16 @@ def edit
...
   def update
-    @issues = current_project.issues.ready_for_review.where(id: params[:ids])
+    @issues = current_project.issues.where(id: params[:ids])
 
     respond_to do |format|
       if @issues.update_all(state: @state, updated_at: Time.now)
-        format.html { redirect_to project_qa_issues_path(current_project), notice: 'State updated successfully.' }
+        format.html do
+          # prevent attempting to redirect the user back to the show view when the QA widget is used. Record will not exist as it'll no longer be ready for review.
+          session[:return_to] = nil if session[:return_to] == project_qa_issue_url(current_project, current_project.issues.find(params[:ids]))
```
aapomm, line 29: Would this work:

```suggestion
          if session[:return_to] == project_qa_issue_path(current_project, params[:ids])
            session[:return_to] = nil
          end
```

spec/features/issues_spec.rb
```diff
@@ -16,10 +16,10 @@
     context 'with an Issue library' do
       let(:issuelib) { current_project.issue_library }
 
-      describe 'index page' do
+      describe 'index page', js: true do
         it 'presents a link to add new issue' do
           visit project_issues_path(current_project)
-          expect(page).to have_xpath("//a[@href='#{new_project_issue_path(current_project)}']")
+          expect(page).to have_xpath("//a[@href='#{new_project_issue_path(current_project)}']", visible: false)
```
aapomm, line 22: Why is the link hidden?

MattBudz: I'm not sure why this started to fail now because it's always been hidden. The user has to click the + sign first then a dropdown is revealed with links to a new issue (blank, default, templates)

aapomm: Ah I think it's due to adding js to the spec. In that case, this also needs to test the clicking of the + sign instead of artificially clicking a hidden link.

app/assets/javascripts/shared/datatable/state.js
MattBudz: It seems like it became the default when [ES5](https://www.w3schools.com/js/js_es5.asp) was released.

app/assets/javascripts/shared/datatable/state.js
```diff
@@ -1,59 +1,83 @@
...
+  );
 });
 
-DradisDatatable.prototype.setupStateButtons = function() {
-  var states = [['Draft', 'fa-pencil-square-o'], ['Published', 'fa-rocket']],
+DradisDatatable.prototype.setupStateButtons = function () {
+  var states = [
+      ['Draft', 'fa-pencil-square-o'],
+      ['Ready for review', 'fa-eye'],
+      ['Published', 'fa-rocket'],
```
etdsoft, line 42: Shouldn't we bake this as data- attributes from the model to avoid having to chase it if the collection changes in the future?

app/controllers/qa/issues_controller.rb
```diff
@@ -20,11 +20,19 @@ def edit
...
 
     respond_to do |format|
       if @issues.update_all(state: @state, updated_at: Time.now)
-        format.html { redirect_to project_qa_issues_path(current_project), notice: 'State updated successfully.' }
+        format.html do
+          # prevent attempting to redirect the user back to the show view when the QA widget is used.
+          # Record will not exist as it'll no longer be ready for review.
+          if session[:return_to] == project_qa_issue_url(current_project, params[:ids])
+            session[:return_to] = nil
+          end
```
etdsoft, line 32: This is a a smell and it seems like it doesn't belong to this action. Where does it belong to?

app/controllers/qa/issues_controller.rb
```diff
@@ -20,11 +20,19 @@ def edit
   end
 
   def update
-    @issues = current_project.issues.ready_for_review.where(id: params[:ids])
+    @issues = current_project.issues.where(id: params[:ids])
```
etdsoft, line 23: the use of :ids instead of :id, breaks the convention we use elsewhere in #update actions. Why is that the right call?

app/assets/javascripts/shared/datatable/state.js
MattBudz: Makes sense to keep it all together in the model, I have pushed this change.

app/controllers/qa/issues_controller.rb
MattBudz: I split this up into `update` and `bulk_update`. Now we have removed this completely

app/models/issue.rb
```diff
@@ -90,6 +90,21 @@ def self.autogenerate_from(evidence)
     )
   end
 
+  def self.state_icons
```
etdsoft, line 93: "icons" don't belong in a model. What about IssuesHelper?

config/routes.rb
```diff
@@ -96,8 +96,8 @@
     resources :tags, except: [:show]
 
     namespace :qa do
-      resources :issues, only: [:edit, :index, :show] do
-        collection { put :update }
+      resources :issues, only: [:edit, :index, :show, :update] do
+        collection { put :bulk_update }
```
etdsoft, line 100: I think it's time to bring Pro's :multiple_update  routing concern to CE. It'd also require a route/action rename from bulk_ to multiple_

PR #1134, opened 2023-03-24
QA: State change activity

MattBudz:
### Summary

Track state changes and render them in the activity feed

### Check List


app/presenters/activity_presenter.rb
```diff
@@ -125,9 +129,9 @@ def trackable_name
 
   def trackable_title
     @title ||= if activity.trackable.respond_to?(:title) && activity.trackable.title?
```
aapomm, line 131: What's the convention here? I'm thinking this should be in the next line like: https://github.com/dradis/dradis-ce/blob/develop/app/presenters/digest_presenter.rb#L56

MattBudz: updated

app/controllers/concerns/activity_tracking.rb
```diff
@@ -30,4 +30,8 @@ def track_recovered(trackable, user: current_user, project: current_project)
   def track_updated(trackable, user: current_user, project: current_project)
     track_activity(trackable, :update, user, project)
   end
+
+  def track_updated_state(trackable, user: current_user, project: current_project)
```
etdsoft, line 34: Don't love the name here, and looking at the presenter further down, what'd be the point of splitting the methods, and the action if we're not going to differentiate the view?

It seems to me that what we care about it's not the updating on an attribute, but the change in the QA state. Naming convention around QA state, and displaying a related message would be more useful

MattBudz: Updated the name to `track_state_change`. 
We are differentiating the view [here](https://github.com/dradis/dradis-ce/pull/1134/files#diff-99b696dd220f1c79d35442c8fce53b3177ba41ed9377401e1b725a715afa2cc5R3)

PR #1142, opened 2023-04-19
Fix QA redirect after changing state in QA Edit views

MattBudz:
### Summary

This PR fixes a bug where users got a 500 after changing a record's state from the QA #edit view.

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/controllers/qa/issues_controller.rb
+++ b/app/controllers/qa/issues_controller.rb
@@ -3,16 +3,19 @@ class QA::IssuesController < AuthenticatedController
   include ProjectScoped
 
   before_action :set_issues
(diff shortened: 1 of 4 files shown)


app/controllers/qa/issues_controller.rb
```diff
@@ -8,7 +8,6 @@ class QA::IssuesController < AuthenticatedController
   before_action :validate_state, only: [:multiple_update, :update]
 
   def index
-    @issues = current_project.issues.ready_for_review
```
MattBudz, line 11: this is redundant, we are already setting this in a before_action callback

PR #1145, opened 2023-05-05
Add liquid resource assigns to tylium

caitmich:
### Summary

Allow users to access LDC in Tylium 

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 [v#.#.#] ([month] [YYYY])
-  - [entity]:
-    - [future tense verb] [feature]
+  - Tylium:
+    - Add support for Liquid Dynamic Content
   - Upgraded gems:
     - [gem]
   - Bugs fixes:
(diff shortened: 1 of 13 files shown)


app/controllers/evidence_controller.rb
```diff
@@ -107,6 +107,13 @@ def autogenerate_issue
     track_created(@evidence.issue)
   end
 
+  def liquid_resource_assigns
+    {
```
aapomm, line 111: Should we add the IssueDrop for the evidence issue here too?

CHANGELOG
```diff
@@ -1,6 +1,6 @@
 [v#.#.#] ([month] [YYYY])
-  - [entity]:
-    - [future tense verb] [feature]
+  - Tylium:
+    - Add support for Liquid Dynamic Content
```
etdsoft, line 3:
```suggestion
    - Extend support for Liquid Dynamic Content
```

PR #1156, opened 2023-06-01
Fix export tabs

MattBudz:
### Summary

In Export Manager, show the first tab on page load, if no other tabs are shown.

### Check List

- [x] Added a CHANGELOG entry


MattBudz:
doing a js-less approach

PR #1172, opened 2023-08-23
Show Liquid content in editor preview

caitmich:
### Summary
We now have more support for Liquid content, but it does not yet show up in the preview when editing an evidence/issue/note

Using a Previewable concern, each resource will handle its own preview route so that the record's Drop is included when evaluating the Liquid

### Check List

- [x] Added a CHANGELOG entry


caitmich:
Handled in #1204

PR #1181, opened 2023-09-15
Fix qa redirect

aapomm:
### Spec
Currently, we're saving the session[:return_to] when the user visits the issue/content blocks page. This has unintended results since the session persists through different tabs.

**Proposed solution**
Pass a param from the QA form to indicate it coming from the QA view. If the param is received, redirect back to the QA #show if the record is still ready for review or #index if not.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


app/views/issues/_form.html.erb
```diff
@@ -29,6 +29,10 @@
     <%= hidden_field :issue, :original_updated_at, value: @issue.updated_at.to_i %>
   <% end %>
 
+  <% if local_assigns.has_key?(:qa) && qa %>
+    <%= hidden_field nil, :return_to, value: :qa %>
+  <% end %>
+
```
etdsoft, line 35: Isn't this flag something a) the controller should set; or b) we can infer from the local `controller` variable?

I don't think we need a local "view" variable that only is present some times... feels like overdesigning

app/controllers/issues_controller.rb
```diff
@@ -85,12 +84,17 @@ def update
...
+          if params[:return_to] == 'qa'
+            return_to =
+              if @issue.ready_for_review?
+                project_qa_issue_path(current_project, @issue)
+              else
+                project_qa_issues_path(current_project)
+              end
+            redirect_to_target_or_default return_to, notice: 'Issue updated.'
+          else
+            redirect_to_target_or_default project_issue_path(current_project, @issue), notice: 'Issue updated.'
```
etdsoft, line 96: this redirect logic can probably be extrated to a helper method?

app/assets/javascripts/shared/datatable/state.js
```diff
@@ -72,6 +72,7 @@ DradisDatatable.prototype.updateRecordState = function (newState) {
       data: {
         ids: DradisDatatable.prototype.rowIds(selectedRows),
         state: newState,
+        return_to: $('[data-behavior=qa-viewer]').length > 0 ? 'qa' : null
```
caitmich, line 75: Should this be `~=` to allow for a case where we add additional data-behavior attributes to the qa-viewer container in the future?

app/controllers/issues_controller.rb
```diff
@@ -135,6 +127,20 @@ def liquid_resource_assigns
     { 'issue' => IssueDrop.new(@issue) }
   end
 
+  def redirect_to_main_or_qa
+    notice = 'Issue updated.'
+
+    if params[:return_to] == 'qa'
+      if @issue.ready_for_review?
+        redirect_to_target_or_default project_qa_issue_path(current_project, @issue), notice: notice
```
caitmich, line 135: Do we need to use `redirect_to_target_or_default` here since we're no longer setting `return_to` in the session? What about using a regular `redirect_to`? (same on L 137 & 140)

PR #1185, opened 2023-09-21
Run the update command for Rails 7

caitmich:
### Summary

Run bin/rails app:update
Adds a migration, updates several config files and adds a new_framework_defaults file

### Check List

- [ ] Added a CHANGELOG entry


config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,8 @@
 # Be sure to restart your server when you modify this file.
 
-# Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password, :password_confirmation]
+# Configure parameters to be filtered from the log file. Use this to limit dissemination of
+# sensitive information. See the ActiveSupport::ParameterFilter documentation for supported
+# notations and behaviors.
+Rails.application.config.filter_parameters += [:password, :password_confirmation,
```
aapomm, line 6:
```suggestion
Rails.application.config.filter_parameters += [
  :password, :password_confirmation,
```

config/environments/production.rb
```diff
@@ -1,5 +1,7 @@
+require 'active_support/core_ext/integer/time'
```
aapomm, line 1: Do we need this? I don't see any lines requiring `time` unlike in test.rb/development.rb.

caitmich: You're right, the only line that uses this is commented out. Removed

PR #1192, opened 2023-10-06
Upgrade from turbolinks to turbo

caitmich:
### Summary

Remove turbolinks and use turbo-rails instead
Update references to turbolinks to use turbo instead

### Check List

- [ ] Added a CHANGELOG entry


app/views/layouts/tylium/_sidebar_header.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="sidebar-header">
-  <%= link_to root_path, class: 'd-flex' do %>
+  <%= link_to main_app.root_path, class: 'd-flex' do %>
```
caitmich, line 2: Unrelated but fixes an exception thrown when using CVSS

PR #1196, opened 2023-10-18
Table actions

jorisdemange:
### Summary

Currently, tables have an actions column on the far right. This comes with a few problems, most notably:
- The actions column becomes "hidden" unless users horizontally scroll over to the last column
- The actions column may be shown just enough to confuse users into thinking there is just one action
- The action column takes up too much white space when there are multiple actions

### Fix description

This solution is to make the far right column float, make it invisible by default and visible on row hover. All the actions are wrapped in a drop-down menu activated with an ellipsis button. This solution is scalable, is valid with accessibility standards, and is responsive. It prevents any overlay no matter the screen size.

### Hover effects alternatives

The request stated: 
> “the visual cue should always be […] visible […] regardless if the row is being hovered over or not”

The issue with this is visual clutter. Tables are inherently data intensive, especially when rows are compact. As the "more" button had to overlay the row, making all of them constantly visible would take some screen real estate, partially hide some data, and make the interface visually heavy. This is why I made them visible on hover only.

This can easily be reverted by deleting the lines `16` and `95-99` of `datatables.scss`.

### Testing instructions

- Navigate to [http://localhost:3000/projects/1/issues](http://localhost:3000/projects/1/issues) ("all issues" from the left menu)
- In the table that appears, activate more columns from the column dropdown to enable horizontal scroll
- Hover over any row to see the related "more" button
- Each "more" button has a dropdown menu containing all actions (currently Edit and Delete)

### Copyright assignment

### DIFF
--- a/app/assets/stylesheets/shared/datatables.scss
+++ b/app/assets/stylesheets/shared/datatables.scss
@@ -36,6 +36,10 @@ div.dataTables_wrapper {
       }
     }
 
+    a {
+      width: 100%;
+    }
+
     tbody {
       tr {
         .select-checkbox {
--- a/app/assets/stylesheets/shared/datatables.scss
+++ b/app/assets/stylesheets/shared/datatables.scss
@@ -5,6 +5,7 @@ div.dataTables_wrapper {
     .column-actions {
       position: sticky;
       right: 0;
+      opacity: 0;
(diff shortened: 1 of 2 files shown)


MattBudz:
thanks @jorisdemange closing this for now!

PR #1198, opened 2023-10-24
Upgrade from turbolinks to turbo

caitmich:
### Summary
Remove turbolinks and use turbo-rails instead
- Remove turbolinks from Gemfile
- Add turbo-rails to Gemfile
- Replace turbolinks namespaces with turbo
- Include turbo with Importmaps

### Check List

- [x] Added a CHANGELOG entry


app/javascript/application.js
```diff
@@ -1,3 +1,8 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
+import '@hotwired/turbo-rails'
 import 'datatables'
 import 'stupidtable'
+
+// Forms by default will not be submitted with turbo unless the
+// data-turbo="true" attribute is present. See: https://github.com/hotwired/turbo/pull/419
+Turbo.setFormMode("optin")
```
etdsoft, line 8: This sounds like something that's not true in the latest Turbo, probably have to either revisit or add a comment to delete this eventually

app/controllers/application_controller.rb
```diff
@@ -3,6 +3,7 @@
 
 class ApplicationController < ActionController::Base
   include Authentication
+  include Turbo::Redirection
```
etdsoft, line 6: This shouldn't be needed, smell alert! 🚨

PR #1201, opened 2023-11-14
Add global drops

aapomm:
### Summary

Add project records as available drops in LiquidEnabledResource

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/controllers/concerns/liquid_enabled_resource.rb
+++ b/app/controllers/concerns/liquid_enabled_resource.rb
@@ -33,6 +33,11 @@ module LiquidEnabledResource
     authorize! :use, project
 
     result['project'] = ProjectDrop.new(project)
+    result['issues'] = project.issues.map { |issue| IssueDrop.new(issue) }
+    result['notes'] = project.notes.map { |note| NoteDrop.new(note) }
+    result['nodes'] = project.nodes.map { |node| NodeDrop.new(node) }
+    result['evidence'] = project.evidence.map { |evidence| EvidenceDrop.new(evidence) }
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/liquid_enabled_resource.rb
```diff
@@ -33,6 +33,10 @@ def project_assigns
     authorize! :use, project
 
     result['project'] = ProjectDrop.new(project)
+    result['issues'] = project.issues.map { |issue| IssueDrop.new(issue) }
+    result['nodes'] = project.nodes.user_nodes.map { |node| NodeDrop.new(node) }
+    result['evidences'] = project.evidence.map { |evidence| EvidenceDrop.new(evidence) }
+    result['tags'] = project.tags.map { |tags| TagDrop.new(tags) }
```
caitmich, line 39:
```suggestion
    result['tags'] = project.tags.map { |tag| TagDrop.new(tag) }
```

spec/support/revision_shared_examples.rb
```diff
@@ -61,6 +61,11 @@
 
       expect do
         visit project_trash_path(current_project)
+
+        within '#trash' do
```
caitmich, line 65: unrelated?

aapomm: Sort of? But this change is needed because this example expects the instance to only have 1 of the item in the trash. If for some reason there are other items in the trash, the spec fails. In the case of the liquid shared examples, it creates multiple test records that break this example.

caitmich:
Not implemented

PR #1202, opened 2023-11-14
Implement inline diff for manually created findings

rachkor:
### Summary

This PR improves the existing version history to show diffs inline rather than wrapping around the entire changed line. 

### Other Information

This change will apply to manually-created findings, not tool-created findings. 

### Copyright assignment

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,8 @@
   - Upgrade to Rails 7.0.8
   - Replace unicorn with puma in production
   - Add importmap-rails to handle js libraries
+  - Tylium:
+    - Added inline diff to version history for manually created findings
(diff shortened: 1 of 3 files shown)


CHANGELOG
```diff
@@ -2,6 +2,8 @@
   - Upgrade to Rails 7.0.8
   - Replace unicorn with puma in production
   - Add importmap-rails to handle js libraries
+  - Tylium:
+    - Added inline diff to version history for manually created findings
```
aapomm, line 6: What about:
```suggestion
  - Revision history: Improve version history for content with carriage return
```

PR #1204, opened 2023-11-20
Liquid Preview

caitmich:
### Summary

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/controllers/concerns/previewable.rb
+++ b/app/controllers/concerns/previewable.rb
@@ -1,0 +1,8 @@
+module Previewable
+  extend ActiveSupport::Concern
+
+  def preview
+    @text = params[:text]
+    render 'markup/preview', layout: false
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -24,6 +24,12 @@ Rails.application.routes.draw do
     end
   end
 
+  concern :previewable do
+    member do
+      post :preview
+    end
+  end
+
   resources :notifications, only: [:index, :update]
 
   resources :projects, only: [:index, :show] do


app/controllers/notes_controller.rb
```diff
@@ -7,13 +7,16 @@ class NotesController < NestedNodeResourceController
...
 
-  before_action :find_or_initialize_note, except: [:index, :new, :multiple_destroy]
+  before_action :find_or_initialize_note, except: [:index, :new, :multiple_destroy, :preview]
   before_action :initialize_nodes_sidebar, only: [:edit, :new, :show]
   before_action :set_auto_save_key, only: [:new, :create, :edit, :update]
+  before_action :set_form_preview_path, only: [:create, :edit, :update]
 
   def new
     @note = @node.notes.new
+    set_form_preview_path
```
aapomm, line 19: Why are we manually calling it here instead of the before_action callback and can we make it so the callback works here?

app/controllers/evidence_controller.rb
```diff
@@ -6,11 +6,13 @@ class EvidenceController < NestedNodeResourceController
   include MultipleDestroy
   include NodesSidebar
   include NotificationsReader
+  include Previewable
 
-  before_action :set_or_initialize_evidence, except: [ :index, :create_multiple ]
+  before_action :set_or_initialize_evidence, except: [ :index, :create_multiple, :preview ]
```
aapomm, line 11: Can we update this to use `only:` ? That way we're using a safer approach (whitelist) and it's more consistent with the rest of the callbacks below.

app/controllers/concerns/previewable.rb
```diff
@@ -0,0 +1,8 @@
+module Previewable
+  extend ActiveSupport::Concern
+
```
aapomm, line 3: What about adding a `set_form_preview_path` method here that needs to be overridden? Seeing as it's needed for this concern to be used.

app/controllers/evidence_controller.rb
```diff
@@ -108,12 +110,18 @@ def autogenerate_issue
...
+    @evidence = @evidence || @node.evidence.find_by(id: params[:id])
     {
       'evidence' => EvidenceDrop.new(@evidence),
-      'node' => NodeDrop.new(@evidence.node)
+      'node' => NodeDrop.new(@node)
     }
   end
 
+  def set_form_preview_path
+    @form_preview_path = @evidence.new_record? ? {} : { preview_url: preview_project_node_evidence_path(current_project, @node, @evidence) }
```
aapomm, line 122: Can we break this down to something like
```suggestion
    @form_preview_path = if @evidence.new_record?
      {}
    else
      { preview_url: preview_project_node_evidence_path(current_project, @node, @evidence) }
    end
```

app/controllers/issues_controller.rb
```diff
@@ -149,6 +151,10 @@ def set_form_cancel_path
     @form_cancel_path = session[:return_to] ? session[:return_to] : path
   end
 
+  def set_form_preview_path
+    @form_preview_path = @issue.new_record? ? {} : { preview_url: preview_project_issue_path(current_project, @issue) }
```
aapomm, line 155: ditto

app/controllers/notes_controller.rb
```diff
@@ -99,4 +104,8 @@ def set_auto_save_key
       "node-#{@node.id}-note"
     end
   end
+
+  def set_form_preview_path
+    @form_preview_path = @note.new_record? ? {} : { preview_url: preview_project_node_note_path(current_project, @node, @note) }
```
aapomm, line 109: ditto

app/views/evidence/_form.html.erb
```diff
@@ -20,7 +20,7 @@
         data: {
           allow_dropdown: @evidence.new_record?,
           behavior: 'auto-save rich-toolbar drop-zone',
-          paths: editor_paths,
+          paths: editor_paths.merge!(@form_preview_path),
```
aapomm, line 23: Can we directly use `@form_preview_path` in the MarkupHelper instead of messing with the view? That way, we have all the paths logic in one place: `editor_paths`

spec/features/evidence_spec.rb
```diff
@@ -80,8 +80,10 @@
     let(:model) { @evidence }
     include_examples 'nodes pages breadcrumbs', :show, Evidence
 
-    let(:record) { create(:evidence, issue: @issue, node: @node, content: "#[Title]#\nTitle\n\n#[Description]#\nLiquid: {{evidence.fields['Title']}}") }
-    include_examples 'liquid dynamic content', :evidence, true
+    describe 'when including liquid content' do
+      let(:record) { create(:evidence, issue: @issue, node: @node, content: "#[Title]#\nFoo\n\n#[Description]#\nLiquid: {{evidence.title}}") }
```
aapomm, line 84: Will we be winning if we add a FactoryBot trait for this type of content? Something like:
```
    trait :with_liquid do
      content '.....'
    end
```

spec/support/liquid_shared_examples.rb
```diff
@@ -1,16 +1,31 @@
...
-      expect(find('.note-text-inner')).not_to have_content("Liquid: {{#{item_type}.fields['Title']}}")
-    end
-  else
-    it 'dynamically renders item properties' do
-      visit polymorphic_path([current_project, record])
-      expect(find('.note-text-inner')).to have_content("Liquid: #{record.fields["Title"]}")
-      expect(find('.note-text-inner')).not_to have_content("Liquid: {{#{item_type}.fields['Title']}}")
-    end
+  before do
+    @path = node_association ?
```
aapomm, line 4: https://github.com/rubocop/ruby-style-guide#multi-line-ternary-operator

spec/support/liquid_shared_examples.rb
```diff
@@ -1,16 +1,31 @@
...
+  it 'dynamically renders item properties' do
+    expect(find('.note-text-inner')).to have_content("Liquid: #{record.title}")
+    expect(find('.note-text-inner')).not_to have_content("Liquid: {{#{item_type}.title}}")
+  end
+end
+
+shared_examples 'liquid preview' do |item_type, node_association|
+
+  before do
+    @path = node_association ?
```
aapomm, line 18: https://github.com/rubocop/ruby-style-guide#multi-line-ternary-operator

app/controllers/notes_controller.rb
caitmich: Because we're setting `@note` inline instead of in the before action, but I don't see any reason for that so I've updated the new action to use the callbacks instead

app/controllers/concerns/previewable.rb
caitmich: The path will fall back to the `markup#preview` action if `@form_preview_path` is not set in the controller

Adding a `before_action :set_form_preview_path` here will not work since we need `@issue/@evidence` etc. to be set before we set the path here. Is this what you meant?

app/controllers/evidence_controller.rb
```diff
@@ -108,12 +110,18 @@ def autogenerate_issue
   end
 
   def liquid_resource_assigns
+    # find evidence inline for preview action
+    @evidence = @evidence || @node.evidence.find_by(id: params[:id])
```
etdsoft, line 114: this is a smell and not #liquid_resource_assigns' task.

app/controllers/evidence_controller.rb
```diff
@@ -6,11 +6,13 @@ class EvidenceController < NestedNodeResourceController
   include MultipleDestroy
   include NodesSidebar
   include NotificationsReader
+  include Previewable
 
-  before_action :set_or_initialize_evidence, except: [ :index, :create_multiple ]
+  before_action :set_or_initialize_evidence, only: [ :show, :new, :create, :edit, :update, :destroy ]
```
etdsoft, line 11: Is this what's causing the later smell (L114)? And if so, why is this the right pattern?

app/controllers/evidence_controller.rb
```diff
@@ -108,12 +110,18 @@ def autogenerate_issue
...
+    @evidence = @evidence || @node.evidence.find_by(id: params[:id])
     {
       'evidence' => EvidenceDrop.new(@evidence),
-      'node' => NodeDrop.new(@evidence.node)
+      'node' => NodeDrop.new(@node)
     }
   end
 
+  def set_form_preview_path
+    @form_preview_path = preview_project_node_evidence_path(current_project, @node, @evidence)
```
etdsoft, line 122: I'm sure you can metaprogram this one, but hey 11th hour and all that.

app/controllers/evidence_controller.rb
caitmich: Removed

app/controllers/evidence_controller.rb
caitmich: This change was made in [peer review](https://github.com/dradis/dradis-ce/pull/1204#discussion_r1403067494) to use a safer whitelist approach

app/controllers/evidence_controller.rb
caitmich: Updated to use a single method in `Previewable`

app/controllers/notes_controller.rb
```diff
@@ -8,13 +8,11 @@ class NotesController < NestedNodeResourceController
   include NodesSidebar
   include NotificationsReader
 
-  before_action :find_or_initialize_note, except: [:index, :new, :multiple_destroy]
+  before_action :find_or_initialize_note, except: [:index, :multiple_destroy]
   before_action :initialize_nodes_sidebar, only: [:edit, :new, :show]
   before_action :set_auto_save_key, only: [:new, :create, :edit, :update]
 
   def new
-    @note = @node.notes.new
```
caitmich, line 16: Unrelated but we can set @note using the callback

PR #1205, opened 2023-11-21
Add LiquidAssignsService

aapomm:
### Summary

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/services/liquid_assigns_service.rb
+++ b/app/services/liquid_assigns_service.rb
@@ -1,0 +1,18 @@
+class LiquidAssignsService
+  attr_accessor :project
+
+  def initialize(project)
+    @project = project
+  end
+
+  def assigns
+    result = {}
+    result['project'] = ProjectDrop.new(project)
+    result['issues'] = project.issues.map { |issue| IssueDrop.new(issue) }
+    result['nodes'] = project.nodes.user_nodes.map { |node| NodeDrop.new(node) }
+    result['evidences'] = project.evidence.map { |evidence| EvidenceDrop.new(evidence) }
+    result['tags'] = project.tags.map { |tag| TagDrop.new(tag) }
+
+    result
+  end
+end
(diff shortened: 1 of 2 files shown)


app/services/liquid_assigns_service.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def assigns
+    result = {}
+    result['project'] = ProjectDrop.new(project)
+    result['issues'] = project.issues.map { |issue| IssueDrop.new(issue) }
+    result['nodes'] = project.nodes.user_nodes.map { |node| NodeDrop.new(node) }
+    result['evidences'] = project.evidence.map { |evidence| EvidenceDrop.new(evidence) }
+    result['tags'] = project.tags.map { |tag| TagDrop.new(tag) }
+
+    if defined?(Dradis::Pro)
```
caitmich, line 16: What about passing in kwargs instead? 
Ex.
`@team = args.fetch(:team, nil)` in the initializer, and
`result['team'] = TeamDrop.new(project.team) if team` 
on L19

aapomm: I like that better than the `Dradis::Pro` check but listing all the args sort of defeats the convenience this brings for the caller. If a caller has to instantiate the class with the following:
```
LiquidAssignsService.new(
  project: project,
  team: project.team,
  document_properties: project.content_library.properties
)
```

Then it starts to look like:
```
result['project'] = ...
result['document_properties'] = ...
result['team'] = ...
```

just without the drop?

spec/support/revision_shared_examples.rb
```diff
@@ -61,6 +61,11 @@
 
       expect do
         visit project_trash_path(current_project)
+
+        within '#trash' do
```
aapomm, line 65: A bit unrelated to the task but it makes this specific spec not dependent on the other specs that create records by counting the records in the trash instead of expecting it to be empty.

app/services/liquid_assigns_service.rb
etdsoft: I'd extract this to a `#assigns_pro` that only gets called if `defined?` as we do in other places.

PR #1207, opened 2023-11-29
Activity filters

skliask:
#### Problem
Dradis has an activity feed on the Dashboard, with a link to view all activities in the project. With the current implementation, users don’t have a way to narrow down the activities that are listed in the all activities view.

#### Solution

We can implement a set of filters that will allow users to narrow down the activities by date range, type (Issue, Evidence, etc), and user.

The result should allow users to easily determine which activities:
were performed by a specific user
were performed during a certain date range
were performed on a certain date
are related to Issues, Evidence, Notes, Nodes, etc

### Check List

- [x] Added a CHANGELOG entry


CHANGELOG
```diff
@@ -3,6 +3,7 @@
   - Replace unicorn with puma in production
   - Add importmap-rails to handle js libraries
   - Revision history: Improve version history for content with carriage return
+  - Activity filtering
```
MattBudz, line 6: This doesn't meet the defined changelog format.

```[entity]: [future tense verb] [feature]```

To be fair a few other entires here don't either. We'll have to update those before the release, but let's get this one sorted out here.

File: app/views/activities/_filters.html.erb [hunk 2]
```diff
@@ -0,0 +1,14 @@
+<div class="activity-filters ">
+  <%= form_with url: project_activities_path(current_project.id), method: :get, remote: true, class: 'form-row d-flex flex-row' do |f| %>
+    <%= f.select :user, options_from_collection_for_select(User.all, :id, :name, params[:user]), { include_blank: 'Filter by user'}, { class: 'form-select m-2 form-select-sm', onchange: '$(this.form).submit();' } %>
```
MattBudz, line 3: - This `<select>` is missing a `<label>`
- Let's change the include blank to `All users` since the label will say something like `User`
- Let's avoid inline `onchange` attributes. I believe this is inconsistent with the rest of the app. Setting up a listener in a JS file for all elements that need it will be cleaner since they all call the same method on change.

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,14 @@
...
+    <%= f.select :type, Activity.select(:trackbable_type).distinct.pluck(:trackable_type), { include_blank: 'Filter by type', selected: params[:type]}, { class: 'form-select m-2 form-select-sm', onchange: '$(this.form).submit();' } %>
```
MattBudz, line 4: ditto

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,14 @@
...
+
+    <div id="" class="input-group">
```
MattBudz, line 6: why do we need this blank ID here?

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,14 @@
...
+      <%= f.date_field :period_start, max: Date.today, placeholder: "From date", value: params[:period_start], class: 'form-control form-control-sm my-2', onchange: '$(this.form).submit();' %>
```
MattBudz, line 7: Let's add a span here that looks like a label, I think `Date range` will work well. Then we need some actual label elements for both date inputs that have a `visually-hidden` class for accessibility.

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,14 @@
...
+      <span class="input-group-text input-group-sm my-2">to</span>
+      <%= f.date_field :period_end, max: Date.today, placeholder: "To date", value: params[:period_end] || Date.today, class: 'form-control form-control-sm my-2', onchange: '$(this.form).submit();' %>
+    </div>
+
+    <%= link_to 'Clear',project_activities_path(current_project.id) , class: "btn btn-danger m-2 pt-2" %>
```
MattBudz, line 12:
```suggestion
    <%= link_to 'Clear', project_activities_path(current_project.id), class: 'btn btn-danger m-2 pt-2' %>
```

Let's add an icon to the button for consistency 
```html
<i class="fa-solid fa-xmark"></i>
```

app/views/activities/_filters.html.erb
MattBudz, line 3:
```suggestion
    <%= f.select :user, options_from_collection_for_select(User.all, :id, :name, params[:user]), { include_blank: 'Filter by user' }, { class: 'form-select m-2 form-select-sm', onchange: '$(this.form).submit();' } %>
```

app/views/activities/_filters.html.erb
MattBudz, line 4:
```suggestion
    <%= f.select :type, Activity.select(:trackbable_type).distinct.pluck(:trackable_type), { include_blank: 'Filter by type', selected: params[:type] }, { class: 'form-select m-2 form-select-sm', onchange: '$(this.form).submit();' } %>
```

app/views/activities/_filters.html.erb
MattBudz, line 7:
```suggestion
      <%= f.date_field :period_start, max: Date.today, placeholder: 'From date', value: params[:period_start], class: 'form-control form-control-sm my-2', onchange: '$(this.form).submit();' %>
```

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="activity-filters ">
+  <%= form_with url: project_activities_path(current_project.id), method: :get, remote: true, class: 'form-row d-flex flex-row' do |f| %>
+    <%= f.select :user, options_from_collection_for_select(User.all, :id, :name, params[:user]), { include_blank: 'Filter by user'}, { class: 'form-select m-2 form-select-sm', onchange: '$(this.form).submit();' } %>
+    <%= f.select :type, Activity.select(:trackbable_type).distinct.pluck(:trackable_type), { include_blank: 'Filter by type', selected: params[:type]}, { class: 'form-select m-2 form-select-sm', onchange: '$(this.form).submit();' } %>
+
+    <div id="" class="input-group">
+      <%= f.date_field :period_start, max: Date.today, placeholder: "From date", value: params[:period_start], class: 'form-control form-control-sm my-2', onchange: '$(this.form).submit();' %>
+      <span class="input-group-text input-group-sm my-2">to</span>
+      <%= f.date_field :period_end, max: Date.today, placeholder: "To date", value: params[:period_end] || Date.today, class: 'form-control form-control-sm my-2', onchange: '$(this.form).submit();' %>
```
MattBudz, line 9:
```suggestion
      <%= f.date_field :period_end, max: Date.today, placeholder: 'To date', value: params[:period_end] || Date.today, class: 'form-control form-control-sm my-2', onchange: '$(this.form).submit();' %>
```

spec/features/activity_pages/activity_pages_spec.rb
```diff
@@ -12,16 +12,36 @@
...
+        10.times do
+          activity = Activity.create(
+            user: second_user,
+            trackable_type: trackable_card.class,
+            trackable_id: trackable_card.id,
+            action: 'update',
+            created_at: Time.current
+          )
+        end
+
```
MattBudz, line 37:
```suggestion
```
or add a line between L27 and L28, let's be consistent.

MattBudz (changes requested):

app/views/activities/_filters.html.erb
skliask: we dont. it's a leftover from the previous implementation with the external library datepicker. i removed it. thanks for noticing.

app/views/activities/_filters.html.erb
skliask: 1. I am a little skeptical about the label here as it will affect the display. That's why I added the `Filter by user` in the `include blank` as a CTA. I tried to be careful with spacing. How do you feel about adding hidden labels?
2. I agree,`All users` is much better.
3. Agree

spec/features/activity_pages/activity_pages_spec.rb
skliask: added a line at L29 to make it consistent. does it look better now?

app/views/activities/_filters.html.erb
skliask: Something like this?
Screenshot 2023-12-06 at 14 05 18

app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -29,4 +29,10 @@ document.addEventListener('turbolinks:load', function() {
       }
     });
   }
+
+  $(function() {
+    $(".activity-filters").change(function() {
+      $(this).children('form').submit();
+    });
+  });
```
MattBudz, line 37:
```suggestion
  $(function() {
    $('[data-behavior~=activity-filters]').change(function() {
      $(this).submit();
    });
  });
```
We don't use classes, IDs, element types, etc for selectors in scripts. We always use data-attributes so it's clear in the view file that this is for JS. A div with a class `activity-filters` will likely make others assume this is for CSS styling rather than a listener in a js file.

app/controllers/activities_controller.rb
```diff
@@ -7,10 +7,11 @@ class ActivitiesController < AuthenticatedController
...
     @page = params[:page].present? ? params[:page].to_i : 1
 
-    activities = current_project.activities
-                                .includes(:trackable)
-                                .order(created_at: :desc)
-                                .page(@page)
+    activities = current_project.activities.includes(:trackable)
+    activities = activities.filter_by_user_id(params[:user]) if params[:user].present?
+    activities = activities.filter_by_type(params[:type]) if params[:type].present?
+    activities = activities.filter_by_date(period_start, period_end) if params[:period_start].present?
```
caitmich, line 13: What about moving these to a helper method so that the index action is less cluttered? 
Ex. `activities = filter_activities`

app/models/activity.rb
```diff
@@ -26,6 +26,11 @@ def project=(new_project); end
   scope :latest, -> do
     includes(:trackable).order('activities.created_at DESC').limit(20)
   end
+  scope :filter_by_user_id, -> (id) { where user_id: id }
```
caitmich, line 29:
```suggestion
  scope :by_user, -> (id) { where user_id: id }
```
Same on `L30-31` to keep consistent with convention/our existing scope names

app/models/activity.rb
```diff
@@ -26,6 +26,11 @@ def project=(new_project); end
...
+  scope :filter_by_type, -> (type) { where trackable_type: type }
+  scope :filter_by_date, -> (period_start, period_end) {
+    where('created_at >= ? AND created_at <= ?', period_start.beginning_of_day, period_end.end_of_day)
```
caitmich, line 32: What about specifying `beginning_of_day` and `end_of_day` in the controller instead?

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,29 @@
+<%= form_with url: project_activities_path(current_project.id), method: :get, remote: true, class: 'd-flex flex-xl-row flex-column gap-3', data: { behavior: 'activity-filters' } do |f| %>
```
caitmich, line 1:
```suggestion
<%= form_with url: project_activities_path(current_project), method: :get, remote: true, class: 'd-flex flex-xl-row flex-column gap-3', data: { behavior: 'activity-filters' } do |f| %>
```

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  <div class="d-flex flex-column flex-sm-row w-100 gap-3">
+    <div class="w-100">
+      <%= f.label :user, class: 'form-label' %>
+      <%= f.select :user, options_from_collection_for_select(User.all, :id, :name, params[:user]), { include_blank: 'All users' }, { class: 'form-select' } %>
+    </div>
+    <div class="w-100">
+      <%= f.label :type, class: 'form-label' %>
+      <%= f.select :type, options_from_collection_for_select(Activity.select(:trackbable_type).distinct.pluck(:trackable_type), :to_s, :demodulize ), { include_blank: 'All types', selected: params[:type] }, { class: 'form-select' } %>
```
caitmich, line 9:
```suggestion
      <%= f.select :type, options_for_select(Activity.pluck(:trackable_type).uniq), { include_blank: 'All types', selected: params[:type] }, { class: 'form-select' } %>
```

spec/features/activity_pages/activity_pages_spec.rb
```diff
@@ -75,6 +99,49 @@
           end
         end
       end
+
+      describe 'filters' do
+        it 'has user filter' do
+          expect(page).to have_selector('#user', count: 1)
+        end
+
+        it 'user filter works' do
```
caitmich, line 108: A more descriptive name would be better, take a look at some of the other ones in this file

spec/models/activity_spec.rb
```diff
@@ -11,18 +11,86 @@
...
         issue    = create(:issue)
         activity = create(:activity, trackable: issue)
-        expect(activity.trackable_type).to eq "Issue"
+        expect(activity.trackable_type).to eq 'Issue'
         expect(activity.reload.trackable).to eq issue
       end
     end
   end
 
+  describe '#filter_by_user_id' do
```
caitmich, line 28: These could be moved to a single block and share a before action

app/assets/javascripts/tylium/pages/activities.js
skliask: Also changed this:
`$(this).children('form').submit();`
to this so it works:
`$(this).submit();`

app/views/activities/_filters.html.erb
skliask: That is how i had done it initially. However, there are cases that the types are namespaced within modules and they look bad on the dropdown. That way we benefit from the demodulize method for display reasons.

spec/features/activity_pages/activity_pages_spec.rb
skliask: Changed the names to make more sense as a sentence based on their describe block.

app/views/activities/_filters.html.erb
caitmich: That's fine, `options_for_collection_from_select` works too

The main changes here are the `:trackable_type` typo fix and the query adjustment. `Activity.pluck(:trackable_type).uniq` 

Screen Shot 2023-12-12 at 3 23 07 PM

spec/models/activity_spec.rb
```diff
@@ -11,18 +11,81 @@
...
+      activity = create(:activity, trackable: issue, user: @user, created_at: DateTime.now.beginning_of_year)
+      activity2 = create(:activity, trackable: card, user: @user, created_at: DateTime.now.beginning_of_year)
+      activity3 = create(:activity, trackable: card, created_at: DateTime.now.beginning_of_year - 1.year)
+    end
+
+    describe '#by_user' do
+      context 'when passed a valid user id' do
+        it 'returns activities of this user' do
+          expect(Activity.by_user(@user.id).count).to eq 2
+          expect(Activity.all.count).to eq 3
```
caitmich, line 42: For the sake of keeping things simple I don't think we need this check in every test, the check for the scoped records should be enough

app/assets/javascripts/tylium/pages/activities.js
MattBudz: > Also changed this:
$(this).children('form').submit();
to this so it works:
$(this).submit();

Perfect, that was part of my suggested change above. 👍

CHANGELOG
```diff
@@ -4,6 +4,7 @@
   - Revision history: Improve version history for content with carriage return
   - Tylium: Show liquid content in editor preview
   - Web-server: Replace unicorn with puma in production
+  - Activity: Filter activities by user, type and date
```
MattBudz, line 7:
```suggestion
  - Activity: Filter activities by user, type, and date
```

app/controllers/activities_controller.rb
```diff
@@ -30,4 +29,38 @@ def poll
...
+      DateTime.parse params[:period_end]
+    end
+  end
+
+  def filtering_params
+    params.slice(:user, :type, :period_start, :period_end)
+  end
+
+  def filter_activities(activities)
+    filtering_params.each do |k, value|
```
MattBudz, line 52: why `k` and `value` rather than `k` and `v` or `key` and `value`?

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <%= f.label :type, class: 'form-label' %>
+      <%= f.select :type, options_from_collection_for_select(current_project.activities.pluck(:trackable_type).uniq, :to_s, :demodulize, params[:type] ), { include_blank: 'All types', selected: params[:type] }, { class: 'form-select' } %>
+    </div>
+  </div>
+
+  <div class="w-100">
+    <span class="form-label mb-2 d-block">Date range</span>
+    <div class="d-flex flex-column flex-sm-row gap-3 w-100">
+      <div class="input-group">
+        <%= f.date_field :period_start, max: params[:period_end], placeholder: 'From date', value: params[:period_start], class: 'form-control form-control-sm', onkeydown: "return false" %>
```
MattBudz, line 17: let's move `onkeydown` to the related js file and [avoid inline js](https://github.com/dradis/dradis-ce/wiki/Style-Guide:-JavaScript#code)

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        <%= f.label :period_start, class: 'visually-hidden' %>
+        <span class="input-group-text input-group-sm">to</span>
+        <%= f.label :period_end, class: 'visually-hidden' %>
+        <%= f.date_field :period_end, max: Date.today, min: params[:period_start], placeholder: 'To date', value: params[:period_end] || Date.today, class: 'form-control form-control-sm', onkeydown: "return false" %>
```
MattBudz, line 21: ditto

app/assets/stylesheets/tylium/modules/activities.scss
```diff
@@ -0,0 +1,16 @@
...
+  input[type="date"]::-webkit-calendar-picker-indicator {
+    position: absolute;
+    top: 0;
+    right: 0;
+    width: 100%;
+    height: 100%;
+    padding: 0;
+    color: transparent;
+    background: transparent;
+  }
```
MattBudz, line 15: Isn't this a bit hacky? It doesn't work in Firefox and it's not consistent with other areas of the app where we have a date picker (cards#new for example). Why don't we allow the default behavior and validate the input before submitting?

MattBudz: Let's remove it and not cause inconsistencies with other views and cross-browser compatibility issues.

app/controllers/activities_controller.rb
skliask: i changed it to `key` and `value`

CHANGELOG
skliask: Should we add a comma when we have the and right after it?

MattBudz: I generally add it in a list context. Without the comma, this can be interpreted as "you can filter by user and/or type + date", suggesting that type and date are always done together. With the comma, it avoids ambiguity by expressing that filtering can be done by any combination of the 3 options. This is just a #suggestion and my personal style of writing. I don't feel strongly about it but Grammarly was correcting it so I added it as a comment.

skliask: I see how it could create confusion. Changed it!

app/views/activities/index.html.erb
```diff
@@ -11,6 +11,10 @@
   </nav>
 <% end %>
 
+<div class="content-container mt-3 py-3 activity-filters">
```
MattBudz, line 14:
```suggestion
<div class="content-container mt-3 py-3">
```

do we need this class? I don't see it being used.

app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -29,4 +29,17 @@ document.addEventListener('turbolinks:load', function() {
       }
     });
   }
+
+  $(function() {
+
```
MattBudz, line 34:
```suggestion
```

app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -29,4 +29,17 @@ document.addEventListener('turbolinks:load', function() {
...
+    $('[data-behavior~=activity-filters]').change(function() {
+      $(this).submit();
+    });
+
+    $('[data-behavior~=activity-filters]').keydown(function() {
+      return false
+    });
```
MattBudz, line 41: This doesn't work in FF

app/views/activities/index.html.erb
skliask: I used it while testing and forgot it there.I removed it now.

app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -29,4 +29,20 @@ document.addEventListener('turbolinks:load', function() {
...
   }
+
+  $(function() {
+    $('[data-behavior~=activity-filters] select').change(function () {
+      $(this).closest('form').submit();
+    });
+
+    $('[data-behavior~=activity-filters] input[type="date"]').change(function () {
+      const dateValue = $(this).val();
+      const periodStart = $("#period_start").attr("min");
```
MattBudz, line 40: We use data-attributes for [selectors in JS](https://github.com/dradis/dradis-ce/wiki/Style-Guide:-JavaScript#code) to keep things clean and clear. Looking at the view file, you may not think that changing an `id` attribute will break js functionality. Modifying a `data-behavior` attribute will likely make you think about JS.

app/controllers/activities_controller.rb
MattBudz: >i changed it to key and value

I don't see this change.

app/helpers/activities_helper.rb
```diff
@@ -0,0 +1,5 @@
+module ActivitiesHelper
+  def users_for_select
+    current_project.activities.map(&:user).uniq.union(current_project.authors.enabled)
```
MattBudz, line 3: do we need a helper just for this?

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,29 @@
...
+
+  <div class="w-100">
+    <span class="form-label mb-2 d-block">Date range</span>
+    <div class="d-flex flex-column flex-sm-row gap-3 w-100">
+      <div class="input-group">
+        <%= f.date_field :period_start, max: params[:period_end] || Date.today, min: "01-01-2016", placeholder: 'From date', value: params[:period_start], class: 'form-control form-control-sm' %>
+        <%= f.label :period_start, class: 'visually-hidden' %>
+        <span class="input-group-text input-group-sm">to</span>
+        <%= f.label :period_end, class: 'visually-hidden' %>
+        <%= f.date_field :period_end, max: Date.today, min: params[:period_start], placeholder: 'To date', value: params[:period_end] || Date.today, class: 'form-control form-control-sm' %>
```
MattBudz, line 21:
```suggestion
        <%= f.date_field :period_start, max: params[:period_end] || Date.today, min: "01-01-2016", placeholder: 'From date', value: params[:period_start], class: 'form-control' %>
        <%= f.label :period_start, class: 'visually-hidden' %>
        <span class="input-group-text">to</span>
        <%= f.label :period_end, class: 'visually-hidden' %>
        <%= f.date_field :period_end, max: Date.today, min: params[:period_start], placeholder: 'To date', value: params[:period_end] || Date.today, class: 'form-control' %>
```

This is just for consistency (I should've noticed this earlier)

app/helpers/activities_helper.rb
skliask: I think it's much more tidy to add it there. Wouldn't it be a lot of logic for a view?

app/controllers/activities_controller.rb
skliask: Must have slipped with all of the changes. Now it's there.

app/helpers/activities_helper.rb
MattBudz: >Wouldn't it be a lot of logic for a view?

yes for a view it would be too much, but I was thinking maybe just have something like:
```ruby
@users_for_select = current_project.activities.map(&:user).uniq.union(current_project.authors.enabled)
```
In the controller's `index` action. But I'll leave this up to you and @aapomm

app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -29,4 +29,20 @@ document.addEventListener('turbolinks:load', function() {
       }
     });
   }
+
+  $(function() {
```
aapomm, line 33: What's the reasoning behind this wrapping of the below functions?

app/controllers/activities_controller.rb
```diff
@@ -30,4 +30,38 @@ def poll
...
+
+  def filter_activities(activities)
+    filtering_params.each do |key, value|
+      next if value.blank?
+
+      if key == 'period_start'
+        value = DateTime.parse(value).beginning_of_day
+      elsif key == 'period_end'
+        value = DateTime.parse(value).end_of_day
+      end
```
aapomm, line 60: I think we need to add an `else` here otherwise, if an evil user can pass any param here and it will still be called by `activities.send("by_#{key}", value)`

app/helpers/activities_helper.rb
aapomm: This is mostly done in the controller unless we use it for other controllers or actions.
Also, what's the reasoning behind showing all the `current_project.authors` ? I imagine a user would only want to filter users that have activities, so ` current_project.activities.map(&:user)` should be enough?

app/controllers/activities_controller.rb
```diff
@@ -30,4 +30,38 @@ def poll
       format.js
     end
   end
+
+  private
+
+  def period_start
```
aapomm, line 36: Thoughts on adding an ActivitiesFilter service class? This would save the controller from having any specific DateTime logic

app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -29,4 +29,20 @@ document.addEventListener('turbolinks:load', function() {
...
+    $('[data-behavior~=activity-filters] select').change(function () {
+      $(this).closest('form').submit();
+    });
+
+    $('[data-behavior~=activity-filters] input[type="date"]').change(function () {
+      const dateValue = $(this).val();
+      const periodStart = $('[data-behavior~=period-start]').attr('min');
```
aapomm, line 40: The variable name is a bit misleading since this isn't the value of the periodStart but the earliest date that's possible, which I believe is hardcoded to be 01-01-2016.

app/assets/javascripts/tylium/pages/activities.js
```diff
@@ -29,4 +29,20 @@ document.addEventListener('turbolinks:load', function() {
...
+      if (dateValue !== '' && !isNaN(Date.parse(dateValue))) {
+        if (Date.parse(dateValue) >= Date.parse(periodStart)) {
+          $(this).closest('form').submit();
+        }
```
aapomm, line 44: Shouldn't we also check if periodEnd >= periodStart?

app/models/activity.rb
```diff
@@ -26,6 +26,10 @@ def project=(new_project); end
   scope :latest, -> do
     includes(:trackable).order('activities.created_at DESC').limit(20)
   end
+  scope :by_user, -> (id) { where user_id: id }
+  scope :by_type, -> (type) { where trackable_type: type }
+  scope :by_period_start, -> (period_start) { where('created_at >= ?', period_start) }
+  scope :by_period_end, -> (period_end) { where('created_at <= ?', period_end) }
```
aapomm, line 32: Unlike the 2 lines above, I can get behind this syntactic sugar. However, `Activity.by_period_start` makes me think that we're searching for activities in that exact date, so maybe we can improve the semantics? Check out this method that does the exact same thing: https://github.com/dradis/dradis-ce/blob/main/app/models/notification.rb#L21. `Activity.since(Date.yesterday)` reads better in my opinion. On the other hand, `Activity.before(Date.today)` sounds better than `Activity.by_period_end(Date.today)`

spec/features/activity_pages/activity_pages_spec.rb
```diff
@@ -12,16 +12,40 @@
   context 'as authenticated user' do
     describe 'index page', js: true do
       let(:trackable) { create(:issue) }
+      let(:trackable_card) { create(:card) }
       let(:user) { create(:user) }
-
+      let(:second_user) { create(:user) }
       let(:create_activities) do
-        50.times do
+        35.times do
```
aapomm, line 19: Instead of keeping the `X.times` approach, what about using factory bot and calling `create_list`?

spec/features/activity_pages/activity_pages_spec.rb
```diff
@@ -12,16 +12,40 @@
...
             trackable_type: trackable.class,
             trackable_id: trackable.id,
             action: 'update',
-            created_at: Time.current - ((1..5).to_a.sample.days)
+            project: current_project,
+            created_at: Time.current - 1.month
+          )
+        end
+
+        10.times do
```
aapomm, line 30: ditto

spec/features/activity_pages/activity_pages_spec.rb
```diff
@@ -12,16 +12,40 @@
...
+            user: second_user,
+            trackable_type: trackable_card.class,
+            trackable_id: trackable_card.id,
+            action: 'update',
+            project: current_project,
+            created_at: Time.current
+          )
+        end
+
+        5.times do
```
aapomm, line 41: ditto

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <%= f.label :type, class: 'form-label' %>
+      <%= f.select :type, options_from_collection_for_select(current_project.activities.pluck(:trackable_type).uniq, :to_s, :demodulize, params[:type] ), { include_blank: 'All types', selected: params[:type] }, { class: 'form-select' } %>
+    </div>
+  </div>
+
+  <div class="w-100">
+    <span class="form-label mb-2 d-block">Date range</span>
+    <div class="d-flex flex-column flex-sm-row gap-3 w-100">
+      <div class="input-group">
+        <%= f.date_field :period_start, max: params[:period_end] || Date.today, min: "01-01-2016", placeholder: 'From date', value: params[:period_start], class: 'form-control', data: { behavior: 'period-start' } %>
```
aapomm, line 17: What's the reasoning behind the min date: `min: "01-01-2016"` ?

app/controllers/activities_controller.rb
```diff
@@ -30,4 +30,38 @@ def poll
...
+  def period_end
+    if params[:period_end].empty?
+      DateTime.now
+    else
+      DateTime.parse params[:period_end]
+    end
+  end
+
+  def filtering_params
+    params.slice(:user, :type, :period_start, :period_end)
```
aapomm, line 49: Do we need some validation around these params? It seems like any user can pass what ever values in the URL: `activities?user=2&type=&period_start=2023-03-20&period_end=2024-02-01`. If I pass a bogus date, it throws a server error. If I pass any user_id, it seems to still use it for querying. Same with `type`.

app/helpers/activities_helper.rb
MattBudz: > Also, what's the reasoning behind showing all the current_project.authors ? I imagine a user would only want to filter users that have activities

If there are many activities by many users, an admin may not know if a particular user on the project has an activity or not (at least not at a glance). If the list of filterable users doesn't include a user who's on the project, it can lead to confusion: "Why can't I filter Aaron's activities if Aaron is on the project?" With this implementation, the admin will get a clear answer showing them that there are no activities that match the query.

app/views/activities/_filters.html.erb
MattBudz: From BC:

> We should also set a min for the period_start input, currently, we allow dates from year 0001. I'd say a reasonable limit would be Jan 1 2016 (Acitivites were [introduced on Feb 29 2016](https://github.com/dradis/dradis-ce/commit/3368f436b43402b80aad2b16f9ffd82a9df7fb93#diff-6c3192e53b5a2e74e2fe29782dce058fe0398676d09068812b92c40d88625755) so no results would exist before this date).

Rather than Feb 29 I suggested Jan 1 so it's a nice round date, but technically it can be any date before Feb 29 2016.

app/assets/javascripts/tylium/pages/activities.js
skliask: The default date is just the fallback date that is hardcoded in the view (01-01-2016) . This variable still refers though to the actual periodStart. When a user changes the starting date the hardcoded date gets overwritten.

app/assets/javascripts/tylium/pages/activities.js
skliask: The reason we have this check is to ensure that the form won't be submitted with an incomplete starting date. We wanted to find a way to make sure that a user can input a date using the keyboard as well as the mouse. 

However, through the keyboard, the form was being submitted as long as we were hitting the year's part(dd/mm/**yyyy**) of the input. To prevent this from happening we are checking that a starting date is complete before submitting the form. This edge case scenario only happens with periodStart since the periodEnd is always filled.

app/models/activity.rb
```diff
@@ -26,6 +26,10 @@ def project=(new_project); end
   scope :latest, -> do
     includes(:trackable).order('activities.created_at DESC').limit(20)
   end
+  scope :by_user, -> (id) { where user_id: id }
+  scope :by_type, -> (type) { where trackable_type: type }
```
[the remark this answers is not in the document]
skliask: Fair point!

app/models/activity.rb
skliask: I like this. changing it!

app/assets/javascripts/tylium/pages/activities.js
skliask: Good catch. It left over when trying out things. Removed it

app/controllers/activities_controller.rb
skliask: Since we will be using Activity.since(date) & Activity.before(date) there is no risk here of sending dynamic methods anymore.

app/controllers/activities_controller.rb
skliask: Updated the controller and we don't need anymore #period_start and #period_end. I believe that since we only use #filter_activities() for filtering we can still leave that there to reduce overhead. If you still think it will be neater with and ActivitiesFilter class we can still go for it.

app/helpers/activities_helper.rb
skliask: Removed the `ActivitiesHelper` and moved `@users_for_selec`t in the controller.

spec/features/activity_pages/activity_pages_spec.rb
skliask: You're right. It was already there so I didn't really notice. Changed it!

app/controllers/activities_controller.rb
skliask: Since we are not submitting any data and we are only querying activities that are already scoped for the current project is it a big issue? The server error with bogus dates is something that could bother me but still, we won't submit it still so maybe I could do something to just prevent it from throwing a server error

app/controllers/activities_controller.rb
```diff
@@ -30,4 +30,28 @@ def poll
...
+      next if value.blank?
+
+      if key == 'since'
+        value = DateTime.parse(value).beginning_of_day rescue ArgumentError
+        activities = activities.since value
+      elsif key == 'before'
+        value = DateTime.parse(value).end_of_day rescue ArgumentError
+        activities = activities.before value
+      elsif ['user_id', 'trackable_type'].include? key
+        activities = activities.where(key.to_sym => value)
```
aapomm, line 51: I don't think we need to symbolize

```suggestion
        activities = activities.where(key => value)
```

app/controllers/activities_controller.rb
```diff
@@ -6,11 +6,11 @@ class ActivitiesController < AuthenticatedController
 
   def index
     @page = params[:page].present? ? params[:page].to_i : 1
+    @users_for_select = current_project.activities.map(&:user).uniq.union(current_project.authors.enabled)
```
aapomm, line 9: Do we need `uniq`?

```suggestion
    @users_for_select = current_project.activities.map(&:user).union(current_project.authors.enabled)
```

app/controllers/activities_controller.rb
```diff
@@ -30,4 +30,28 @@ def poll
...
+  def filtering_params
+    params.permit(:user_id, :trackable_type, :since, :before)
+  end
+
+  def filter_activities(activities)
+    filtering_params.each do |key, value|
+      next if value.blank?
+
+      if key == 'since'
+        value = DateTime.parse(value).beginning_of_day rescue ArgumentError
```
aapomm, line 45: Using inline `rescue` looks clean but I'm not sure if it's ideal. Can we check the valid date with `Date.valid_date?` instead?

spec/features/activity_pages/activity_pages_spec.rb
```diff
@@ -75,6 +95,49 @@
           end
         end
       end
+
+      describe 'filters' do
```
aapomm, line 99: Can we also add a few cases to catch passing bad params (e.g. invalid date, invalid user, invalid type)?

CHANGELOG
```diff
@@ -1,6 +1,5 @@
 [v#.#.#] ([month] [YYYY])
-  - [entity]:
-    - [future tense verb] [feature]
+  - Activity: Filter activities by user, type, and date
```
caitmich, line 2:
```suggestion
  - Activities: Filter activities by user, type, and date
```

app/controllers/activities_controller.rb
```diff
@@ -6,11 +6,11 @@ class ActivitiesController < AuthenticatedController
 
   def index
     @page = params[:page].present? ? params[:page].to_i : 1
+    @users_for_select = current_project.activities.map(&:user).union(current_project.authors.enabled)
```
caitmich, line 9: Do we want to include users that have no activities?

app/controllers/activities_controller.rb
```diff
@@ -30,4 +30,36 @@ def poll
...
+
+    if params[:trackable_type].present?
+      activities = activities.where(trackable_type: params[:trackable_type])
+    end
+
+    activities
+  end
+
+  def valid_date?(date_string)
+    Date.parse(date_string)
```
caitmich, line 61: This will allow dates in the future. Do we want to prevent this?

app/controllers/activities_controller.rb
```diff
@@ -30,4 +30,36 @@ def poll
...
+
+  private
+
+  def filtering_params
+    params.permit(:user_id, :trackable_type, :since, :before)
+  end
+
+  def filter_activities(activities)
+    if params[:since].present? && valid_date?(params[:since])
+      activities = activities.since(DateTime.parse(params[:since]).beginning_of_day)
```
caitmich, line 42: We don't need `beginning_of_day` since DateTime.parse on a date like `2024-08-01` will return beginning of day: `Thu, 01 Aug 2024 00:00:00 +0000`

app/controllers/activities_controller.rb
caitmich: With the first part of this query we're also including users who have activities in the project even if they're no longer on the project. Do we  want to include these?

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,29 @@
+<%= form_with url: project_activities_path(current_project), method: :get, remote: true, class: 'd-flex flex-xl-row flex-column gap-3', data: { behavior: 'activity-filters' } do |f| %>
+  <div class="d-flex flex-column flex-sm-row w-100 gap-3">
+    <div class="w-100">
+      <%= f.label :user_id, class: 'form-label' %>
+      <%= f.select :user_id, options_from_collection_for_select(@users_for_select, :id, :name, params[:user_id]), { include_blank: 'All users' }, { class: 'form-select' } %>
+    </div>
+    <div class="w-100">
+      <%= f.label :trackable_type, class: 'form-label' %>
```
caitmich, line 8: Can we label this something more user-friendly?

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <%= f.select :trackable_type, options_from_collection_for_select(current_project.activities.pluck(:trackable_type).uniq, :to_s, :demodulize, params[:trackable_type] ), { include_blank: 'All types', selected: params[:type] }, { class: 'form-select' } %>
```
caitmich, line 9: Can we get the collection of trackable types in the controller?

caitmich, line 9:
```suggestion
      <%= f.select :trackable_type, options_from_collection_for_select(current_project.activities.pluck(:trackable_type).uniq, :to_s, :demodulize, params[:trackable_type] ), { include_blank: 'All types' }, { class: 'form-select' } %>
```

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    </div>
+  </div>
+
+  <div class="w-100">
+    <span class="form-label mb-2 d-block">Date range</span>
+    <div class="d-flex flex-column flex-sm-row gap-3 w-100">
+      <div class="input-group">
+        <%= f.date_field :since, max: params[:before] || Date.today, min: "01-01-2016", placeholder: 'From date', value: params[:since], class: 'form-control', data: { behavior: 'since' } %>
```
caitmich, line 17: Any reason for the min date choice?

app/views/activities/_filters.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        <%= f.label :since, class: 'visually-hidden' %>
+        <span class="input-group-text">to</span>
+        <%= f.label :before, class: 'visually-hidden' %>
+        <%= f.date_field :before, max: Date.today, min: params[:since], placeholder: 'To date', value: params[:before] || Date.today, class: 'form-control' %>
+      </div>
+
+      <%= link_to project_activities_path(current_project.id), class: 'text-error-hover text-nowrap d-flex align-items-center' do %>
```
caitmich, line 24:
```suggestion
      <%= link_to project_activities_path(current_project), class: 'text-error-hover text-nowrap d-flex align-items-center' do %>
```

app/controllers/activities_controller.rb
```diff
@@ -30,4 +30,36 @@ def poll
       format.js
     end
   end
+
+  private
+
+  def filtering_params
```
caitmich, line 36: we're not using this anywhere but probably a good idea

app/views/activities/_filters.html.erb
caitmich, line 9: Can we split this across multiple lines? ditto for the other fields

PR #1213, opened 2023-12-13
Use file name numbering on kit upload instead of overwriting files with the same name

caitmich:
### Summary

When uploading a kit, note/methodology/project templates that have the same file name as an existing template will overwrite the existing one. We don't want templates to be overwritten, but instead want to name them in sequence like we do when a user manually uploads a report template

### Check List

- [X] Added a CHANGELOG entry


app/jobs/kit_import_job.rb
```diff
@@ -17,6 +18,12 @@ def perform(file_or_folder, logger:, user_id: nil)
     @logger = logger
     @project = nil
     @report_templates_dir = Configuration.paths_templates_reports
+    @templates_dirs = TEMPLATE_TYPES.map do |template_type|
```
etdsoft, line 21: Why not make "reports" part of TEMPLATE_TYPES? They pretty much quack alike.

app/jobs/kit_import_job.rb
```diff
@@ -17,6 +18,12 @@ def perform(file_or_folder, logger:, user_id: nil)
...
+      [
+        template_type,
+        Pathname.new(Configuration.send("paths_templates_#{template_type}"))
+      ]
```
etdsoft, line 25: If you want to 🤓 out, you can do this in a single SQL query, by prebuilding the names from the TEMPLATE_TYPEs.

app/jobs/kit_import_job.rb
```diff
@@ -42,7 +49,7 @@ def perform(file_or_folder, logger:, user_id: nil)
   end
 
   private
-  attr_reader :current_user, :logger, :report_templates_dir, :working_dir
+  attr_reader :current_user, :logger, :report_templates_dir, :templates_dirs, :working_dir
```
etdsoft, line 52: ditto, no need of separate attrs

app/jobs/kit_import_job.rb
```diff
@@ -175,13 +180,23 @@ def import_rules
...
+      Dir["#{kit_template_dir}/*"].each do |file|
+        return unless File.file?(file)
+        file_name = NamingService.name_file(
+          original_filename: File.basename(file),
+          pathname: template_pwd
+        )
+
+        FileUtils.cp(file, "#{template_pwd}/#{file_name}")
+      end
+    end
```
etdsoft, line 199: This method is doing much more than what it says in the tin, with different code paths for different argument values, etc.

- Should this be one method or several based on type? or,
- Is there an abstraction that lets you handle all cases? or,
- should the naming piece be extraced?

caitmich: We don't need the `template_type == 'plugins'` path anymore so I removed that and also extracted the naming part 👍

app/jobs/kit_import_job.rb
```diff
@@ -16,22 +17,27 @@ def perform(file_or_folder, logger:, user_id: nil)
     @current_user = user_id ? User.find(user_id) : User.first
     @logger = logger
     @project = nil
-    @report_templates_dir = Configuration.paths_templates_reports
+    @templates_dirs = TEMPLATE_TYPES.map do |template_type|
+      [
+        template_type,
+        Pathname.new(Configuration.send("paths_templates_#{template_type}"))
```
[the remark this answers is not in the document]
caitmich: `paths_templates_#{template_type}` doesn't because we call [to_s](https://github.com/dradis/dradis-ce/blob/5ed0b25bc2149228b2ff77c515ed42a01075de4e/app/models/configuration.rb#L52)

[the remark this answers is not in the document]
caitmich: If i build the path using `paths_templates` it may not be accurate because they could be using a different path for any specific configuration

app/jobs/kit_import_job.rb
```diff
@@ -175,12 +180,22 @@ def import_rules
...
+      file_name = name_file(File.basename(file), destination)
+      FileUtils.cp(file, "#{destination}/#{file_name}")
+    end
+  end
 
-    FileUtils.cp_r(
-      "#{template_directory}/.",
-      Configuration.send("paths_templates_#{template_type}")
+  def name_file(original_filename, pathname)
+    NamingService.name_file(
```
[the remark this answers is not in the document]
caitmich: moved to a `copy_file` helper 👍

PR #1214, opened 2023-12-19
Nokogiri lock

skliask:
#### Problem
While uploading a burp file, there was a parsing issue that was not letting the upload to complete.

#### Solution
Locking Nokogiri to version 1.14.5 solves this issue.

Check List
- [x] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,10 +6,10 @@
   - Web-server: Replace unicorn with puma in production
   - Upgraded gems:
     - rails
-    - nokogiri (lock to 1.14.5)
   - Bugs fixes:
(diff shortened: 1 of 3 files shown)


CHANGELOG
```diff
@@ -9,6 +9,9 @@
   - Bugs fixes:
     - Evidence: Prevent loading old Evidence template content at the Issue level
     - Methodologies: validate presence of content
+    - Burp HTML: Upload file without issues
```
aapomm, line 12: I'd argue that this doesn't really solve the bug. What about just adding `nokogiri` to the `Upgraded gems` section?

skliask: Yeah, i think its better. Moved it

PR #1217, opened 2023-12-20
Issue liquid title

skliask:
### Summary

When an `Issue` has liquid content in the title, it is not being displayed properly in some parts of the app.

### Solution

Parse liquid content where the `Issue` titles are being displayed

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,7 @@
     - [gem]
   - Bugs fixes:
     - Tylium: Fix redirection when updating an issue or content block
+    - Issues: Render liquid content within the title throughout the app
     - Bug tracker items:
       - [item]
   - New integrations:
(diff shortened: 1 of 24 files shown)


CHANGELOG
```diff
@@ -9,6 +9,7 @@
   - Bugs fixes:
     - Evidence: Prevent loading old Evidence template content at the Issue level
     - Methodologies: validate presence of content
+    - Issues: Displays liquid content within the title throughout the app
```
MattBudz, line 12:
```suggestion
    - Issues: Render liquid content within the title throughout the app
```

also this should be in :alpha order

app/controllers/issues/evidence_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Issues::EvidenceController < AuthenticatedController
   include ActivityTracking
+  include LiquidEnabledResource
```
MattBudz, line 3: :alpha

app/controllers/projects_controller.rb
```diff
@@ -1,5 +1,6 @@
 class ProjectsController < AuthenticatedController
   include NotificationsReader
+  include LiquidEnabledResource
```
MattBudz, line 3: :alpha

app/helpers/application_helper.rb
```diff
@@ -27,6 +27,10 @@ def markup(text, options={})
     result[:output].to_s.html_safe
   end
 
+  def liquid_text(text)
```
MattBudz, line 30: is this the best method name for what we are doing here? Does it appropriately describe what this does?

skliask: Maybe `deliquify_text` would be better?

PR #1232, opened 2024-02-26
Migrate plugin templates to db-backed mappings & mapping_fields

caitmich:
### Summary

- Migrate existing `.template` file mappings to db records
- Run as a hidden, one-off process by using a migration
- For each Mapping, create MappingField records
- Append `.legacy` to `.template` files once migrated

### Check List

- [ ] Added a CHANGELOG entry


app/services/mapping_migration_service.rb
```diff
@@ -0,0 +1,90 @@
...
+        create_mapping_field(mapping, field_title)
+      end
+    end
+  end
+
+  def migrate_pro
+  end
+
+  def template_fields
+    template_content = File.open(template_file).read
```
aapomm, line 74:
```suggestion
    template_content = File.read(template_file)
```

app/services/mapping_migration_service.rb
```diff
@@ -0,0 +1,90 @@
+class MappingMigrationService
+  attr_reader :integration_name, :rtp_id, :template_file, :templates_dir, :upload_integrations
+
+  def initialize
+    @templates_dir = Configuration.paths_templates_plugins
+    @upload_integrations = Dradis::Plugins::with_feature(:upload)
```
aapomm, line 6: Do we need `upload_integrations` as an attribute considering we're only using it one spot?

app/services/mapping_migration_service.rb
```diff
@@ -0,0 +1,90 @@
...
+      template_files.each do |template_file|
+        @template_file = template_file
+
+        if defined?(Dradis::Pro)
+          migrate_pro
+        else
+          migrate
+        end
+        # delete the .template files after migrating them to the db
+        File.delete(template_file)
```
aapomm, line 22: Are we 100% sure about making this is irreversible action of completely deleting the templates? What about we simply rename templates and give them `.old` (e.g. `evidence.template.old`) in the likely scenario that something goes wrong?

app/services/mapping_migration_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def call
+    upload_integrations.each do |integration|
+      @integration_name = integration.plugin_name.to_s
+
+      template_files.each do |template_file|
+        @template_file = template_file
+
+        if defined?(Dradis::Pro)
```
aapomm, line 16: I think we can improve the parity here with Pro. What about this approach instead:
```
rtp_ids = defined?(Dradis::Pro) ? ReportTemplateProperties.pluck(:id) : [nil]
rtp_ids.each do |rtp_id|
  migrate
  ...
```

This will remove the need for a pro-specific change and migrate_pro entirely. Thoughts?

app/services/mapping_migration_service.rb
```diff
@@ -0,0 +1,90 @@
...
+    # for each file, create a mapping for the uploader&plugin_name combination
+    ActiveRecord::Base.transaction do
+      mapping = create_mapping
+
+      template_fields.each do |field_title, field_content|
+        # set source_field by taking the first match to the existing %% syntax
+        source_field = field_content.match(/%(?<field>\S*?)%/)
+        @source_field =
+          if source_field && !source_field['field'].empty?
+            source_field['field']
```
aapomm, line 58: I'm not sure if I'm missing something but I think we can use the same regex as in line 86.
```suggestion
        source_field = field_content.match(/%(\S*?)%/)
        @source_field =
          if source_field
            source_field[1]
```

db/migrate/20240226193757_migrate_templates_to_mappings.rb
```diff
@@ -0,0 +1,5 @@
+class MigrateTemplatesToMappings < ActiveRecord::Migration[7.0]
+  def change
```
aapomm, line 2: Let's use `up` here and leave `down` empty (unless we can recreate the `.template` files). If a dev somehow decides to rollback, the migration service will get called again.

```suggestion
  def up
```

app/services/mapping_migration_service.rb
caitmich: You're right, would be good to have the safety of this

app/services/mapping_migration_service.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      ActiveRecord::Base.transaction do
+        mapping = create_mapping(mapping_source)
+        create_mapping_fields(mapping, template_file)
+        File.rename template_file, "#{template_file}.legacy"
+      end
+    end
+  end
+
+  def migrate_multiple_upload_integration(integration)
```
aapomm, line 70: Can you add a comment why we need this method? With specific template name examples?

app/services/mapping_migration_service.rb
```diff
@@ -0,0 +1,100 @@
...
+    legacy_mapping_reference = integration.module_parent::Mapping.legacy_mapping_reference
```
aapomm, line 71: Can you add a comment as to what plugins have a legacy mapping reference?

db/migrate/20240226193757_migrate_templates_to_mappings.rb
```diff
@@ -0,0 +1,8 @@
+class MigrateTemplatesToMappings < ActiveRecord::Migration[7.0]
+  def up
+    MappingMigrationService.new.call
+  end
+
+  def down
```
aapomm, line 6: Not important, but do we need to add a reverse of the migration service? It would

- Delete the Mapping and MappingFields
- Rename the .legacy template files to .template

app/services/mapping_migration_service.rb
```diff
@@ -0,0 +1,100 @@
...
+    integration_templates_dir = File.join(@templates_dir, integration_name)
+
+    legacy_mapping_reference.each do |source_field, legacy_template_name|
+      template_file = Dir["#{integration_templates_dir}/#{legacy_template_name}.template*"]
```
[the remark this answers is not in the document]
caitmich: This is for the case where a single template is shared (Burp) because after it's copied the first time it'll be renamed

PR #1262, opened 2024-05-30
Load all project drops and parse liquid async

aapomm:
### Spec
Currently, only the document properties, project, and team are available as global drops. This prevents us from making full use of Liquid in content and cases where a user may want to iterate over the project's issues in a content block is just not possible.

Parsing Liquid inline will start to have a performance penalty considering that we'll be pulling the project records for every #show page load, so we need the following changes:

- Load only the Liquid collections found in the content
- Parse liquid asynchronously

**Proposed solution**
Update the LiquidAssignsService class with all the possible drops available in a project. This task includes:

- Update the LiquidAssignsService
  - Add all the available project drops to the liquid_assigns
  - Use low level caching to cache the drops
  - Parse the content with Liquid::VariableLookup and only pull the liquid assigns found in the content
- Parse liquid in content asynchronously
  - Add liquid_async JS that fetches the full Liquid-parsed content from the #preview action
  - Add a spinner to indicate that the Liquid-parsed content is currently being fetched
- Specs
  - Update specs to wait for the spinner before asserting the Liquid-parsed content

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry


app/services/liquid_assigns_service.rb
```diff
@@ -15,4 +15,27 @@ def assigns
 
   def assigns_pro
   end
+
+  def project_drops
+    {
+      'evidence' => project_records(type: :evidence),
```
caitmich, line 21: This will be overwritten by any `liquid_resource_assigns` coming from the controllers when it reaches `LiquidFilter` so it will have to be named differently ex 'evidences'

app/services/liquid_assigns_service.rb
```diff
@@ -6,7 +6,7 @@ def initialize(project)
   end
 
   def assigns
-    result = { 'project' => ProjectDrop.new(project) }
+    result = project_drops
```
caitmich, line 9: what about `project_assigns` instead to be consistent with the rest of the methods?

app/services/liquid_assigns_service.rb
```diff
@@ -15,4 +15,27 @@ def assigns
...
+      'evidence' => project_records(type: :evidence),
+      'issues' => project_records(type: :issue),
+      'nodes' => project_records(type: :node),
+      'notes' => project_records(type: :note),
+      'project' => ProjectDrop.new(project),
+      'tags' => project_records(type: :tag)
+    }
+  end
+
+  def project_records(type:)
```
caitmich, line 30: What about naming this one `project_record_drops` since it creates the drops

app/services/liquid_assigns_service.rb
```diff
@@ -15,4 +15,27 @@ def assigns
...
+    records = project.send(type.to_s.pluralize)
+    records = records.user_nodes if type == :node
```
caitmich, line 32: Instead of fetching the records dynamically, what about passing them into this method as an argument so that it doesn't have to worry about specific cases like if type == :node. Then it only cares about creating the drops & caching them

app/services/liquid_assigns_service.rb
aapomm: I went with `cached_drops` instead since "caching" seems to be the main job of the method

CHANGELOG
```diff
@@ -1,6 +1,5 @@
 [v#.#.#] ([month] [YYYY])
-  - [entity]:
-    - [future tense verb] [feature]
+  - Liquid: Make project-level drops available for Liquid syntax
```
caitmich, line 2:
```suggestion
  - Liquid: Make project-level collections available for Liquid syntax
```

app/assets/javascripts/tylium/modules/liquid_async.js
```diff
@@ -0,0 +1,22 @@
+document.addEventListener('turbolinks:load', function () {
+  $('[data-behavior~=liquid-async]').each(function () {
+    const that = this,
+      data = { text: $(that).attr('data-content') },
+      $spinner = $(that).prev().find('[data-behavior~=liquid-spinner');
+
+    fetch($(that).attr('data-path'), {
+      method: 'POST',
```
caitmich, line 8: I know the route is a POST, but is there a reason that we can't use GET here?

app/services/liquid_assigns_service.rb
```diff
@@ -15,4 +18,45 @@ def assigns
...
+
+    Rails.cache.fetch(cache_key) do
+      records.map { |record| drop_class.new(record) }
+    end
+  end
+
+  def project_assigns
+    project_assigns = { 'project' => ProjectDrop.new(project) }
+
+    assigns_from_content.each do |var|
```
caitmich, line 46:
```suggestion
    assigns_from_content.each do |assigns_type|
```
Or something more descriptive?

app/services/liquid_assigns_service.rb
```diff
@@ -15,4 +18,45 @@ def assigns
...
+    return AVAILABLE_PROJECT_ASSIGNS if text.nil?
+
+    variable_lookup = Liquid::VariableLookup.parse(text)
+    return (variable_lookup.lookups & AVAILABLE_PROJECT_ASSIGNS)
+  end
+
+  def cached_drops(records)
+    return [] if records.empty?
+
+    type = records.first.class.to_s.underscore
```
caitmich, line 34: What about having `type` as a parameter to save us this line since we already have the value we need in `project_assigns` and it would simplify this method a little

app/services/liquid_assigns_service.rb
```diff
@@ -1,12 +1,15 @@
 class LiquidAssignsService
-  attr_accessor :project
+  AVAILABLE_PROJECT_ASSIGNS = %w{ evidences issues nodes notes tags }.freeze
 
-  def initialize(project)
+  attr_accessor :project, :text
+
+  def initialize(project:, text: nil)
     @project = project
```
caitmich, line 7: What about passing the `project_id` instead and finding it here?

spec/support/liquid_shared_examples.rb
```diff
@@ -9,14 +9,15 @@
     visit @path
   end
 
-  it 'dynamically renders item properties' do
+  it 'dynamically renders item properties', js: true do
+    expect(page).to have_no_css('span.text-nowrap', text: 'Loading liquid dynamic content', wait: 10)
```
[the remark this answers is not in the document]
aapomm: 10 seconds is only the maximum wait time and it won't really take this long. We're only setting the value here because [2 seconds](https://github.com/teamcapybara/capybara/blob/master/README.md#asynchronous-javascript-ajax-and-friends) might be too short and we need to account for possible hitches in CI that could cause a false negative.

app/services/liquid_assigns_service.rb
aapomm: I'm not sure if we're winning with this change if we have to update every instance of this initialization as well as possibly move the `can?` check. See `liquid_enabled_resource`:
```
  def project_assigns
    # This is required because we may be in Markup#preview that's passing
    # :project_id for Tylium rendered editors
    project = Project.find(params[:project_id])
    authorize! :use, project

    LiquidAssignsService.new(project: project, text: params[:text]).assigns
```

The check requires a reference to the `current_user`

app/services/liquid_assigns_service.rb
```diff
@@ -15,4 +18,44 @@ def assigns
 
   def assigns_pro
   end
+
+  # This method uses Liquid::VariableLookup to find all liquid variables from
+  # a given text. We use the list to know which project assign we need.
+  def assigns_from_content
+    return AVAILABLE_PROJECT_ASSIGNS if text.nil?
```
etdsoft, line 25: should the default be [] rather than a full list?

app/services/liquid_assigns_service.rb
```diff
@@ -15,4 +18,44 @@ def assigns
...
+
+    variable_lookup = Liquid::VariableLookup.parse(text)
+    return (variable_lookup.lookups & AVAILABLE_PROJECT_ASSIGNS)
```
etdsoft, line 28: seeing this code and the intersect/validation, is L25 really needed? How does VariableLookup behave on empty input?

app/services/liquid_assigns_service.rb
```diff
@@ -15,4 +18,44 @@ def assigns
...
+  end
+
+  def cached_drops(records, record_type)
+    return [] if records.empty?
+
+    cache_key = "liquid-project-#{project.id}-#{record_type.pluralize}:#{records.maximum(:updated_at).to_i}-#{records.count}"
```
etdsoft, line 34: it seems we're working hard here to craft a unique key. Have you looked into how the Rails code does this when you pass a collection in your view cache call? Meybe there's a method we can use that takes a prefix and a collection and spits the unique cache key?

app/views/evidence/show.html.erb
```diff
@@ -30,11 +30,11 @@
             <div class="note-text-inner">
               <h4 class="header-underline">
                 <span class="text-truncate" title="Evidence for this instance">Evidence for this instance</span>
+                <%= render partial: 'shared/liquid_loading' %>
                 <%= render partial: 'actions' %>
               </h4>
-
-              <div class="content-textile" data-behavior="content-textile">
-                <%= markup(@evidence.content, liquid: true) %>
+              <div class="content-textile" data-behavior="content-textile liquid-async" data-path="<%= preview_project_node_evidence_path(current_project, @evidence.node, @evidence) %>" data-content="<%= @evidence.content %>" >
```
etdsoft, line 36: should we turn this into a `content_tag` call?

I'm hesitant about the `data-content="<%%>"` call without any escaping of the user-provided content.

app/views/issues/show.html.erb
```diff
@@ -33,10 +33,11 @@
             <div class="note-text-inner">
               <h4 class="mb-4 header-underline">
                 <span class="text-truncate" title="<%= @issue.title %>"><%= @issue.title %></span>
+                <%= render partial: 'shared/liquid_loading' %>
                 <%= render partial: 'actions' %>
               </h4>
-              <div class="content-textile" data-behavior="content-textile">
-                <%= markup(@issue.text, liquid: true) %>
+              <div class="content-textile" data-behavior="content-textile liquid-async" data-path="<%= preview_project_issue_path(current_project, @issue) %>" data-content="<%= @issue.text %>" >
```
etdsoft, line 39: ditto re: escaping.

Perhaps even extracte the :content_tag call to a helper that does all the data- stuff by receiving a path and context. Perhaps not needed.

app/views/notes/show.html.erb
```diff
@@ -21,10 +21,11 @@
             <div id="<%= dom_id(@note) %>" class="note-text-inner">
               <h4 class="header-underline">
                 <span class="text-truncate" title="<%= @note.title %>"><%= @note.title %></span>
+                <%= render partial: 'shared/liquid_loading' %>
                 <%= render partial: 'actions' %>
               </h4>
-              <div class="content-textile" data-behavior="content-textile">
-                <%= markup(@note.text, liquid: true) %>
+              <div class="content-textile" data-behavior="content-textile liquid-async" data-path="<%= preview_project_node_note_path(current_project, @note.node, @note) %>" data-content="<%= @note.text %>" >
```
etdsoft, line 27: ditto

app/views/qa/issues/show.html.erb
```diff
@@ -29,8 +30,8 @@
               </span>
             </span>
           </h4>
-          <div class="content-textile" data-behavior="content-textile">
-            <%= markup(@issue.text, liquid: true) %>
+          <div class="content-textile" data-behavior="content-textile liquid-async" data-path="<%= preview_project_issue_path(current_project, @issue) %>" data-content="<%= @issue.text %>" >
```
etdsoft, line 33: and here

app/services/liquid_assigns_service.rb
aapomm: There is a case in Word where `LiquidAssignsService` is invoked globally and there is no text passed (report_processor_job.rb#L25 and the word thorfile Line 51. In those cases, we need to pass all the assigns available.

app/services/liquid_assigns_service.rb
aapomm: See comment above. If the `:text` param is not passed, we need to fetch all the assigns.

app/views/evidence/show.html.erb
aapomm: Used `content_tag` and added a partial

app/services/liquid_assigns_service.rb
etdsoft: This is what I meant:
https://api.rubyonrails.org/classes/ActiveSupport/Cache.html#method-c-expand_cache_key

PR #1266, opened 2024-06-14
Implement LiquidCachedAssigns

aapomm:
### Spec
The LiquidAssignsService does a lookahead of the content but this can be skipped if we implement a lazily loaded liquid assigns that would simplify the code.

**Proposed solution**
To lazily load the drops, we implement a LiquidCachedAssigns class to replace the LiquidAssignsService and does the following:

- quacks like a hash {}
- when it receives a message, it validates if it's associated with a list of records ('issues', 'notes', etc.)
  - if valid, it fetches the records from the project
  - if invalid, check if if it present in the current @assigns hash (to catch the case for record-based assigns)

### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry


app/services/liquid_cached_assigns.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    @assigns = { 'project' => ProjectDrop.new(project) }
+    @assigns.merge!(assigns_pro)
+  end
+
+  def [](record_type)
+    assigns[record_type] ||= cached_drops(record_type)
+  end
+
+  def key?(key)
```
caitmich, line 17: I'm assuming this is needed on the Liquid side of things?

aapomm: Yes. This is used here: https://github.com/Shopify/liquid/blob/77bc56a1c28a707c2b222559ffb0b7b1c5588928/lib/liquid/context.rb#L211. Liquid checks if the assign is present in the hash, but since we're lazily loading the assigns we need to return the available assigns even if they're not yet part of the hash.

aapomm: Added comment for future reference

app/services/liquid_cached_assigns.rb
```diff
@@ -0,0 +1,68 @@
...
+  # SEE: https://github.com/Shopify/liquid/blob/77bc56/lib/liquid/context.rb#L211
+  # Liquid is checking if the variable is present in the assigns hash by
+  # calling the `key?` method. Since we're lazily loading the keys, the variable
+  # may not yet be present in the assigns hash.
+  def key?(key)
+    AVAILABLE_PROJECT_ASSIGNS.include?(key.to_s) || assigns.key?(key)
+  end
+
+  def merge(hash)
+    LiquidCachedAssigns.new(project: project).merge!(@assigns.merge(hash))
```
etdsoft, line 26: isn't this f'ed up?

`LiquidCachedAssigns.new(project: project)` gives you exactly what you have in the context (`self`). Then you're calling merge on self's @assigns before passing to merge! which does it again (!?)

```
@assigns.merge!(hash)
```

at which point, why do you need `merge!`

app/services/liquid_cached_assigns.rb
```diff
@@ -0,0 +1,68 @@
...
+    return [] unless AVAILABLE_PROJECT_ASSIGNS.include?(record_type)
+
+    case record_type
+    when 'evidences'
+      project.evidence
+    when 'nodes'
+      project.nodes.user_nodes
+    when 'notes'
+      # FIXME - ISSUE/NOTE INHERITANCE
+      project.notes.includes(:node).where.not(node: { type_id: Node::Types::ISSUELIB })
```
etdsoft, line 63: is this a smell? Why do we exclude this Type but not the other system types?

https://github.com/dradis/dradis-ce/blob/develop/app/models/node.rb#L74

app/services/liquid_cached_assigns.rb
aapomm: > LiquidCachedAssigns.new(project: project) gives you exactly what you have in the context (self).

Not quite as the new instance `LiquidCachedAssigns.new(project: project)` will not have the `self`'s current version of `@assigns`. If we remove the `@assigns` in the `merge` method:
```
def merge(hash)
  LiquidCachedAssigns.new(project: project).merge!(hash)
end
```
The following happens:
```
irb(main):003:0> assigns1 = LiquidCachedAssigns.new(project: Project.find(1))
irb(main):004:0> assigns1.merge!({'issue' => IssueDrop})
irb(main):005:0> assigns2 = assigns1.merge({'evidence' => EvidenceDrop})
irb(main):006:0> assigns2.assigns
=> {"project"=>ProjectDrop, "evidence"=>EvidenceDrop}
```
Notice that the `{ 'issue' => IssueDrop }` is missing. This is because `LiquidCachedAssigns.new(project: project)` is instantiated without the original's `@assigns` hash (merged with the IssueDrop in the above example).

---

I do see that the multiple `merge` calls can be confusing here so some possible alternatives to the approach are:

- Adding `assigns` to the initializer params so the merge method will look like this instead:
```
def merge(hash)
  LiquidCachedAssigns.new(assigns: @assigns, project: project).merge!(hash)
end
```
- Further abstraction by adding a `dup` method. The methods will look something like:
```
def dup
  LiquidCachedAssigns.new(project: project).merge!(@assigns)
end

def merge(hash)
  dup.merge!(hash)
end
```

@etdsoft thoughts?

app/services/liquid_cached_assigns.rb
aapomm: Updated to check for node_id != project.issue_lib instead

PR #1271, opened 2024-06-18
Add AttachmentsCopier concern

aapomm:
### Spec
When moving an instance of evidence to another node the attachments are not moved. This means that if we delete the original node, the attachments are destroyed and the attachment can no longer be referenced in the evidence.

**Proposed solution**
Copy the attachment to the new node when an evidence or note are moved
- In the controller's update action, check if the node_id is being changed
- If yes, check whether the text includes any attachment syntax (!/projects/1/nodes/1/attachments/img.jpeg!)
- If yes, find the attachment by file name and node_id and copy the attachment to the new node 
- Update the evidence/note text with the reference to the new node's attachment.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/controllers/concerns/attachments_copier.rb
+++ b/app/controllers/concerns/attachments_copier.rb
@@ -7,7 +7,7 @@ module AttachmentsCopier
 
       if attachment
         new_attachment = attachment.copy_to(record.node)
-        new_filename = ERB::Util.url_encode(new_attachment.filename)
+        new_filename = new_attachment.url_encoded_filename
         new_path = full_screenshot_path.gsub(
           /nodes\/[0-9]+\/attachments\/.+/,
           "nodes/#{new_attachment.node_id}/attachments/#{new_filename}"
--- a/app/controllers/concerns/attachments_copier.rb
+++ b/app/controllers/concerns/attachments_copier.rb
@@ -1,0 +1,19 @@
+module AttachmentsCopier
+  def copy_attachments(record)
+    record.content.scan(Attachment::SCREENSHOT_REGEX).each do |screenshot_url|
+      _, _, _, _, project_id, node_id, filename, _ = screenshot_url
+
+      attachment = Attachment.find_by(filename: filename, node_id: record.node_id_was)
+
+      if attachment
+        new_attachment = attachment.copy_to(record.node)
+        new_url = screenshot_url[0].gsub(
+          /nodes\/[0-9]+\/attachments/,
+          "nodes/#{new_attachment.node_id}/attachments"
+        )
+
+        record.content = record.content.gsub(screenshot_url[0], new_url)
+      end
+    end
+  end
+end
--- a/app/controllers/concerns/attachments_copier.rb
+++ b/app/controllers/concerns/attachments_copier.rb
@@ -1,18 +1,19 @@
 module AttachmentsCopier
   def copy_attachments(record)
-    record.content.scan(Attachment::SCREENSHOT_REGEX).each do |screenshot_url|
-      _, _, _, _, project_id, node_id, filename, _ = screenshot_url
+    record.content.scan(Attachment::SCREENSHOT_REGEX).each do |screenshot_path|
+      full_screenshot_path, _, _, _, project_id, node_id, filename, _ = screenshot_path
 
-      attachment = Attachment.find_by(filename: filename, node_id: record.node_id_was)
+      attachment = Attachment.find_by(filename: CGI::unescape(filename), node_id: record.node_id_was)
 
       if attachment
         new_attachment = attachment.copy_to(record.node)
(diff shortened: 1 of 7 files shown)


app/controllers/concerns/attachments_copier.rb
```diff
@@ -0,0 +1,19 @@
+module AttachmentsCopier
+  def copy_attachments(record)
+    record.content.scan(Attachment::SCREENSHOT_REGEX).each do |screenshot_url|
+      _, _, _, _, project_id, node_id, filename, _ = screenshot_url
```
caitmich, line 4:
```suggestion
      screenshot_path, _, _, _, project_id, node_id, filename, _ = screenshot_url
```

app/controllers/concerns/attachments_copier.rb
```diff
@@ -0,0 +1,19 @@
+module AttachmentsCopier
+  def copy_attachments(record)
+    record.content.scan(Attachment::SCREENSHOT_REGEX).each do |screenshot_url|
```
caitmich, line 3: Maybe `path` is better here since it's not really a url

app/controllers/concerns/attachments_copier.rb
caitmich: We can name this part so it can be used on L10 and L15. screenshot_path is a suggestion but depends what you decide to go with on L3

app/controllers/concerns/attachments_copier.rb
```diff
@@ -0,0 +1,19 @@
...
+      _, _, _, _, project_id, node_id, filename, _ = screenshot_url
+
+      attachment = Attachment.find_by(filename: filename, node_id: record.node_id_was)
```
caitmich, line 6: We'll need to escape the filename like we do in attachment_validator https://github.com/secroots/dradis-pro/blob/2d14058431f2760904edbb2d74bfb83d4c7a12b8/engines/dradispro-word/lib/dradis/pro/plugins/word/validators/attachment_validator.rb#L108

app/controllers/concerns/attachments_copier.rb
```diff
@@ -1,18 +1,19 @@
...
 
-      attachment = Attachment.find_by(filename: filename, node_id: record.node_id_was)
+      attachment = Attachment.find_by(filename: CGI::unescape(filename), node_id: record.node_id_was)
 
       if attachment
         new_attachment = attachment.copy_to(record.node)
-        new_url = screenshot_url[0].gsub(
-          /nodes\/[0-9]+\/attachments/,
-          "nodes/#{new_attachment.node_id}/attachments"
+        new_filename = ERB::Util.url_encode(new_attachment.filename)
```
caitmich, line 10: Can we use [attachment.url_encoded_filename](https://github.com/secroots/dradis-pro/blob/2d14058431f2760904edbb2d74bfb83d4c7a12b8/app/models/attachment.rb#L238) here?

app/controllers/notes_controller.rb
```diff
@@ -44,7 +45,11 @@ def edit
   # Update the attributes of a Note
   def update
     updated_at_before_save = @note.updated_at.to_i
-    if @note.update(note_params)
+
+    @note.assign_attributes(note_params)
+    copy_attachments(@note) if @note.node_changed?
+
+    if @note.save
```
etdsoft, line 52: good consistency move.

app/controllers/concerns/attachments_copier.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      attachment = Attachment.find_by(filename: CGI::unescape(filename), node_id: record.node_id_was)
+
+      if attachment
+        new_attachment = attachment.copy_to(record.node)
+        new_filename = new_attachment.url_encoded_filename
+        new_path = full_screenshot_path.gsub(
+          /nodes\/[0-9]+\/attachments\/.+/,
+          "nodes/#{new_attachment.node_id}/attachments/#{new_filename}"
+        )
```
etdsoft, line 14: I'm concerned about this text manipulation. Then I thought about the "copy link" feature of the attachments sidebar, perhaps we can use the same approach?

https://github.com/dradis/dradis-ce/blob/main/app/views/attachments/_attachment_box.html.erb#L59

I could even consider moving this into a helper to avoid duplication, but we don't currently have an attachment_helper.rb it could fit in.

aapomm: It's tricky to DRY this considering that the contexts are different. The attachment box is in the view that has access to the route helpers while this code isn't. There's also some possible rabbit hole with the `default_url_options = /pro` once we move the link helpers out of the view. Happy to explore this more if we need this code on many more places but currently, this seems like the simplest approach.

PR #1273, opened 2024-06-26
Ensure the `Affected` column in Issues#index is updated when an affected node is renamed or merged

MattBudz:
### Summary

This PR fixes a bug where the issues table would not show an accurate list of affected nodes after an affected node has been renamed or merged.

### Check List

- [x] Added a CHANGELOG entry


app/views/issues/_table.html.erb
```diff
@@ -1,4 +1,4 @@
-<% cache ['issues-table', @all_columns, issues.map(&:id), @issues.map(&:updated_at).map(&:to_i).sort.last, @tags] do %>
+<% cache ['issues-table', @all_columns, issues.map(&:affected), issues.map(&:id), @issues.map(&:updated_at).map(&:to_i).sort.last, @tags] do %>
```
aapomm, line 1: Do we need this change? If only 1 node is updated, then the whole table's cache busts. Maybe we can go with just the per-issue caching change

app/views/issues/_table.html.erb
```diff
@@ -21,7 +21,7 @@
     </thead>
     <tbody>
       <% issues.each do |issue| %>
-        <% cache [issue, @all_columns, issue.try(:affected_count), issue.try(:state), 'issues-table', @tags] do %>
+        <% cache [issue, @all_columns, issue.try(:affected), issue.try(:state), 'issues-table', @tags] do %>
```
aapomm, line 24: Instead of using all the node details in the cache key, can we instead use the `updated_at`?

```suggestion
        <% cache [issue, @all_columns, issue.try(:affected_count), issue.affected.pluck(:updated_at), issue.try(:state), 'issues-table', @tags] do %>
```

app/views/issues/_table.html.erb
MattBudz: Without this change, the table doesn't show the correct affected column because the cache isn't busted. Do we need to cache the table if we are caching each row?

aapomm: Ah looks like I'm mistaken and the inner fragment caching are preserved and not busted: https://guides.rubyonrails.org/caching_with_rails.html#russian-doll-caching

etdsoft: @aapomm why is the outer cache key  change needed then? Affected is always defined as a column.

etdsoft (changes requested):
## Automated peer review — PR #1273

**Reviewers:** Claude (sub-agent 1), Codex (sub-agent 2, FAILED)
**Verdict:** request-changes
**Findings:** 1 High, 1 Medium, 1 Low

### What I checked

### Claude only

#### F1 [High|performance] app/views/issues/_table.html.erb:24 — N+1 query via pluck in cache key

**What:** The inner cache key uses `issue.affected.pluck(:updated_at)` to track node updates:

```ruby
issue.affected.pluck(:updated_at)
```

`pluck` always issues a SQL query, even though `affected` is preloaded (via `includes(:affected)` in the controller). With N issues on the page, this adds N extra queries just to build the cache key.

**Why:** Preloaded associations are held in memory and should be accessed without additional queries. Using `pluck` defeats the preloading and causes N+1 queries on every Issues index render.

**How to fix:** Use `map` instead to use the in-memory preloaded records:

```ruby
issue.affected.map(&:updated_at)
```

If association ordering is not guaranteed, add `.sort` for key stability:

```ruby
issue.affected.map { |n| n.updated_at.to_i }.sort
```

#### F2 [Medium|hygiene] app/views/issues/_table.html.erb:PR — duplicate commits in branch history

**What:** The branch contains six commit subjects that appear twice with different hashes (e.g., "move current_project fallback", "fix nav render", etc.). These are duplicate commits from an incomplete rebase.

**Why:** Duplicate commits clutter `git log` and `git bisect` history and will appear in the merge commit. They signal a sloppy rebase workflow.

**How to fix:** Interactively rebase before merge to remove duplicates:

```bash
git rebase -i develop
```

Remove the lower (original) copies of each duplicate commit, keeping only the final version. Force-push to the branch.

#### F3 [Low|tests] app/views/issues/_table.html.erb:0 — missing test coverage for cache invalidation

**What:** No specs verify that the Issues index correctly invalidates and re-renders the Affected column when a node is renamed or merged.

**Why:** Cache invalidation is a common source of bugs, and this fix is specifically about fixing a cache-invalidation bug. Test coverage would prevent regressions.

**How to fix:** Add a request spec to `spec/requests/issues_spec.rb`:

```ruby
it "invalidates affected column cache when a node is renamed" do
  issue = create(:issue, affected: [node])
  get issues_path
  expect(response.body).to include(node.name)
  
  node.update(name: "new-name")
  get issues_path
  expect(response.body).to include("new-name")
end
```

One example is sufficient.

### Notes

**Codex sub-agent failed:** binary not found. All findings are from Claude. The N+1 query via `pluck` (F1) is the only blocker — it must be fixed before merge. Cleaning up the duplicate commits (F2) is recommended for history cleanliness but not a blocker. Test coverage (F3) is optional.

---

_Reviewed automatically. Raw outputs in `agents/product-context/reviews/dradis-ce/1273/review/`._

PR #1274, opened 2024-06-26
Test manual ci with fallback branch in workflow

MattBudz:
### Summary

PoC for adding a fallback branch for rubocop to compare against when trigger CI manually.

### Check List

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
       IS_MERGE: ${{ github.base_ref == '' }}
       PREV_SHA: ${{ github.event.before }}
       TARGET_BRANCH: ${{ github.base_ref }}
-      FALLBACK_BRANCH: 'develop'
+      FALLBACK_BRANCH: 'origin/develop'
     steps:
       - name: Checkout code
         uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -40,7 +41,7 @@ jobs:
         with:
           bundler-cache: true
       - name: Lint changed files
-        run: bin/rubocop-ci ${{ env.TARGET_BRANCH || env.PREV_SHA }} ${{ env.IS_MERGE }}
+        run: bin/rubocop-ci ${{ env.TARGET_BRANCH || env.PREV_SHA || env.FALLBACK_BRANCH }} ${{ env.IS_MERGE }}
```
caitmich, line 44: What if we have a different env (not `IS_MERGE`) when it's triggered manually and then we can set 'develop' as the branch to use in that case?

.github/workflows/ci.yml
```diff
@@ -32,6 +32,7 @@ jobs:
       IS_MERGE: ${{ github.base_ref == '' }}
       PREV_SHA: ${{ github.event.before }}
       TARGET_BRANCH: ${{ github.base_ref }}
+      FALLBACK_BRANCH: 'origin/develop'
```
caitmich, line 35: Is this needed since we have [this](https://github.com/secroots/dradis-pro/blob/981dad091512db7e5fcd3790999857216bf8c854/bin/rubocop-ci#L9) check in rubocop-ci

PR #1285, opened 2024-08-12
Add My Tasks view & dashboard widget

nicolachr:
### Summary

Currently, when methodology cards are assigned to users, the assigned users do not have a way to see which cards are assigned to them unless they manually go through each card. This is tedious and prone to letting cards fall through the cracks.

The proposed solution, allows the users to:
1. Use the dashboard widget to see the five tasks with the highest priority that are assigned to them. Highest priority is given to tasks that are overdue, and lowest priority to tasks that don't have a due date.
2. Access a dedicated "My Tasks" view that displays all the tasks that are assigned to them in the form of a table.

This will allow users to have better command of their tasks, make the process of finding a task easier, and will help the users stay on top of things.

### Copyright assignment

### Check List

- [x] Added a CHANGELOG entry


app/controllers/projects_controller.rb
```diff
@@ -16,12 +17,23 @@ def show
     @authors       = [current_user]
     @boards        = current_project.methodology_library.boards
     @issues        = current_project.issues.includes(:tags).sort
-    @methodologies = current_project.methodology_library.notes.map{|n| Methodology.new(filename: n.id, content: n.text)}
+    @methodologies = current_project.methodology_library.notes.map { |n| Methodology.new(filename: n.id, content: n.text) }
     @nodes         = current_project.nodes.in_tree
     @tags          = current_project.tags
 
+    project_cards = @boards.map(&:cards)[0]
```
caitmich, line 24: This will only pick up the cards from the first board in the project. Instead we can move this to a helper and do:

L23: `@tasks = assigned_tasks`
```
def assigned_tasks
  @assigned_tasks ||= begin
    cards = @boards.flat_map(&:cards)
    user_cards = cards.select { |card| card.assignees.include? current_user }
    user_cards.sort_by { |card| card.due_date || Float::INFINITY }
  end
end
```

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,40 @@
+class TasksController < AuthenticatedController
+  include ProjectScoped
+
+  skip_before_action :set_project, if: -> { params[:project_id].blank? }
+  skip_before_action :set_nodes, if: -> { params[:project_id].blank? }
+
+  layout :set_layout
```
caitmich, line 7:
```suggestion
  layout :set_layout
  
```

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  def index
+    @default_columns = ['Title', 'Methodology', 'Due Date', 'Assigned']
+
+    if params[:project_id].present?
+      cards = current_project.methodology_library.boards.map(&:cards)[0]
+      @local_storage_key = "project.ce.project_#{current_project.id}.tasks_datatable"
```
caitmich, line 13: is the `ce` needed here?

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  skip_before_action :set_project, if: -> { params[:project_id].blank? }
+  skip_before_action :set_nodes, if: -> { params[:project_id].blank? }
+
+  layout :set_layout
+  def index
+    @default_columns = ['Title', 'Methodology', 'Due Date', 'Assigned']
+
+    if params[:project_id].present?
+      cards = current_project.methodology_library.boards.map(&:cards)[0]
```
caitmich, line 12: Ditto. This can be moved to a `before_action`

```
def set_tasks
  @tasks ||= begin
    cards = current_user.cards

    if params[:project_id].present?
      cards.select { |card| card.project.id == current_project.id }
    else
      cards
    end
  end
end
```

app/controllers/tasks_controller.rb
caitmich: What about `"project#{params[:project_id]}.tasks_datatable"`
This would work regardless of whether `params[:project_id]` is present or not?

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    else
+      []
+    end
+  end
+
+  protected
+
+  # ProjectScoped always calls current_project. We are overwriting it here to
+  # prevent errors in Pro when viewing tasks outside of projects.
+  def current_project
```
caitmich, line 31: Instead of including it and opting out of several methods, what if we don't include `ProjectScoped` and call only the methods we need when we need them?

ex. Can we instead look up the project here using `params[:project_id]` if it's present?

app/views/layouts/tylium/_sidebar.html.erb
```diff
@@ -19,6 +19,11 @@
           <i class="fa-brands fa-trello sidebar-link-icon"></i> <span class="sidebar-link-label">Methodologies</span>
         <% end %>
       <% end %>
+      <%= content_tag :li, :class => controller_path.include?('tasks') ? 'active' : '' do %>
```
caitmich, line 22:
```suggestion
      <%= content_tag :li, :class => controller_path == 'tasks' ? 'active' : '' do %>
```

app/views/projects/tasks/_summary.html.erb
```diff
@@ -0,0 +1,38 @@
...
+      <div class="action">
+        <%= link_to project_tasks_path(current_project) do %>
+          <i class="fa-solid fa-tasks sidebar-link-icon"></i> All Tasks
+        <% end %>
+      </div>
+    </span>
+  </h4>
+  <% if @tasks.any? %>
+    <ul class="list-group tasks">
+      <% @tasks.first(@tasks_limit).each do |task| %>
```
caitmich, line 13: How about limiting this list in the controller instead and having `TASKS_LIMIT` as a constant

app/views/projects/tasks/_summary.html.erb
```diff
@@ -0,0 +1,38 @@
...
+          <% end %>
+          <span class="small text-muted"><%= task.board.name %></span>
+          <%= due_date_badge(task) %>
+        </li>
+      <% end %>
+    </ul>
+
+    <p class="small mt-2 mb-0 text-center">&nbsp;
+      <% if @tasks.size > @tasks_limit %>
+        + <%= @tasks.size - @tasks_limit %> more
```
caitmich, line 26: This could also be done in the controller or the view helper using something like `overflow_tasks`

`overflow_tasks.any?` /  `overflow_tasks.count`

app/views/tasks/index.html.erb
```diff
@@ -0,0 +1,51 @@
...
+          <th>Assigned</th>
+        </tr>
+      </thead>
+      <tbody>
+        <% @tasks.each do |task| %>
+          <tr>
+            <td data-sort="<%= task.name %>">
+              <%= link_to task.name, project_board_list_card_path(task.project, task.board, task.list, task) %>
+            </td>
+            <% unless params[:project_id] %>
```
caitmich, line 26: Ditto to the 2-controller comment above

app/controllers/tasks_controller.rb
MattBudz: This is consistent with the other local storage keys we define. In CE we need `project.ce.project_#.tasks_databale` and in pro, we need `pro.tasks_datatable`

app/controllers/tasks_controller.rb
MattBudz: I tried that but we also need access to current_project in the views and ProjectScoped also sets `@nodes` for us (used by the main sidebar).

This is also the approach we take in https://github.com/dradis/dradis-ce/blob/develop/app/controllers/notifications_controller.rb

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    return if params[:project_id].blank?
+
+    super
+  end
+
+  def set_layout
+    params[:project_id].present? ? 'tylium' : 'mintcreek'
```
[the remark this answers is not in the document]
MattBudz: This is the same approach we take with notifications_controller but this should be in the pro-changes branch. Moved.

app/helpers/tasks_helper.rb
```diff
@@ -0,0 +1,46 @@
+module TasksHelper
+  TASKS_LIMIT = 5
+
+  def all_assigned_tasks
+    @assigned_tasks ||= begin
+      tasks = current_user.cards
+        # Using Arel.sql to sort the records by due_date with null due_date records last
+        .order(Arel.sql('due_date IS NULL, due_date ASC'))
```
caitmich, line 8:
```suggestion
      # Using Arel.sql to sort the records by due_date with null due_date records last
      tasks = current_user.cards.order(Arel.sql('due_date IS NULL, due_date ASC'))
```

app/helpers/tasks_helper.rb
```diff
@@ -0,0 +1,46 @@
+module TasksHelper
+  TASKS_LIMIT = 5
+
+  def all_assigned_tasks
+    @assigned_tasks ||= begin
+      tasks = current_user.cards
```
caitmich, line 6:
```suggestion
      cards = current_user.cards
```

app/helpers/tasks_helper.rb
```diff
@@ -0,0 +1,46 @@
...
+        # Using Arel.sql to sort the records by due_date with null due_date records last
+        .order(Arel.sql('due_date IS NULL, due_date ASC'))
+
+      if params[:project_id]
+        return tasks.select { |card| card.project.id == current_project.id }
+      end
+
+      tasks
+    end
```
caitmich, line 15:
```suggestion
      if params[:project_id]
        tasks.select { |card| card.project.id == current_project.id }
      else
        tasks
      end
    end
```

app/helpers/tasks_helper.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def assigned_tasks
```
caitmich, line 18: what about `assigned_tasks_for_widget` or something to make it clear that this is a limited list?

app/helpers/tasks_helper.rb
```diff
@@ -0,0 +1,46 @@
+module TasksHelper
+  TASKS_LIMIT = 5
+
+  def all_assigned_tasks
+    @assigned_tasks ||= begin
```
caitmich, line 5:
```suggestion
    @all_assigned_tasks ||= begin
```

app/helpers/tasks_helper.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      tasks
+    end
+  end
+
+  def assigned_tasks
+    all_assigned_tasks.first(TASKS_LIMIT)
+  end
+
+  def overflow_tasks_count
```
caitmich, line 22: #suggestion 

`overflow_tasks_count` leads you to expect a number, so what about returning `0` on L23 and checking for > 0 in the view?

app/controllers/tasks_controller.rb
MattBudz: I removed the pro bit since that should only be in pro. It's moved to the pro-changes branch.

app/views/tasks/index.html.erb
MattBudz: Because this is still a basically identical view shared between layouts, I went with the notifications_controller approach which also shares a view between layouts. Happy to change this if you feel strongly about it.

app/views/projects/tasks/_summary.html.erb
```diff
@@ -0,0 +1,38 @@
...
+          <i class="fa-solid fa-tasks sidebar-link-icon"></i> All Tasks
+        <% end %>
+      </div>
+    </span>
+  </h4>
+  <% if @tasks.any? %>
+    <ul class="list-group tasks">
+      <% @tasks.each do |task| %>
+        <li class="d-flex flex-column list-group-item">
+          <%= link_to project_board_list_card_path(current_project, task.board, task.list, task), class: "list-group-item-action text-truncate" do %>
```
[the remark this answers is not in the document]
MattBudz: We can add a switch statement or a new helper method for this once we introduce another record type as a task

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+  def index
+    @default_columns = ['Title', 'Methodology', 'Due Date', 'Assigned']
+    @tasks = assigned_cards
+
+    if params[:project_id].present?
+      @local_storage_key = "project.ce.project_#{current_project.id}.tasks_datatable"
+    end
+  end
+
+  protected
```
caitmich, line 17: Any reason for choosing `protected` over `private` here?

MattBudz: flipped to `private` I'm not sure why I went with `protected` it's possible I got it from the notifications controller (was exploring that implementation initially)

app/helpers/tasks_helper.rb
```diff
@@ -0,0 +1,42 @@
+module TasksHelper
+  TASKS_LIMIT = 5
+
+  def assigned_cards(project_id = nil)
+    @assigned_cards ||= begin
+      # Using Arel.sql to sort the records by due_date with null due_date records last
+      cards = current_user.cards.order(Arel.sql('due_date IS NULL, due_date ASC'))
+
+      if project_id
+        cards.select { |card| card.project.id == current_project.id }
```
caitmich, line 10: If we're passing project_id we should use it here 
```suggestion
        cards.select { |card| card.project.id == project_id }
```

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,26 @@
+class TasksController < AuthenticatedController
+  include ProjectScoped
+  include TasksHelper
+
+  skip_before_action :set_project, unless: -> { current_project }
+  skip_before_action :set_nodes, unless: -> { current_project }
```
etdsoft, line 6: These lines are at odds w/ L2, no? Under what conditions we won't have a project?

app/controllers/projects_controller.rb
```diff
@@ -16,12 +17,13 @@ def show
     @authors       = [current_user]
     @boards        = current_project.methodology_library.boards
     @issues        = current_project.issues.includes(:tags).sort
-    @methodologies = current_project.methodology_library.notes.map{|n| Methodology.new(filename: n.id, content: n.text)}
+    @methodologies = current_project.methodology_library.notes.map { |n| Methodology.new(filename: n.id, content: n.text) }
     @nodes         = current_project.nodes.in_tree
     @tags          = current_project.tags
+    @tasks         = assigned_cards(current_project.id)
```
etdsoft, line 23: This approach is a bit backwards and inconsistent w/ what we have for other records.

Also, we typically go from the project to the nested resources at the controllers, we don't delegate that searchto the view

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def index
+    @default_columns = ['Title', 'Methodology', 'List', 'Due Date', 'Assigned']
+
+    if current_project
```
etdsoft, line 11: This seems to be a smell? Reusing a controller for in/out of project context?

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+      @local_storage_key = "project.ce.project_#{current_project.id}.tasks_datatable"
```
etdsoft, line 12: Another smell, a local storage key is typically a view concern. Why is it in the controller?

```bash
› grep -r local_storage_key app/controllers
```

Didn't yield other instances of this.

app/controllers/tasks_controller.rb
etdsoft: What we win in conciseness (less files) we're loosing in readability of the code.

app/helpers/tasks_helper.rb
```diff
@@ -0,0 +1,42 @@
+module TasksHelper
+  TASKS_LIMIT = 5
```
etdsoft, line 2: Even if we want to hard code, and not make this configurable, a Helper seems to be the wrong place for this constant?

Perhaps the `Configuration` class?

app/helpers/tasks_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def assigned_cards(project_id = nil)
+    @assigned_cards ||= begin
+      # Using Arel.sql to sort the records by due_date with null due_date records last
```
etdsoft, line 6: I'd assume due_date null would be grouped at the beginning or the end of the regular `order(:due_date)`, no? Then it's a matter of choosing :desc / :asc and you get the same behavior w/ less code?

app/helpers/tasks_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+      cards = current_user.cards.order(Arel.sql('due_date IS NULL, due_date ASC'))
+
+      if project_id
```
etdsoft, line 9: 🚨

app/helpers/tasks_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+        cards.select { |card| card.project.id == current_project.id }
+      else
+        valid_project_ids = Project.kept.current.accessible_by(current_ability).ids
+        cards.select { |card| card.project.id.in? valid_project_ids }
```
etdsoft, line 13: This is controller code, not view-helper code.

app/views/projects/tasks/_summary.html.erb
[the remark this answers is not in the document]
nicolachr: Closing this as we'll be focusing only on cards for now

app/controllers/tasks_controller.rb
MattBudz: moved to the view

app/controllers/tasks_controller.rb
MattBudz: refactored to be in project always in CE

app/controllers/tasks_controller.rb
MattBudz: removed the pro only code from CE and kept ProjectScoped since the navbar is expecting current_project to be available for certain nav-links

app/helpers/tasks_helper.rb
MattBudz: removed since CE always is within a project context

app/helpers/tasks_helper.rb
MattBudz: this was just for presentation, but more geared towards the pro dashboard for consistency there. I have removed this from CE since it's introducing an inconsistency in the project-level dashboard.

etdsoft (changes requested):
## Code Review — PR #1285

**Verdict:** request-changes (1 High, 1 Medium, 3 Low)

---

## Summary

The PR introduces a useful "My Cards" dashboard widget and tasks index, but the `assigned_tasks` concern has a structural N+1 problem that will fire O(N-cards) queries on every project dashboard load. The view compounds this by re-accessing associations that were not preloaded. Tests cover the happy and empty-state paths but are missing isolation coverage and a dashboard widget spec.

## Findings

### F1 [High|performance] app/controllers/concerns/tasks.rb:7 — N+1 queries for assignees fire on every project dashboard load

**What:** `card.assignees.include?(current_user)` is called inside Ruby `select` for every card in the project. `assignees` is a HABTM association on `Card` and is not included in `includes(lists: :cards)`. Each call fires a separate `SELECT` against the join table. With 100 cards, that is 100 extra queries — on every `ProjectsController#show` render, for every user.

The view compounds this: `task.assignees.each` in `tasks/index.html.erb:42` fires the same N+1 again on the index page, and `task.project` in line 35 traverses `card.list.board.project` via delegates, which are also not preloaded from the card's perspective.

**Why:** The eager load is board-centric (`Board.includes(lists: :cards)`). The flattened card objects have their `list` and `board` delegates resolved through `belongs_to :list`, and `assignees` is a HABTM — neither is backfilled by AR onto the individual card instances from this include chain. The result is O(N) queries per request, hidden behind Ruby iteration.

**How to fix:** Rewrite the concern to load cards card-centrically with full includes, then push the assignee filter into SQL:

```ruby
def assigned_tasks
  Card
    .joins(list: :board)
    .where(boards: { project_id: current_project.id })
    .joins(:assignees)
    .where(users: { id: current_user.id })
    .includes(:assignees, list: :board)
    .sort_by { |card| [card.due_date.nil? ? 1 : 0, card.due_date] }
end
```

With this approach every touched association (`list`, `board`, `assignees`) is preloaded from the card's perspective. In the view, replace `task.project` (line 35) with `current_project` — all tasks are scoped to it already and the delegate chain fires redundant queries for every row.

### F2 [Medium|tests] spec/features/tasks_spec.rb — missing isolation test and no dashboard widget coverage

**What:** Two gaps: (1) There is no spec asserting that cards assigned to other users in the same project are hidden from the current user. The "with assigned tasks" context only checks that the assigned card appears, not that a different user's card does not appear. (2) The dashboard widget (`projects/tasks/_summary.html.erb`) added to `projects/show` has no test at all.

**Why:** The core security property of this feature is that users see only their own cards. Without a negative assertion (assign a card to user B, log in as user A, assert the card is absent), a future regression in the `current_user` filter would go undetected by CI. The dashboard widget is wired into `ProjectsController#show` via `@tasks = assigned_tasks`; any breakage there (nil render, route error, missing partial) is also currently uncovered.

**How to fix:** Add a context to the existing spec:

```ruby
context 'with tasks assigned to another user' do
  it 'does not render cards assigned to other users' do
    other_user = create(:user)
    @card.assignees = [other_user]
    @card.save

    visit project_tasks_path(current_project)
    expect(page).not_to have_selector('td a', text: @card.name)
  end
end
```

And a smoke-test for the dashboard widget in a `projects/show` spec (or the existing `projects_spec.rb` if one exists):

```ruby
it 'renders the My Cards widget when cards are assigned to the user' do
  @card.assignees = [current_user]
  @card.save
  visit project_path(current_project)
  expect(page).to have_content('My Cards')
  expect(page).to have_selector('ul.tasks')
end
```

### F3 [Low|consistency] app/assets/stylesheets/hera/views/projects.scss:114 — border uses px unit

**What:** `border: 1px solid var(--border-color)` uses a `px` value.

**Why:** `CLAUDE.md` CSS rules require relative units (`rem`, `em`, `ch`) and explicitly say "avoid px".

**How to fix:** Use `border: 0.0625rem solid var(--border-color)` (1px ÷ 16 = 0.0625rem), or use a SASS variable if one exists for hairline borders in `smacss/variables.scss`.

### F4 [Low|hygiene] spec/features/tasks_spec.rb:1 — spec title says "My Tasks", feature is "My Cards"

**What:** `describe 'My Tasks'` at line 1, but the breadcrumb, page title, and widget header all use "My Cards".

**Why:** When the name diverges between spec and UI, searching for the feature by name in the test suite becomes confusing and the spec reads as covering a different thing than it tests.

**How to fix:** Change to `describe 'My Cards'`.

### F5 [Low|tests] spec/features/tasks_spec.rb:26 — `User.first` used instead of the logged-in user

**What:** `@card.assignees = [User.first]` in the "with assigned tasks" context assumes the logged-in user is the first user in the database.

**Why:** `login_to_project_as_user` creates and logs in a specific user. In a parallel test environment, or if factories for other users were called earlier in the suite, `User.first` may not be the authenticated user, causing the card to be invisible to the logged-in user and the test to fail intermittently.

**How to fix:** Use the user the helper exposes. Based on the project's login helpers: `@card.assignees = [current_user]` (the same `current_user` helper available in feature specs after `login_to_project_as_user`).

## Notes

- Codex sub-agent did not run (binary not found on system). Review is based on Claude Code only.
- The rename of `mark_notifications_from_assignments` → `mark_notifications_from_tasks` in `notifications_reader_job.rb` is a private method with a single internal call site; it is safe.
- `fa-solid fa-tasks` is used in `_summary.html.erb` for the "All Cards" link icon. In Font Awesome 6 (which the project uses per `CLAUDE.md`), `fa-tasks` was renamed to `fa-list-check`. Worth verifying the icon renders rather than silently producing an empty space.
- The `due_date_badge` helper's `case/when` with a lambda is valid Ruby (`===` calls the proc) and the ordering (today → overdue → future) is logically correct.
- The `data-local-storage-key` in the table uses `current_project.id`, which correctly scopes the DataTables state per project.

PR #1289, opened 2024-09-03
Feature issue entry association diff

aapomm:
### Summary

- Backported CSS changes from Pro
- Hide IssueLib widget
- Implement DiffedContent and specs
- Add `delete_field` method for the HasFields concern
- Implement `fields_has_to_source` to convert the hash of fields to Dradis-style item content.

### Check List

### DIFF
--- a/app/models/concerns/has_fields.rb
+++ b/app/models/concerns/has_fields.rb
@@ -56,7 +56,7 @@ module HasFields
         updated_fields[field] = value
         self.send(
           :"#{container_field}=",
-          updated_fields.to_a.map { |h| "#[#{h[0]}]#\n#{h[1]}" }.join("\n\n")
+          FieldParser.fields_hash_to_source(updated_fields)
         )
       end
 
@@ -71,5 +71,4 @@ module HasFields
       end
     end
   end
-
 end
--- a/app/models/concerns/has_fields.rb
+++ b/app/models/concerns/has_fields.rb
@@ -20,6 +20,15 @@ module HasFields
     fields["Title"].present?
   end
 
+  private
+
+  def update_container(container_field, updated_fields)
+    self.send(
+      :"#{container_field}=",
+      FieldParser.fields_hash_to_source(updated_fields)
+    )
+  end
+
   module ClassMethods
     # Method for models to define which attribute is to be converted to fields.
     #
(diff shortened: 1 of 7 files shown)


app/views/layouts/tylium.html.erb
```diff
@@ -94,6 +94,10 @@
       <%= render partial: 'tags/modals/form' %>
     <% end %>
 
+    <% if defined?(Dradis::Pro) && (controller_name == 'issues' && action_name == 'show') %>
+      <%= render partial: 'dradis/pro/plugins/issuelib/issues/diff_modal' %>
+    <% end %>
```
MattBudz, line 99:
```suggestion
```

app/views/layouts/tylium.html.erb
```diff
@@ -94,6 +94,10 @@
...
+
```
MattBudz, line 100:
```suggestion
```

Let's leave this out. We have other modals we are only rendering in Pro by simply omitting them in the CE version of this file.

app/services/diffed_content.rb
```diff
@@ -0,0 +1,53 @@
...
+    @source = source
+    @target = target
+
+    normalize_content(@source)
+    normalize_content(@target)
+  end
+
+  def diff
+    Differ.format = :html
+    differ_result = Differ.diff_by_word(source.content, target.content)
```
etdsoft, line 14: Why do we need a custom format rather than using the existing approach we have for Revision history?

https://github.com/dradis/dradis-ce/blob/develop/app/models/diffed_revision.rb#L12-L16

app/models/concerns/has_fields.rb
```diff
@@ -56,10 +56,19 @@ def dradis_has_fields_for(container_field)
         updated_fields[field] = value
```
etdsoft, line 56: Consider updating spacing in L55

app/models/concerns/has_fields.rb
```diff
@@ -56,10 +56,19 @@ def dradis_has_fields_for(container_field)
...
+        )
+      end
+
+      # Completely removes the field (field header and value) from the content
+      define_method :delete_field do |field|
+        updated_fields = fields
+        updated_fields.except!(field)
+        self.send(
+          :"#{container_field}=",
+          FieldParser.fields_hash_to_source(updated_fields)
```
etdsoft, line 69: this is the same as L57-L70, is it worth extracting into a (private?) helper (e.g. #save_fields - mave "save" isn't right; #update_container)?

app/models/field_parser.rb
etdsoft, line 1: I don't recall having reviewed this class before but several questions come to mind.

1. ".source_to_fields" and ".source_to_fields_array" sound a lot like one should be colling the other, instead you're duplicating code.
2. what's parse_fieldless_string used for?

I first noticed the use of "source" which I wasn't a fan of, but fine, it's our own internal speak. But then in the last method you use "_string" instead of "_source" and I don't that works (arguably I don't understand that method either).

aapomm: > ".source_to_fields" and ".source_to_fields_array" sound a lot like one should be colling the other, instead you're duplicating code.

Fixed ✅ 

> what's parse_fieldless_string used for?

This method handles strings that are out side the Dradis fields format (`#[Field]#\nValue\n...`). This method solves the problem: if I have the following content in an issue
```
asdf

#[Title]#
Issue#1
```
How will `asdf` appear when the user switches to the Fields view?

> I first noticed the use of "source" which I wasn't a fan of, but fine, it's our own internal speak. But then in the last method you use "_string" instead of "_source" and I don't that works (arguably I don't understand that method either).

I've renamed the method to `parse_fields_without_headers(source)` to better describe the method.

PR #1290, opened 2024-09-03
Touch evidence, children and notes on node merge & add affected_ids to issues#index table cache

caitmich:
### Summary

After merging nodes, 
- The issues#index table cache doesn't update because we are not including `:affected` in the cache keys. This leads to broken nodes#show links in the affected column
- The issues/evidence#index table cache doesn't bust because the evidence `updated_at` does not get updated on node merge since we are using `update_all`. This leads to broken #edit links (since they point to the old node)

This PR adds `:affected_ids` to the issues#index table cache key and adds `updated_at: Time.current` to the list of columns to update in [merger.rb](https://github.com/dradis/dradis-ce/blob/e6539a7b2f69feb36b09b728fdef05aa7b901387/app/services/nodes/merger.rb#L43)'s `move_descendents` method

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/services/nodes/merger.rb
+++ b/app/services/nodes/merger.rb
@@ -31,66 +31,70 @@ class Nodes::Merger
 
   private
 
-    attr_accessor :target_node, :source_node, :copied_attachments
+  attr_accessor :target_node, :source_node, :copied_attachments
 
-    DESCENDENT_RELATIONSHIPS = {
-      activities: :trackable_id,
-      children: :parent_id,
-      evidence: :node_id,
-      notes: :node_id
-    }.freeze
+  DESCENDENT_RELATIONSHIPS = {
+    activities: :trackable_id,
+    children: :parent_id,
+    evidence: :node_id,
+    notes: :node_id
+  }.freeze
 
-    def move_descendents
-      DESCENDENT_RELATIONSHIPS.each do |relation, column|
-        source_node.send(relation).update_all(column => target_node.id)
-      end
+  def move_descendents
+    DESCENDENT_RELATIONSHIPS.each do |relation, column|
+      # update_all doesn't update timestamps so we need to do it manually
+      source_node.send(relation).update_all(
+        column => target_node.id,
+        :updated_at => Time.current
+      )
     end
+  end
 
-    def reset_counter_caches
-      Node.reset_counters target_node.id, :children_count
-    end
+  def reset_counter_caches
+    Node.reset_counters target_node.id, :children_count
+  end
 
-    def update_properties
-      source_node.properties.each do |key, value|
-        case key.to_sym
-        when :services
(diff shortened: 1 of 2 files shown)


app/services/nodes/merger.rb
```diff
@@ -31,66 +31,74 @@ def call
...
+    DESCENDENT_RELATIONSHIPS.each do |relation, column|
+      if relation == :activities
         source_node.send(relation).update_all(column => target_node.id)
+      else
+        # update_all doesn't update timestamps so we need to do it manually
+        source_node.send(relation).update_all(
+          column => target_node.id,
+          :updated_at => Time.current
+        )
       end
```
caitmich, line 53: Main changes. The rest is spacing updates

app/views/issues/_table.html.erb
```diff
@@ -1,4 +1,4 @@
-<% cache ['issues-table', @all_columns, issues.map(&:id), @issues.map(&:updated_at).map(&:to_i).sort.last, @tags] do %>
+<% cache ['issues-table', @all_columns, @issues.map(&:affected_ids), issues.map(&:id), @issues.map(&:updated_at).map(&:to_i).sort.last, @tags] do %>
```
aapomm, line 1: Should issues be `touched` when nodes are updated so we don't have to add this to the cache key?

caitmich: We already have `touch: true` in the `belongs_to` association in the evidence model, and since issues and nodes are connected by evidence, this should suffice in most cases. The reason this is a concern is specific to the merge action, which uses `update_all` so no callbacks are run

caitmich: In order to fix this we could call `update` instead like you mentioned above, but I am concerned about performance implications

etdsoft: You can call .touch from the merge code.

app/services/nodes/merger.rb
etdsoft: I think this is a sensible suggestion, that was unaddressed from the peer review.

etdsoft (changes requested):
Not sure on this approach, either #update or calling #touch would apparently address the issue w/o requiring cache key updates.

app/services/nodes/merger.rb
```diff
@@ -31,66 +31,68 @@ def call
...
+    evidence: :node_id,
+    notes: :node_id
+  }.freeze
+
+  def move_descendents
+    DESCENDENT_RELATIONSHIPS.each do |relation, column|
+      collection = source_node.send(relation)
+      collection.update_all(column => target_node.id)
+      collection.touch_all
     end
```
caitmich: Main change. The rest is spacing

PR #1291, opened 2024-09-04
Update evidence screenshot links after node merge

caitmich:
### Summary
After merging nodes, screenshots in the Evidence content are left pointing to the old node. The attachments are copied over but the links are not updated. This means users face validation errors and have to manually update all screenshot references after merging nodes.

### Check List

- [x] Added a CHANGELOG entry


app/services/nodes/merger.rb
```diff
@@ -90,15 +92,33 @@ def update_properties
...
-      copied_attachments << attachment.copy_to(target_node)
+      new_attachment = attachment.copy_to(target_node)
+      copied_attachments[attachment.filename] = new_attachment
+    end
+    update_attachment_references
+  end
+
+  def update_attachment_references
+    target_node.evidence_ids.each do |evidence_id|
+      evidence = Evidence.find(evidence_id)
```
aapomm, line 106: What about
```suggestion
    target_node.evidence.each do |evidence|
```

app/services/nodes/merger.rb
```diff
@@ -90,15 +92,33 @@ def update_properties
   end
 
   def copy_attachments
```
aapomm, line 94: This class is including the `AttachmentsCopier` that defines a `copy_attachments` method but this also defines a method with the same name. Should we differentiate the 2 and use a different name?

app/controllers/concerns/attachments_copier.rb
```diff
@@ -7,14 +7,16 @@ def copy_attachments(record)
...
-          "nodes/#{new_attachment.node_id}/attachments/#{new_filename}"
-        )
-
-        record.content = record.content.gsub(full_screenshot_path, new_path)
+        record.content = updated_record_content(record.content, full_screenshot_path, new_attachment)
       end
     end
   end
+
+  def updated_record_content(record_content, full_screenshot_path, new_attachment)
```
aapomm, line 15: `updated_record_content` doesn't quite describe what this method does. What about `update_attachment_link` or `update_attachment_reference`, or something else?

app/services/nodes/merger.rb
```diff
@@ -90,15 +92,33 @@ def update_properties
...
+      new_attachment = attachment.copy_to(target_node)
+      copied_attachments[attachment.filename] = new_attachment
+    end
+    update_attachment_references
+  end
+
+  def update_attachment_references
+    target_node.evidence_ids.each do |evidence_id|
+      evidence = Evidence.find(evidence_id)
+      evidence.content.scan(Attachment::SCREENSHOT_REGEX).each do |screenshot_path|
```
aapomm, line 107: This is very similar to the `AttachmentsCopier#copy_attachments` method. Is there something we can reuse from there?

app/controllers/concerns/attachments_copier.rb
```diff
@@ -1,9 +1,15 @@
 module AttachmentsCopier
-  def copy_attachments(record)
+  # Scans record content and for each attachment reference found,
+  # 1. Copies the attachment to the new node
+  # 2. Updates the content to point to the new attachment
+  # Returns an array of copied attachments for that record
+  def copy_attachments(record, previous_node_id = nil)
```
gcelanji: I think using` source_node_id = nil` makes more sense here. `previous_node_id` feels a bit confusing.

etdsoft (changes requested):
## Automated peer review — PR #1291

**Reviewers:** Claude (sub-agent 1), Codex (sub-agent 2, FAILED)
**Verdict:** request-changes
**Findings:** 1 High, 2 Low, 1 Medium

### What I checked

### Claude only

#### F1 [Medium|correctness] app/services/attachments_copier.rb — source_node_id assignment inside loop

**What:** In `copy_attachments`, the source node ID is assigned inside the `scan.each` loop:

```ruby
record.content.scan(Attachment::SCREENSHOT_REGEX).each do |screenshot_path|
  full_screenshot_path, _, _, _, project_id, node_id, filename, _ = screenshot_path
  source_node_id ||= record.node_id_was   # ← inside the loop
```

The `||=` no-ops after the first iteration, but the placement inside the loop is confusing and implies the value might differ per-iteration.

**Why:** Variable assignment belongs at the scope where it logically applies. Placing it inside the loop makes the code harder to read and suggests a value that doesn't change might vary per-iteration.

**How to fix:** Move the assignment before the loop:

```ruby
source_node_id = record.node_id_was
record.content.scan(Attachment::SCREENSHOT_REGEX).each do |screenshot_path|
  full_screenshot_path, _, _, _, project_id, node_id, filename, _ = screenshot_path
  # ... rest of loop
```

#### F2 [High|tests] spec/services/nodes/merger_spec.rb — rollback specs check unmutated in-memory objects, not DB state

**What:** The rollback tests for evidence and issues check the original Ruby objects created in the test, not reloaded DB state:

```ruby
# spec: "does not update evidence screenshot links to point to new node"
expect(evidence.content).to include("/nodes/#{source_node.id}/...")
```

Inside `update_attachments`, the code loads `target_node.evidence.each` — a different Ruby object from the DB. The test never touched the local `evidence` variable, so it passes because the object was never mutated in memory, not because the transaction rolled back.

**Why:** Tests that don't reload from the database don't actually verify database state after rollback. The tests give false confidence that rollback works when they're really just testing that the original Ruby object is unchanged.

**How to fix:** Use `.reload` to check actual database state after rollback:

```ruby
expect(evidence.reload.content).to include("/nodes/#{source_node.id}/...")
expect(issue.reload.text).to include("/nodes/#{source_node.id}/...")
```

#### F3 [Low|consistency] app/services/nodes/merger.rb — negative guard with unless / key? pattern

**What:** The update loop uses a negative conditional:

```ruby
unless copied_attachments.key?(attachment.filename)
  new_attachment = attachment.copy_to(target_node)
  self.copied_attachments[attachment.filename] = new_attachment.filename
end
```

**Why:** Positive guards (`next if`) are easier to read than negative conditions. Negative conditions add a mental negation step.

**How to fix:** Rewrite as a positive guard:

```ruby
next if copied_attachments.key?(attachment.filename)
new_attachment = attachment.copy_to(target_node)
self.copied_attachments[attachment.filename] = new_attachment.filename
```

#### F4 [Low|tests] spec/services/nodes/merger_spec.rb — missing test coverage for multi-screenshot evidence

**What:** Evidence with multiple screenshots is supported by the code (`scan` finds all matches), but there are no test examples with more than one screenshot in evidence or issues.

**Why:** Multi-screenshot cases are a common complexity that can hide bugs in loop logic or string replacement.

**How to fix:** Add one spec example:

```ruby
it "updates all screenshots in evidence when multiple are present" do
  evidence = create(:evidence, node: source_node, content: "![](/nodes/1/#{file1}) ![](/nodes/1/#{file2})")
  # ... merge and verify both paths updated
end
```

### Notes

**Codex sub-agent failed:** binary not found. All findings are from Claude. The core logic is correct and transaction handling is sound. The main blocker is F2 — the rollback tests don't actually verify database state. F1 is a readability fix, and F3–F4 are low priority.

---

_Reviewed automatically. Raw outputs in `agents/product-context/reviews/dradis-ce/1291/review/`._

PR #1295, opened 2024-09-16
Issue library QA index 

aapomm:
### Summary

Implement the following changes:

- Datatable state JS cleanup and update bootstrap class
- Remove QA JSON endpoint response

### Check List


app/controllers/qa/issues_controller.rb
etdsoft, line 1: Why are these changes required?

aapomm: Removed as it is unused

app/assets/javascripts/shared/datatable/state.js
```diff
@@ -1,53 +1,61 @@
-$(document).on('preInit.dt', 'body.issues.index', function (e, settings) {
-  var $paths = $('[data-behavior~=datatable-paths]');
-  if ($paths.data('table-state-url') == undefined) {
-    return;
-  }
+$(document).on(
+  'preInit.dt',
+  'body.issues.index',
```
etdsoft, line 3: Is this code in the right location? If it's tied to a page, then it shouldn't live under /shared/datable/ but under tylium/pages/

If this is really shared code across multiple DTs then we need to remove the hard-coding and use a data-behavior or some other trigger to hook into it.

app/assets/javascripts/shared/datatable/state.js
```diff
@@ -1,53 +1,61 @@
...
+  }
+);
 
 DradisDatatable.prototype.setupStateButtons = function () {
   var states = $('[data-behavior~=dradis-datatable]').data('state-icons'),
     stateButtons = [],
     api = this;
 
-  if ($('[data-behavior~=qa-viewer]').length > 0) {
+  if ($('[data-behavior~=qa-viewer]').length) {
```
etdsoft, line 51: same sort of thing. Doesn't belong hard-coded in a shared/ module.

app/controllers/qa/issues_controller.rb
```diff
@@ -46,11 +46,8 @@ def multiple_update
             redirect_to project_issues_path(current_project), notice: 'State updated successfully.'
           end
         end
-
-        format.json { head :ok }
```
etdsoft, line 50: did you do a sweep to see anywhere else where we can remove unnecessary `format.json`s?

app/assets/javascripts/shared/datatable/state.js
aapomm: Apparently the hardcoding is not needed considering that we're already testing for the presence of the `state-url`

app/assets/javascripts/shared/datatable/state.js
aapomm: Moved the state icon handling to the view ✅

app/controllers/qa/issues_controller.rb
aapomm: It doesn't look like there's other unused format.jsons

app/assets/javascripts/shared/datatable/state.js
```diff
@@ -72,7 +74,7 @@ DradisDatatable.prototype.updateRecordState = function (newState) {
       data: {
         ids: DradisDatatable.prototype.rowIds(selectedRows),
         state: newState,
-        return_to: $('[data-behavior~=qa-viewer]').length > 0 ? 'qa' : null
+        return_to: $('[data-behavior~=qa-viewer]').length > 0 ? 'qa' : null,
```
etdsoft, line 77: does this need to go?

aapomm: Replaced with a data attribute instead ✅

PR #1313, opened 2024-11-01
Upgrade Ruby to v3.4.4 and Rails to v8

caitmich:
### Summary

Upgrade Ruby to 3.4.4
Upgrade Rails to 8.0.2.1

### Check List

- [x] Added a CHANGELOG entry


app/models/note.rb
```diff
@@ -41,7 +41,7 @@ class Note < ApplicationRecord
   # the Note or the Issue is pulled when calling Activity.includes(:trackable).
   # (Since we're not using STI, .includes only joins the class of the most recent
   # Activity instead of pulling the correct class (Note/Issue) for each activity)
-  enum state: [:draft, :ready_for_review, :published]
+  enum :state, { draft: 0, ready_for_review: 1, published: 2 }
```
caitmich, line 44: Any reason for switching to a hash instead of `enum :state [:draft, :ready_for_review, :published]`

bin/dev
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env ruby
+exec './bin/rails', 'server', *ARGV
```
caitmich, line 2: or this one?

bin/setup
```diff
@@ -45,7 +44,9 @@ FileUtils.chdir APP_ROOT do
   puts "\n== Removing old logs and tempfiles =="
   system! 'bin/rails log:clear tmp:clear'
 
-  puts "\n== Restarting application server =="
-  # system! 'bin/rails restart'
-  system! 'bin/rails server -p 8080'
+  unless ARGV.include?('--skip-server')
+    puts "\n== Starting development server =="
+    STDOUT.flush # flush the output before exec(2) so that it displays
+    exec 'bin/dev', '-p', '8080'
```
caitmich, line 50: What's the benefit of using `bin/dev` over explicitly using `bin/rails server`?

config/environments/production.rb
```diff
@@ -51,30 +48,41 @@
...
+  config.log_level = ENV.fetch('RAILS_LOG_LEVEL', :info)
+
+  # Use default logging formatter so that PID and timestamp are not suppressed.
+  config.log_formatter = ::Logger::Formatter.new
 
-  # Use a different cache store in production.
+  if ENV['RAILS_LOG_TO_STDOUT'].present?
+    logger           = ActiveSupport::Logger.new(STDOUT)
+    logger.formatter = config.log_formatter
+    config.logger    = ActiveSupport::TaggedLogging.new(logger)
```
caitmich, line 72:
```suggestion
    logger = ActiveSupport::Logger.new(STDOUT)
    logger.formatter = config.log_formatter
    config.logger = ActiveSupport::TaggedLogging.new(logger)
```

Gemfile
```diff
@@ -142,7 +142,9 @@ gem 'net-smtp'
 gem 'net-pop'
 gem 'net-imap', '>= 0.5.7'
 
-gem 'puma', '>= 6.4.3'
+gem 'puma', '>= 6.5.0'
+
+gem 'csv'
```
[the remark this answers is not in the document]
aapomm: `csv` has been removed from the ruby standard lib since 3.4 https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-0-released/ so we need to add it manually here

bin/dev
aapomm: It seems this is used in the context of https://github.com/rails/cssbundling-rails and https://github.com/rails/jsbundling-rails. We don't use either of the two but in case we need to make the switch, this file is here. I don't believe this breaks any workflows so no harm in keeping it.

bin/setup
aapomm: See comment: https://github.com/dradis/dradis-ce/pull/1313#discussion_r2230342200

CHANGELOG
```diff
@@ -1,7 +1,8 @@
 [v#.#.#] ([month] [YYYY])
   - Font: Improve language support
+  - Ruby: Upgrade ruby version to 3.4.4
   - Upgraded gems:
-    - thor
+    - rails, thor
```
etdsoft, line 5: This is a bit inconsistent, particularly considering the major Rails bump. Not sure what we've done before, but I'd expect "upgraded gems" to be mostly minor stuff.

PR #1317, opened 2024-11-14
Allow users to define tag sorting order

caitmich:
### Summary

- Add position column to tags table
- Add `acts_as_list` gem so that we can include it in `Tag`
- Use jQuery sortable to sort tags table

### Check List

- [x] Added a CHANGELOG entry


app/controllers/concerns/sortable.rb
```diff
@@ -0,0 +1,24 @@
+module Sortable
+  extend ActiveSupport::Concern
+
+  def sort
+    sortable_class.transaction do
+      sort_params[:sortable_ids].each_with_index do |id, index|
+        sortable_class.where(id: id.to_i).update_all({ position: index + 1 })
```
aapomm, line 7: We need to scope tags by project, otherwise authors will be able to edit tags even for projects they dont have access to. I think the issue here is we're passing `sortable_class` to this concern but this concern does not have any concept of scoping. Instead, we should pass the collection. The TagsController/RulesController should already have the necessary context to build the `@tags` or `@rules` that we could use here.

app/views/tags/_table.html.erb
```diff
@@ -1,25 +1,32 @@
 <% cache ['tags-table', @all_columns, @tags] do %>
```
aapomm, line 1: Should we bust this cache key when the order is changed?

app/controllers/tags_controller.rb
```diff
@@ -43,14 +44,18 @@ def destroy
...
+
   def tag_params
     modified_params = params.require(:tag).permit(:name, :color)
     modified_params[:name] = "#{modified_params[:color].gsub('#', '!')}_#{modified_params[:name]}"
     modified_params.except(:color)
   end
 
   def set_columns
-    default_field_names = ['Name'].freeze
+    default_field_names = ['Sort', 'Name'].freeze
```
aapomm, line 58: Are we using the `Sort` column? I'm not seeing it in the options

app/controllers/concerns/sortable.rb
caitmich: Updated to pull `sortable_records` from each controller like we're doing to get the class, so now we're limiting the list accordingly in each controller

app/views/tags/_table.html.erb
caitmich: Updated the `Sortable` concern so it's now updating the records' `updated_at` values when we change the order. Previously it wasn't being updated because we're using `update_all`

app/controllers/tags_controller.rb
caitmich: The `th` is hidden in the view but we need this column to show the sort handle

app/controllers/concerns/sortable.rb
```diff
@@ -0,0 +1,29 @@
...
+    head :ok
+  end
+
+  private
+
+  def sort_params
+    params.permit(sortable_ids: [])
+  end
+
+  def sortable_class
```
aapomm, line 20: If `transaction` is the only purpose of this method then we don't need it since we can just call `ActiveRecord::Base.transaction` regardless of the model type

app/views/issues/_table.html.erb
```diff
@@ -31,7 +31,7 @@
                   [issue.title, (link_to(issue.title, [current_project, issue], data: { qa_visible: false } ) + link_to(issue.title, project_qa_issue_path(current_project, issue), class: 'd-none', data: { qa_visible: true }))]
                 when 'Tags'
                   [
-                    issue.tags.any? ? issue.tags.first.display_name : '',
+                    issue.tags.any? ? issue.tags.first.position : false,
```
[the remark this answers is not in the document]
MattBudz: We are using ints to sort tags: 0 = highest priority, next comes 1, then 2, and so on. 

Due to how JS sorts, having a `''` or 'null' value for `sort` will keep untagged issues above the highest priority issue (ie "Critical"), since `''`/`null` will come before 0. Strings/bools sort after the ints so having a bool or a string, allows the untagged issues to be sorted after the issues tagged with the lowest priority tag. I went with `false` as it seemed the most reasonable over an arbitraray string.

app/views/tags/_table.html.erb
```diff
@@ -1,25 +1,32 @@
...
+    data-default-columns="<%= @default_columns.to_json %>"
+    data-item-name="tags"
+    data-local-storage-key="project.ce.<%= dom_id(current_project) %>.tags_datatable">
     <thead>
       <tr>
         <% @all_columns.each do |column| %>
-          <th><%= column %></th>
+          <th class="no-sort" data-column-visible="<%= column != 'Sort' %>">
+            <span class="<%= class_names('visually-hidden': column == 'Sort'  )%>"><%= column %></span>
+          </th>
```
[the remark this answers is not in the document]
MattBudz: Removed all the sort-related stuff. Class `no-sort` remains otherwise the table will be sortable.

app/views/tags/_table.html.erb
```diff
@@ -1,25 +1,32 @@
...
       <% @tags.each do |tag| %>
         <% cache [tag, @all_columns, 'tags-table'] do %>
           <tr id="tag-<%= tag.id %>">
             <% @all_columns.each do |column| %>
               <%
               sort, display =
               case column
+              when 'Sort'
+                [
+                  'Change sort order',
```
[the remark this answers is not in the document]
MattBudz: per column sorting was already disabled. This should've also been removed, it's removed now.

app/controllers/concerns/sortable.rb
```diff
@@ -0,0 +1,24 @@
+module Sortable
+  extend ActiveSupport::Concern
+
+  def sort
+    ActiveRecord::Base.transaction do
+      sort_params[:sortable_ids].each_with_index do |id, index|
+        sortable_records.where(id: id.to_i).update_all({ position: index + 1, updated_at: Time.now })
```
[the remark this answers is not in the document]
caitmich: The `to_i` is needed because it's coming as a list of strings from the front end

We need to check agains the list of permitted records from the controller first, so I updated this to check the list of permitted ids and update if the id is included
```ruby
if sortable_records[:ids].include?(id)
  klass.update(id, position: index)
end
```

db/migrate/20241114154140_add_position_to_tags.rb
```diff
@@ -0,0 +1,11 @@
+class AddPositionToTags < ActiveRecord::Migration[7.0]
+  def change
+    add_column :tags, :position, :integer
+
+    Tag.transaction do
+      Tag.all.each.with_index(1) do |tag, index|
+        tag.update_attribute(:position, index)
```
etdsoft, line 7: This is going to break Pro big time.

PR #1321, opened 2024-11-26
Hera/add node tree

MattBudz:
### Summary

This PR adds the Node tree as a sidebar to Hera

### Check List


app/assets/stylesheets/hera/modules.scss
```diff
@@ -110,10 +111,6 @@
   width: 0;
 }
 
-.fa-fw {
-  margin-right: 0.25rem;
```
MattBudz, line 114: This should be a base branch change since it's affecting a global class, not something nested under the node tree sidebar specifically.

This type of declaration doesn't belong under `fa-fw` since we won't always want margin-right on every icon. It caused issues here so I removed this.

app/views/layouts/hera/_sidebar_header.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="sidebar-header">
+  <%= link_to main_app.root_path, class: 'd-flex' do %>
```
caitmich, line 2: since this will only be accessed in-project do we need `main_app`

nicolachr: We're not using this file anymore. I'm not sure why/how it ended up in this branch when it's not in the base branch. Checking with Matt.

MattBudz: Looks like a bad merge, it's removed now.

PR #1322, opened 2024-11-28
Hera/add secondary sidebar

nicolachr:
### Summary

This PR adds a secondary sidebar to Hera's main layout

### Check List


app/assets/stylesheets/hera/variables.scss
```diff
@@ -26,6 +26,7 @@ $text-error: $red;
 $text-link: $brand-700;
 $text-link-hover: darken($text-link, 15%);
 $text-muted: $grey-700;
+$text-small: 0.9rem;
```
MattBudz, line 29: can't we just use bootstrap's `small` class as needed?

nicolachr: We definitely could. Sometimes I find it easier to bundle it together instead of going through each item to add the class. I'm happy to remove it if you find it unnecessary

app/views/shared/_add_item_dropdown.html.erb
```diff
@@ -1,15 +1,10 @@
-<div class="dropdown">
+<div class="dropdown me-2">
   <% if local_assigns[:new_path] %>
-    <%= link_to 'javascript:void(0)',
-      class: local_assigns.fetch(:button_class, '') + ' dropdown-toggle',
+   <%= link_to 'javascript:void(0)',
+      class: local_assigns.fetch(:button_class, '') + ' dropdown-toggle no-caret',
       data: { name: name.downcase, bs_toggle: 'dropdown' } do %>
-      <i class="fa-solid fa-plus"> </i> <%= local_assigns.fetch(:text, '') %><span class="visually-hidden">Add <%= name.downcase %></span>
+      <i class="fa-solid fa-plus"> </i> <%= local_assigns.fetch(:text, '') %><span class="fa-sr-only">New <%= name.downcase.singularize %></span>
```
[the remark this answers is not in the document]
nicolachr: No particular reason. I was just following font-awesome's accessibility guidelines.

app/views/shared/_add_item_dropdown.html.erb
```diff
@@ -1,15 +1,10 @@
-<div class="dropdown">
+<div class="dropdown me-2">
   <% if local_assigns[:new_path] %>
     <%= link_to 'javascript:void(0)',
-      class: local_assigns.fetch(:button_class, '') + ' dropdown-toggle',
+      class: local_assigns.fetch(:button_class, '') + ' dropdown-toggle no-caret',
       data: { name: name.downcase, bs_toggle: 'dropdown' } do %>
-      <i class="fa-solid fa-plus"> </i> <%= local_assigns.fetch(:text, '') %><span class="visually-hidden">Add <%= name.downcase %></span>
+      <i class="fa-solid fa-plus"> </i> <%= local_assigns.fetch(:text, '') %><span class="visually-hidden">New <%= name.downcase.singularize %></span>
```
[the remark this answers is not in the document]
nicolachr: Yes. The updated actions will be "+ New" to avoid using "+ Add" which is technically duplicating the same action

app/assets/stylesheets/hera/modules/_attachments.scss
```diff
@@ -38,26 +37,22 @@
...
     }
 
     .error {
       width: 120px;
     }
 
     .filename {
-      white-space: nowrap;
+      -webkit-line-clamp: 2;
+      -webkit-box-orient: vertical;
```
MattBudz, line 50: note, we are using -webkit here but this is universally adapted according to can i use. Even FF uses the -web-kit prefix for this.

nicolachr: I'm sorry, it's not clear to me what you mean. It doesn't require the prefix?

MattBudz: This is just an FYI for reviews wondering why we are using -webkit prefixes in production (typically features behind experimental flags are a no-no in production). In this case, the feature was released to all browsers with the -webkit prefix.

PR #1323, opened 2024-11-29
Fix incorrect parsing of field values that contain Liquid when creating new records

caitmich:
### Summary

When creating new issue/evidence from a note template or from the default template (RTP), if you have Liquid content that contains filters Ex. `{{ issue.fields['CVSSv3.BaseScore'] | times: 10 }}`, it will be split into different options in the dropdown
```
["{{ issue.fields['CVSSv3.BaseScore']", "times: 10 }}"]
```

#### Solution
Use regex to scan the string and split it into the correct options

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,7 +7,6 @@
     - rexml
   - Bugs fixes:
     - Editor: Fix incorrect parsing of field value options that contain Liquid
-      - [future tense verb] [bug fix]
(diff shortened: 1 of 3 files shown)


caitmich:
Addressed in https://github.com/dradis/dradis-ce/pull/1376

PR #1327, opened 2024-12-23
API add Node properties

cbp1983:
### Summary

This change allows users to view Node Properties as part of API Node requests.

### Copyright assignment

### Check List

- [X] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -22,8 +22,8 @@
     - [report type]:
       - [future tense verb] [reporting enhancement]
   - REST/JSON API enhancements:
-    - [API entity]:
-      - [future tense verb] [API enhancement]
+    - Nodes endpoint:
+      - include Node Properties in Node API requests
   - Security Fixes:
     - High: (Authenticated|Unauthenticated) (admin|author|contributor) [vulnerability description]
     - Medium: (Authenticated|Unauthenticated) (admin|author|contributor) [vulnerability description]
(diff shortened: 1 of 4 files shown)


CHANGELOG
```diff
@@ -22,8 +22,8 @@
     - [report type]:
       - [future tense verb] [reporting enhancement]
   - REST/JSON API enhancements:
-    - [API entity]:
-      - [future tense verb] [API enhancement]
+    - Nodes endpoint:
```
etdsoft, line 25:
```suggestion
    - Nodes:
```

CHANGELOG
```diff
@@ -22,8 +22,8 @@
...
+      - include Node Properties in Node API requests
```
etdsoft, line 26:
```suggestion
      - include Node properties
```

engines/dradis-api/app/controllers/dradis/ce/api/v3/nodes_controller.rb
```diff
@@ -44,7 +44,7 @@ def destroy
       protected
 
       def node_params
-        params.require(:node).permit(:label, :type_id, :parent_id, :position)
+        params.require(:node).permit(:label, :type_id, :parent_id, :position, properties: {})
```
etdsoft, line 47: This is important, as using JSON opens a can of worms around validation. If you look at the app's NodesController [we stick to raw_properties](https://github.com/dradis/dradis-ce/blob/develop/app/controllers/nodes_controller.rb#L134) to let the user submit a "text blob" with the JSON properties, but we [then apply strict validation](https://github.com/dradis/dradis-ce/blob/develop/app/models/concerns/node_properties.rb#L14) at the model level.

I'd expect for the API flow to follow a very similar pattern.

PR #1331, opened 2025-01-22
Add View History link in QA

aapomm:
### Spec

If a user doing QA on an issue/content block wants to see the history, they will have to navigate out of the QA page, find the relevant record and click on it. Only then will they have access to the revisions history of the record. This takes a lot of clicks when we can just add a link to the revision history in the QA#show.

**Proposed solution**

- Add View history button to the QA Issues and Content blocks show pages
- Dynamically render the revisions breadcrumbs depending on where the user is coming from
  - If the user is from the QA page, render breadcrumbs that link to QA pages
  - Otherwise, render the breadcrumbs that link to the non-QA page

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/helpers/revisions_helper.rb
+++ b/app/helpers/revisions_helper.rb
@@ -28,7 +28,7 @@ module RevisionsHelper
     when Card
       project_board_list_card_revision_path(current_project, @board, @list, record, revision)
     when Issue
-      project_issue_revision_path(current_project, record, revision)
+      project_issue_revision_path(current_project, record, revision, params)
     when Note
       project_node_note_revision_path(current_project, record.node, record, revision)
     when Evidence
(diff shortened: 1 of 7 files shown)


app/helpers/revisions_helper.rb
```diff
@@ -28,7 +28,7 @@ def record_revision_path(record, revision, params = {})
     when Card
       project_board_list_card_revision_path(current_project, @board, @list, record, revision)
     when Issue
-      project_issue_revision_path(current_project, record, revision)
+      project_issue_revision_path(current_project, record, revision, params)
```
etdsoft, line 31: if we only care about :return_to, why are we passing more?

app/views/revisions/show.html.erb
```diff
@@ -43,7 +43,7 @@
                     <% if revision == @revision %>
                       Currently Viewing
                     <% elsif revision.event == "update" %>
-                      <%= link_to "View Changes", record_revision_path(@record, revision) %>
+                      <%= link_to "View Changes", record_revision_path(@record, revision, return_to: @return_to_qa ? :qa : nil) %>
```
etdsoft, line 46: this is another instance of a path helper, but we're solving it in a different way.

PR #1332, opened 2025-01-23
Include methodology board in card and list activities

rachkor:
### Summary
Update the activity to include the methodology for all List activities:
Use "methodology" instead of "board" for consistent terminology.

### Check List
- [x] Added a CHANGELOG entry


CHANGELOG
```diff
@@ -1,5 +1,6 @@
 [v#.#.#] ([month] [YYYY])
   - Hera: Add new layout with redesigned navigation
+  - Activities: Include methodology name in all methodology actions and clean up syntax
```
caitmich, line 3:
```suggestion
  - Activities: Include methodology name in all methodology actions
```

caitmich: we don't need the last part because it's dev speak

PR #1333, opened 2025-01-27
QA Reviewers

aapomm:
### Spec
Teams have had issues with authors erroneously setting issues/content blocks as "published" and would prefer that Dradis put guards in place that would prevent anyone (with access to a project) to publish any issue/content block in that project.

**Proposed solution**
This task implements the following:

- Add a "reviewer" virtual role
  - This is represented by a "qa-review" permission to a project
- Add a way to assign authors as reviewers
- Hide the "published" state option for non-reviewers. This is hidden in the following views:
  - issues/content blocks #index
  - qa issues/content blocks #index
  - qa issues/content blocks #show
  - issues/content blocks form (submit_button)

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/controllers/concerns/publishable.rb
+++ b/app/controllers/concerns/publishable.rb
@@ -8,17 +8,28 @@ module Publishable
   private
 
   def ensure_reviewer
-    params_method = "#{controller_name.singularize}_params"
-    return if can?(:publish, current_project) || !self.respond_to?(params_method)
+    return if can?(:publish, current_project)
 
-    record_params = self.send(params_method)
-    current_state =
-      self.get_instance_variable("@#{controller_name.singularize}")&.state
+    record_params = self.send("#{controller_name.singularize}_params")
+    current_state = record&.state
 
-    if (record_params[:state] == 'published' && current_state != 'published') ||
+    if (current_state != 'published' && record_params[:state] == 'published') ||
         params[:state] == 'published'
 
       redirect_to project_issues_path(current_project), alert: 'Unable to publish record!'
+    elsif current_state == 'published' && record_params[:state] == ''
+      # If the user is not a reviewer, and the record is currently set as
+      # 'Published', the radio button will be disabled and record_params[:state]
+      # will be '' (empty string). Since an empty state is invalid, we delete
+      # the param here instead.
+      params[controller_name.singularize].delete(:state)
     end
+  end
+
+  def record
+    return unless params[:id]
+
+    klass = controller_name.classify.constantize
+    klass.find(params[:id])
   end
 end
--- a/app/controllers/concerns/publishable.rb
+++ b/app/controllers/concerns/publishable.rb
@@ -9,7 +9,7 @@ module Publishable
 
   def ensure_reviewer
     params_method = "#{controller_name.singularize}_params"
-    return if can?(:publish, current_project) || !self.respond_to?(params_method) 
+    return if can?(:publish, current_project) || !self.respond_to?(params_method)
 
     record_params = self.send(params_method)
     current_state =
--- a/app/controllers/concerns/publishable.rb
+++ b/app/controllers/concerns/publishable.rb
@@ -1,9 +1,7 @@
-module QAStates
+module Publishable
   extend ActiveSupport::Concern
 
   included do
-    before_action :set_states, only: [:new, :create, :edit, :update]
-
     before_action :ensure_reviewer, only: [:create, :update, :multiple_update]
   end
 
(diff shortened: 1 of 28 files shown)


app/controllers/concerns/qa_states.rb
```diff
@@ -0,0 +1,28 @@
+module QAStates
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :set_states, only: [:new, :create, :edit, :update]
+
```
caitmich, line 6:
```suggestion
```

app/controllers/concerns/qa_states.rb
```diff
@@ -0,0 +1,28 @@
+module QAStates
```
caitmich, line 1: Any reason for `QAStates` over `QA::States`? Or what about a more rails concern-y name like `Publishable` or

app/controllers/concerns/qa_states.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    if (record_params[:state] == 'published' || params[:state] == 'published') &&
+        !can?(:publish, @issue)
+
+      redirect_to project_issues_path(current_project), alert: 'Unable to publish record!'
+    end
+  end
+
+  def set_states
+    @states = Issue.states.dup
```
caitmich, line 26: What's the purpose of this change and passing the states to the partial now?

caitmich: This is used in CB controllers in Pro so this shouldn't be issue-specific

app/views/qa/_state_button.html.erb
```diff
@@ -7,21 +7,27 @@
     <span class="visually-hidden">Toggle Dropdown</span>
   </button>
   <div class="dropdown-menu">
-    <%= f.collection_radio_buttons(:state, record.class.states, :first, :first) do |b| %>
-      <%= b.label class: 'state' do %>
-        <%= b.radio_button class: 'd-none', data: { behavior: 'state-radio'} %>
+    <%= f.collection_radio_buttons(:state, states, :first, :first) do |b| %>
+      <% label_options = {} %>
```
caitmich, line 11: Can we clean this up at all? Maybe in a helper? 
We can pass in the state to a method that returns an empty hash or
```ruby
{ data: { bs_toggle: 'tooltip', bs_title: 'You are not a Reviewer for this project.' } }
```
and call it inline on L15

Or in general can we structure this so that we have an object that provides us with what we'll need ex.
label_options, disabled: true/false, state-label and we just pull each section in the view?
```ruby
{ published: { disabled: true, label_options: {}, state_label: 'Content is final, ready for the report.' }
```

app/controllers/concerns/publishable.rb
```diff
@@ -0,0 +1,35 @@
+module Publishable
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :ensure_reviewer, only: [:create, :update, :multiple_update]
+  end
+
+  private
+
+  def ensure_reviewer
```
caitmich, line 10: I think this method is doing too much. `ensure_reviewer` makes sense for up to L19 but I think the rest should be its own method

app/controllers/concerns/publishable.rb
```diff
@@ -0,0 +1,35 @@
...
+    return if can?(:publish, current_project)
+
+    record_params = self.send("#{controller_name.singularize}_params")
+    current_state = record&.state
```
caitmich, line 14: This won't currently work for `multiple_update`

caitmich: On that note, we should include `ids: []` in permitted params for `multiple_update`

app/controllers/concerns/publishable.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    if (current_state != 'published' && record_params[:state] == 'published') ||
```
caitmich, line 16: What case would we be using params instead of strong params?

app/views/qa/_state_button.html.erb
```diff
@@ -8,20 +8,15 @@
   </button>
   <div class="dropdown-menu">
     <%= f.collection_radio_buttons(:state, record.class.states, :first, :first) do |b| %>
-      <%= b.label class: 'state' do %>
-        <%= b.radio_button class: 'd-none', data: { behavior: 'state-radio'} %>
+      <%= b.label class: 'state', **label_options_for(b.value) do %>
+        <%= b.radio_button class: 'd-none', 
+          data: { behavior: 'state-radio' }, 
+          disabled: (b.value == 'published' && !can?(:publish, current_project)) %>
```
caitmich, line 14: Since we're doing the rest in the helper, we could do this there as well

app/controllers/concerns/publishable.rb
aapomm: strong params for `issues#update` while params for `qa/issues#multiple_update`

app/controllers/concerns/publishable.rb
aapomm: Refactored the method ✅ 

> On that note, we should include ids: [] in permitted params for multiple_update

I'm not sure if this needs strong params considering that we're just using the `ids` for #find while the state param is also validated

PR #1334, opened 2025-02-03
Redirect to the next item when clicking on the QA widget

aapomm:
### Spec
Going through the list of items that are ready for review requires a lot of clicks since clicking on the widget redirects the users to the QA#index page. We can make it easier for users by redirecting the users to the next ready for review item instead.

**Proposed solution**
Update the redirect_to path on the QA#update controller

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/controllers/qa/issues_controller.rb
+++ b/app/controllers/qa/issues_controller.rb
@@ -64,10 +64,11 @@ class QA::IssuesController < AuthenticatedController
   end
 
   def redirect_to_next_issue_or_index
-    notice = 'State updated successfully.'
+    notice = "State successfully updated for #{@issue.title}."
     next_issue = current_project.issues.ready_for_review.first
 
     if next_issue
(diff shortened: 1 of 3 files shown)


app/controllers/qa/issues_controller.rb
```diff
@@ -23,7 +23,8 @@ def edit
   def update
     if @issue.update(state: @state, updated_at: Time.now)
       track_state_change(@issue)
-      redirect_to project_qa_issues_path(current_project), notice: 'State updated successfully.'
+
+      redirect_to_next_issue_or_index
```
etdsoft, line 27: from an esthetics point of view, I would have preferred something like `redirect_to next_issue_or_index_path` and have the helper return a path. More familiar.

I see now it may be because of the :notice param :( although potentially the method could return **[path, notice: ""] which could be fed to `redirect_to`... not sure if too fancy.

PR #1336, opened 2025-02-05
Add working HTML report template to default config

rachkor:
### Summary
The Welcome Kit HTML report template (`dradis_template-welcome.v0.4.html.erb`) does not work with CE as it references content blocks and throws an error on export. This PR replaces the welcome kit report template with a CE-specific HTML report template. It also removes the `.rb` file since RTPs are not used by CE. 

### Check List
- [X] Added a CHANGELOG entry


CHANGELOG
```diff
@@ -16,8 +16,8 @@
       - [integration bug fixes]:
         - [future tense verb] [integration bug fix]
   - Reporting enhancements:
-    - [report type]:
-      - [future tense verb] [reporting enhancement]
+    - HTML:
+      - Fix default HTML report template to work with CE
```
caitmich, line 20: This should go under integration bug fixes on L16

PR #1337, opened 2025-02-06
Hera/Improve parity 

nicolachr:
### Summary

This PR brings additional CSS to improve parity between Dradis CE and Dradis Pro.

### Check List

- [ ] Added a CHANGELOG entry


app/helpers/hera_helper.rb
```diff
@@ -1,4 +1,28 @@
 module HeraHelper
+  def colored_icon_for_model(model, icon_class, extra_class = nil)
+    css =  ['fa-solid']
+    css << icon_class
+    css << extra_class if extra_class
+
+    options = { class: css.join(' ') }
+    tag     = nil
```
caitmich, line 8:
```suggestion
    tag = nil
```

app/helpers/hera_helper.rb
```diff
@@ -26,34 +50,18 @@ def flash_messages
     end.join("\n").html_safe
   end
 
+  def in_project?
+    defined?(current_project) && current_project&.persisted?
```
caitmich, line 54:
```suggestion
    current_project && current_project.persisted?
```

caitmich: we don't need the safeguard since we shouldn't end up on the right side of this expression unless `current_project` exists

app/views/layouts/hera.html.erb
```diff
@@ -28,98 +29,105 @@
...
+  <%= 
+    content_tag(
+      :body, 
+      class: class_names(controller_path.gsub('/', '-'), 
+        controller_name, 
+        action_name, 
+        'in-project': in_project?
+      ),
+      data: {
+        env: (Rails.env unless Rails.env.production?),
```
caitmich, line 41: What's the purpose of this line?

app/helpers/hera_helper.rb
caitmich: Unless this is due to a strange case where it's defined and not set? There is probably a nicer way to handle this case

PR #1338, opened 2025-02-10
Activities/add filters

nmdc00:
### Summary

Code changes in this merge request create a search filter that enables the user to filter activities based on user_id, trackable_type and start_date/end_date.

Through this the user can see what activities were made using these params. Works in conjunction with the infite scroll already present. Added said params to the url and the ajax request. Changed activities query logic so that it iterates through every activity and not just the one on the first page as defined by kaminari. Implemented a can_load_more logic to make sure infinity scroll knows when it's reached the end of activities and created a reset button which reloads the page and the url so params disappear and all activities load as they were previous to the filters.

Also added a DateTimeHelper for the date range filter logic between the activity groups date and the input data on the search filter.

I've also created some tests for good measure to insure the logic works well 
### Other Information

### Copyright assignment

file for the details.

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/assets/javascripts/hera/pages/activities.js
+++ b/app/assets/javascripts/hera/pages/activities.js
@@ -22,7 +22,8 @@ document.addEventListener('turbolinks:load', function () {
         
         var selectedStartDate = $('#start_date').val();
         var selectedEndDate = $('#end_date').val();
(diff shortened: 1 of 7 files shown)


MattBudz:
thanks @nmdc00 closing this for now

PR #1386, opened 2025-07-07
Use string uuid for logs

aapomm:
### Summary

Refactor the `Log` model to use string UUIDs

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/controllers/concerns/uploadable.rb
+++ b/app/controllers/concerns/uploadable.rb
@@ -14,7 +14,7 @@ module Uploadable
 
     job_logger.write 'Enqueueing job to start in the background.'
 
-    uid = is_api? ? job_logger.id : upload_params[:item_id]
+    uid = is_api? ? job_logger.uid : upload_params[:item_id]
 
     # NOTE: call the bg job as last thing in the action helps us
     # avoid SQLite3::BusyException when using sqlite and
--- a/app/controllers/concerns/uploadable.rb
+++ b/app/controllers/concerns/uploadable.rb
@@ -14,7 +14,7 @@ module Uploadable
 
     job_logger.write 'Enqueueing job to start in the background.'
 
-    uid = is_api? ? job_logger.id : upload_params[:item_id].to_i
+    uid = is_api? ? job_logger.id : upload_params[:item_id]
 
     # NOTE: call the bg job as last thing in the action helps us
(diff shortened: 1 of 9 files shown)


app/views/upload/create.js.erb
```diff
@@ -1,5 +1,5 @@
 <% if @success -%>
-  ConsoleUpdater.jobId = <%= @item_id %>
+  ConsoleUpdater.jobId = '<%= @item_id %>'
```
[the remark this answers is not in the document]
aapomm: It's the erb -> js conversion. Once `<%= @item_id %>` is resolved, the line becomes:
```
ConsoleUpdater.jobId = any_string_here
```

JS sees `any_string_here` as a variable and tries to resolve it as such, which causes an error. Ideally the resulting JS should look like this instead:

```
ConsoleUpdater.jobId = 'any_string_here'
```

which explains the wrapped quotes

db/migrate/20250707070805_change_log_uid_to_string.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeLogUidToString < ActiveRecord::Migration[7.2]
+  def change
+    change_column :logs, :uid, :string
```
etdsoft, line 3: Is this a straightforward change when there are rows with numbers?

We could preemptively truncate the table before applying the change, no?

app/controllers/concerns/uploadable.rb
```diff
@@ -14,7 +14,7 @@ def process_upload_background(args = {})
 
     job_logger.write 'Enqueueing job to start in the background.'
 
-    uid = is_api? ? job_logger.id : upload_params[:item_id].to_i
+    uid = is_api? ? job_logger.id : upload_params[:item_id]
```
etdsoft, line 17: the whole reason of being for this parameter is to track the job. Why is .id used? Shouldn't it be .uid?

db/migrate/20250707070805_change_log_uid_to_string.rb
aapomm: The integer uids will just be converted to the strings. So if a Log record exists with uid=1, after this migration it will just be uid="1". I'm relying on the log cleanup to eventually delete the outdated logs

PR #1404, opened 2025-08-06
Add docker compose and update Dockerfile

caitmich:
### Summary

- Add `compose.yaml` for docker compose
- Update Dockerfile for ruby 3.4.4 and remove unneeded steps

### Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -221,7 +221,7 @@ end
 #
 
 # Base framework classes required by other plugins
-gem 'dradis-plugins', github: 'dradis/dradis-plugins', branch: 'rails/upgrade-3.4'
+gem 'dradis-plugins', '~> 4.17.0'
 
 gem 'dradis-api', path: 'engines/dradis-api'
 
--- a/Gemfile
+++ b/Gemfile
@@ -111,7 +111,7 @@ gem 'sanitize', '6.0.2'
 
 # SQLite3 DB driver
 gem 'sqlite3'
-gem 'pg'
+# gem 'pg'
 
 # --------------------------------------------------------- Dradis Professional
 # Authorisation
@@ -221,7 +221,7 @@ end
 #
 
 # Base framework classes required by other plugins
-gem 'dradis-plugins', '~> 4.17.0'
+gem 'dradis-plugins', github: 'dradis/dradis-plugins', branch: 'rails/upgrade-3.4'
 
 gem 'dradis-api', path: 'engines/dradis-api'
 
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,0 +1,14 @@
+name: "dradis-ce-production"
+
+services:
+  web:
+    build:
+      context: .
+      platforms:
+      - linux/amd64
+    volumes:
+    - dradis-ce-db:/app/db
+    ports:
+    - 3000:3000
+volumes:
+  dradis-ce-db:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
+ARG RUBY_VERSION=3.4.4
 # We're sticking to non-slim version: https://hub.docker.com/_/ruby/
-FROM --platform=linux/amd64 ruby:3.4.4
+FROM ruby:${RUBY_VERSION}
 
 WORKDIR /app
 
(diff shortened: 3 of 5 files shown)


compose.yaml
```diff
@@ -0,0 +1,11 @@
+name: "dradis-ce-production"
+
+services:
+  web:
+    image: dradis/dradis-ce:latest
+    volumes:
+    - dradis-ce-db:/app/db
```
aapomm, line 7:
```suggestion
      - dradis-ce-db:/app/db
```

Gemfile
```diff
@@ -111,7 +111,7 @@ gem 'sanitize', '6.0.2'
 
 # SQLite3 DB driver
 gem 'sqlite3'
-gem 'pg'
+# gem 'pg'
```
aapomm, line 114: Are we supporting CE with postgres as the DB? If not, then we should just remove this line

caitmich: AFAIK this is for Heroku only. If we want to continue to support Heroku we can make some tweaks so it'll work with our CE Docker image, but that can be done in a different task. For that reason I'll leave this commented for now.

compose.yaml
```diff
@@ -0,0 +1,35 @@
...
+services:
+  web:
+    # image: dradis/dradis-ce:latest
+    build:
+      context: .
+      platforms:
+        - linux/amd64
+      args:
+        # Build arguments:
+        SSL_CERT_DIR: ${SSL_CERT_DIR}
```
caitmich, line 12: This will be removed before merge since we want to pull the image from Docker hub instead

caitmich: But leaving it now for testing & since we'll want to use this for documenting how to build the image

compose.yaml
```diff
@@ -0,0 +1,35 @@
...
+    environment:
+      # Runtime environment variables:
+      - RAILS_ENV=${RAILS_ENV}
+      - RAILS_SERVE_STATIC_FILES=${RAILS_SERVE_STATIC_FILES}
```
aapomm, line 16: Where is this used and why is this configurable?

compose.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - RAILS_X_SENDFILE_HEADER=${RAILS_X_SENDFILE_HEADER} # set to nil for use without reverse proxy
```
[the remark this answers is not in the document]
caitmich: That's because you didn't have the ENV set. Commented out the relevant line in `production.rb` since we're not using Nginx

config/puma.rb
```diff
@@ -32,12 +32,23 @@
...
-# Specifies the `port` that Puma will listen on to receive requests; default is 3000.
-port ENV.fetch('PORT', 3000)
-
 # Specifies the `environment` that Puma will run in.
 environment ENV.fetch('RAILS_ENV') { 'development' }
 
+key_file = ENV.fetch('SSL_KEY_FILE', '/etc/ssl/dradis.local/dradis.local.key')
+cert_file = ENV.fetch('SSL_CERT_FILE', '/etc/ssl/dradis.local/bundle.dradis.local.crt')
+
+if ENV['RAILS_ENV'] == 'production' && File.exist?(key_file) && File.exist?(cert_file)
```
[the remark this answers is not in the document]
caitmich: Only if we're running a Docker image in production. If, for example, this is being run in DO App Platform it would be production but these files wouldn't exist since it'd be using Nginx as a reverse proxy, which is why I'm defaulting to http if they aren't present

compose.yaml
```diff
@@ -0,0 +1,27 @@
...
+        - linux/amd64
+    volumes:
+      # Persist the database
+      - dradis-ce-storage:/app/storage
+    ports:
+      - 443:443
+      - 80:80
+    environment:
+      # Set domain from ENV var so thruster can handle let's encrypt certificates
+      - TLS_DOMAIN=${TLS_DOMAIN}
```
etdsoft, line 18: is this syntax really needed? wouldn't Docker do this by default?

etdsoft, line 18: and related, isn't here where RAILS_MASTER_KEY would go?

caitmich: You're right it will automatically load all env vars in `.env` without the need for explicitly defining them here. Removed

PR #1406, opened 2025-08-29
Update .dockerignore to omit development/test files and `vendor/bundle`

caitmich:
### Summary

Update .dockerignore to omit:
- `vendor/bundle` - so that we're not duplicating the bundled gems (they are located in `/usr/local/bundle`)
- `spec/`
- other files/directories needed for development or test

### Check List


Dockerfile
```diff
@@ -6,7 +6,7 @@ WORKDIR /dradis
 
 # Install dependencies for Ruby build and gems
 RUN apt-get update -qq && \
-apt-get install --no-install-recommends -y curl git libjemalloc2 libvips sqlite3 && \
+apt-get install --no-install-recommends -y curl git libjemalloc2 libvips sqlite3 redis-server && \
```
etdsoft, line 9: I know it's silly, but I'd split the regis-server into another apt-line.

It's likely we'll end up splitting this in a separate container, in the future or in Pro, and that would make the difference clearer.

PR #1411, opened 2025-09-15
Allow issue creation from card

gcelanji:
### Summary

Allow users creating issues while navigating though methodologies cards. Card fields & values are used as a template for the issue.

### Check List

- [ ] Added a CHANGELOG entry


app/controllers/issues_controller.rb
```diff
@@ -219,4 +220,13 @@ def set_auto_save_key
...
     end
   end
+
+  def set_card_content
+    return unless params[:board_id] && params[:card_id]
+
+    board = current_project.boards.includes(:cards).find(params[:board_id])
+    card = board.cards.find(params[:card_id])
+
+    @issue.text = "\n#[Title]#\n#{card.name}\n#{card.description}"
```
aapomm, line 230: - What is the newline at the beginning?
- The description needs to be under a `Description` field
- This needs another newline after the title

spec/features/issues_spec.rb
```diff
@@ -159,6 +159,30 @@
           end
         end
 
+        context 'when passed a card', js: true do
+          let (:board)  { create(:board, project: current_project, node: current_project.methodology_library) }
+          let (:list)  { create(:list, board: board) }
```
aapomm, line 164:
```suggestion
          let (:list) { create(:list, board: board) }
```

spec/features/issues_spec.rb
```diff
@@ -159,6 +159,30 @@
           end
         end
 
+        context 'when passed a card', js: true do
```
aapomm, line 162: Can we move the `js: true` call to only the editor example?

app/controllers/issues_controller.rb
gcelanji: The first and third bullets have been addressed. For the second one, by adding a description field for the card description, we get an empty description field in the new issue, which seems like is not required. In the description field of a card record, it has saved its fields and values in this format: `"#[One]#\r\n\r\n\r\n#[Two]#\r\n`

Screenshot 2025-09-26 100032

PR #1414, opened 2025-09-16
Improve Editor UI

nicolachr:
### Summary

The current state of the Editor doesn't fully utilize the available space, making editing difficult. This PR, makes the necessary changes to the UI in order to improve the editing experience, and give the Editor a more modern look.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/assets/stylesheets/hera/modules/_textile.scss
+++ b/app/assets/stylesheets/hera/modules/_textile.scss
@@ -196,6 +196,7 @@
         margin: 0 1.5rem 0.5rem 0;
         padding-left: 0.5rem;
         position: relative;
+        transition: border 0.2s ease-in-out;
 
         .actions {
           top: 0;


app/assets/stylesheets/hera/modules/_textile.scss
```diff
@@ -172,68 +172,62 @@
     background: transparent;
     padding: 0;
 
+    .editor-field {
+      .textile {
+        border: 0;
+        border-right: 1px solid $border-color;
+        border-radius: 0;
+        box-shadow: none !important;
```
MattBudz, line 180: do we need `!important`? If so, is there a better way?

app/assets/stylesheets/hera/modules/_textile.scss
```diff
@@ -172,68 +172,62 @@
...
 
-        &:first-of-type:hover {
-          border-top-color: transparent;
-        }
-
         &:hover {
-          background-color: darken($secondary-bg, 2%);
-          border-bottom-color: $border-color;
-          border-top-color: $border-color;
+          border-left: 4px solid $light;
```
MattBudz, line 220: you're already setting this on L196, here you just need to update the color:
```suggestion
          border-color: $light;
```

MattBudz (changes requested):

app/assets/stylesheets/hera/modules/_textile.scss
nicolachr: The styling of form-control overwrites the above. I couldn't think of a better approach to address this. Any ideas?

app/views/qa/issues/edit.html.erb
```diff
@@ -16,7 +16,6 @@
 
 <div class="content-container" id="issues_editor">
   <div class="note-text-inner">
-    <h4 class="header-underline">Edit issue (<%= @issue.state.humanize %>)</h4>
```
MattBudz, line 19: why do we have to remove this from the QA view?

PR #1462, opened 2025-11-27
Usage analytics 

etdsoft:
### Summary

Today we don't have a great way to understand how users are navigating Dradis, whether they're getting stuck, or what features are used (or ignored). It makes it tricky to allocate time & resources to the right tasks.

After much searching it seems that the open-source privacy-friendly analytics from Plausible - https://github.com/plausible/analytics strikes the right balance of giving us a hint about usage, without compromising privacy.

This PR adds the analytics opt in/out step to the wizard to allow users to choose to sare telemetry info.

### Check List

- [x] Added a CHANGELOG entry
- [x] Add a [CE Privacy FAQ](https://dradis.com/ce/privacy.html) page to the site.


app/controllers/setup/analytics_controller.rb
```diff
@@ -0,0 +1,22 @@
+module Setup
+  class AnalyticsController < BaseController
+    def new
+    end
```
aapomm:
```suggestion
    def new; end

```

app/views/layouts/application.html.erb
```diff
@@ -10,6 +10,7 @@
     <%= javascript_importmap_tags %>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <%= render 'layouts/analytics' %>
```
aapomm: The name of the file is a bit misleading as it isn't a layout. What about moving this to `shared` and going with a more obvious naming: `shared/_plausible_tracking` ?

MattBudz: What about `layouts/head/_plausible_tracking` or just`layouts/head/_plausible`?

etdsoft: > The name of the file is a bit misleading as it isn't a layout

The convention is that layout partials live under layouts/

I'm happy with what it is tbh.

Could be persuaded to use `layouts/shared/` but having a folder for a single file seems to be overcomplicating things. Similar to `layouts/head/`

PR #1468, opened 2025-12-03
SessionsController should inherit from SetupRequired

etdsoft:
### Summary

This is to ensure you're not attempting to log in without having run
through the Setup Wizard yet. It's a rare situation that can happen
either in Development or if in the Sandbox environment your DB gets
nuked while you were signed in. Then your session expires, and you're
sent to /login on a pristine DB, which before this PR the controller
wouldn't detect.

After this change, even in that scenario, /login redirects to /setup if
the DB becomes pristine.

The problem is that our Specs were relying on "login trickery" to bypass
the Setup Wizard rather than addressing the problem head on. This PR
also creates the necessary support to bypass the Setup Wizard steps
inside the specs (but continue to be able to test it.

### Check List

- ~~[ ] Added a CHANGELOG entry~~

### DIFF
--- a/app/controllers/setup_required_controller.rb
+++ b/app/controllers/setup_required_controller.rb
@@ -10,7 +10,7 @@ class SetupRequiredController < ApplicationController
   def setup_required_ce
     if (::Configuration.shared_password == 'improvable_dradis')
       redirect_to new_setup_password_path
-    elsif  ::Configuration.where(name: 'admin:usage_sharing').empty?
+    elsif ::Configuration.where(name: 'admin:usage_sharing').empty?
       redirect_to new_setup_analytics_path
     elsif Node.count.zero?
       # We're using this as a proxy for whether the instance has been configured
--- a/app/controllers/setup_required_controller.rb
+++ b/app/controllers/setup_required_controller.rb
@@ -10,6 +10,8 @@ class SetupRequiredController < ApplicationController
   def setup_required_ce
     if (::Configuration.shared_password == 'improvable_dradis')
       redirect_to new_setup_password_path
+    elsif  ::Configuration.where(name: 'admin:usage_sharing').empty?
+      redirect_to new_setup_analytics_path
     elsif Node.count.zero?
       # We're using this as a proxy for whether the instance has been configured
       # when we visit a project, the methodology_ and issue_library nodes are
--- a/app/views/setup/_progress.html.erb
+++ b/app/views/setup/_progress.html.erb
@@ -1,5 +1,8 @@
 <div class="setup-progress">
+  <% if defined?(Dradis::Pro) %>
+  <% else %>
   <div class="setup-progress-indicator <%= 'active' if (request.path == root_path) || request.path.include?('/password') %>"></div>
   <div class="setup-progress-indicator <%= 'active' if request.path.include?('/analytics') %>"></div>
+  <% end %>
   <div class="setup-progress-indicator <%= 'active' if request.path.include?('/kit') %>"></div>
 </div>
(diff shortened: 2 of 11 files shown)


app/views/setup/_progress.html.erb
```diff
@@ -1,5 +1,8 @@
 <div class="setup-progress">
+  <% if defined?(Dradis::Pro) %>
+  <% else %>
```
aapomm, line 3:
```suggestion
  <% unless defined?(Dradis::Pro) %>
```

spec/support/support_helper.rb
```diff
@@ -0,0 +1,19 @@
+module SupportHelper
+  def self.included(base)
+    included_ce(base)
+    included_pro(base) if defined?(Dradis::Pro)
```
aapomm: This is a no-op in CE anyway

```suggestion
    included_pro(base)
```

app/controllers/setup_required_controller.rb
```diff
@@ -10,6 +10,8 @@ def setup_required
   def setup_required_ce
     if (::Configuration.shared_password == 'improvable_dradis')
       redirect_to new_setup_password_path
+    elsif  ::Configuration.where(name: 'admin:usage_sharing').empty?
```
aapomm:
```suggestion
    elsif ::Configuration.where(name: 'admin:usage_sharing').empty?
```

spec/rails_helper.rb
```diff
@@ -82,9 +82,9 @@
   config.include ControllerMacros, type: :feature
   config.include ControllerMacros, type: :request
   # config.include SelecterHelper,   type: :feature
-  # config.include SupportHelper,    type: :controller
-  # config.include SupportHelper,    type: :feature
-  # config.include SupportHelper,    type: :request
+  config.include SupportHelper,    type: :controller
+  config.include SupportHelper,    type: :feature
+  config.include SupportHelper,    type: :request
```
aapomm:
```suggestion
  config.include SupportHelper, type: :controller
  config.include SupportHelper, type: :feature
  config.include SupportHelper, type: :request
```

spec/support/support_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def self.included_ce(base)
+    base.before(:each) do |example|
+      # Disable the setup tour
+      unless example.metadata[:skip_setup_mock]
```
aapomm, line 10: If the `skip_setup_mock` metadata doesn't exist, it mocks the `setup_required` callback. Shouldn't it be the other way around?

spec/features/sessions_spec.rb
```diff
@@ -13,13 +12,16 @@
     )
   end
 
+  # this is needed to bypass the Setup steps
+  before do
+    create(:configuration, name: 'admin:password', value: ::BCrypt::Password.create('rspec_pass'))
```
aapomm, line 17:
```suggestion
    create(:configuration, name: 'admin:password', value: ::BCrypt::Password.create(password))
```

app/views/setup/_progress.html.erb
aapomm: Apparently this empty if block is used in Pro

etdsoft: Correct

spec/features/sessions_spec.rb
etdsoft: I had this originally, but it breaks the "bad password" spec, as the `password` variable gets updated inside the block, and you end up using the "bad" one to save the config.

spec/support/support_helper.rb
etdsoft: No. You want Setup Wizard to be generally ignored, except when you're testing those controllers.

spec/rails_helper.rb
etdsoft: I uncommented and left the spacing for another day, at least we got Pro parity now.

PR #1472, opened 2025-12-09
Nodes: add more types and icons

etdsoft:
### Summary

Widen the selection of icons available to use in your nodes to make it simpler to organise your project:

Screenshot 2025-12-09 at 12 45 23

To do so, we've extracted the Node::Types module into its own file to keep things contained, and reused the helper method that displayed Issue/Evidence icons.

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/helpers/hera_helper.rb
+++ b/app/helpers/hera_helper.rb
@@ -18,7 +18,7 @@ module HeraHelper
     when Issue
       tag = model.tags.first
     when Node
-      css = ['fa-solid']
+      css = ['fa-fw fa-solid']
       css << model.icon
     end
 
--- a/app/helpers/hera_helper.rb
+++ b/app/helpers/hera_helper.rb
@@ -71,6 +71,13 @@ module HeraHelper
     end
   end
 
+  def node_types_collection
+    Node::Types::LABELS.map.with_index do |type, index|
+      next if Node::Types::SYSTEM_TYPES.include?(index)
+      [ type.titleize, index ]
+    end.compact
+  end
+
   def page_title
     [content_for(:title), "Dradis #{defined?(Dradis::Pro) ? 'Professional' : 'Community' } Edition"].compact.join(' | ')
   end
--- a/app/helpers/hera_helper.rb
+++ b/app/helpers/hera_helper.rb
@@ -72,7 +72,7 @@ module HeraHelper
   end
 
   def node_types_collection
-    Node::Types::LABELS.map.with_index do |type, index|
+    Node::Types::LABELS.filter_map.with_index do |type, index|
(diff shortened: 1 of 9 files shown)


app/helpers/hera_helper.rb
```diff
@@ -68,6 +71,19 @@ def navbar_brand
     end
   end
 
+  def node_types_collection
```
aapomm, line 74: This looks like it should be in a node-specific helper defined in a the controller (or node helper) and not in the more general hera helper

app/helpers/hera_helper.rb
```diff
@@ -68,6 +71,19 @@ def navbar_brand
...
+    Node::Types::LABELS.map.with_index do |type, index|
```
aapomm:
```suggestion
    Node::Types::LABELS.filter_map.with_index do |type, index|
```

If we use `filter_map` we don't need to call `compact` in line 84

app/models/node/types.rb
```diff
@@ -0,0 +1,82 @@
+module Node::Types
+  extend ActiveSupport::Concern
+
+  DEFAULT = 0
```
aapomm, line 4: The handling of the types looks so unwieldy. We're managing 3 different arrays with the order and indexing of each relies on one another. The consequence is that we're repeating information (label/constant name) and adding empty strings at the specific indexes for the system types.

What about using hash/object for this?


If we want to keep using indexes:
```
TYPES = [
...
  { label: 'Host', icon: 'fa-laptop' },
...
]
```

Alternatively, we can use the labels as keys:
```
TYPES = {
...
  'Host' => 'fa-laptop',
...
}
```

app/helpers/hera_helper.rb
etdsoft: Agreed, but alas, this is where the layout helpers are. There's still some legacy problems around the rendering of the Nodes tree, which isn't handled nicely by a NodesController, so again, a battle for another day.

app/models/node/types.rb
etdsoft: Both of those approaches would break the rest of the framework that has come to rely on the individual constants being defined.

I went back and forth on a few approaches to define the TYPES and ended up with this as the path of least resistance towards a self-contained set of changes.

The biggest issue with both of your suggestions is that an Addon (i.e. dradis-plugins ContentService) wants to add a HOST node. And that choice need to be stored as an `int` in the DB - respecting previous numbers stored over the years.

So on the one hand you need a programmatic way of identifying the type (e.g. `TYPES[:host]`), but also, of mapping that type to deterministic integer in the database.

PR #1480, opened 2025-12-10
Support liquid fields for issue and evidence tables

gcelanji:
### Summary
Currently, even though Liquid works as expected in Word filters and other contexts, it cannot be displayed properly in the Issues/Evidence tables.

Leverage the existing liquid-async and preview flow to render liquid content for issues and evidence field tables. 

### Check List

- [x] Added a CHANGELOG entry


app/views/issues/evidence/_table.html.erb
```diff
@@ -31,10 +34,22 @@
               when 'Updated'
                 [evidence.updated_at.to_i, local_time_ago(evidence.updated_at)]
               else
-                [evidence.fields.fetch(column, ''), markup(evidence.fields.fetch(column, ''))]
+                liquid_content = evidence.fields.fetch(column, '')
+                preview_path =  preview_project_issue_evidence_path(current_project, @issue, evidence)
```
caitmich: I'd set this in the controller like we do with `@form_preview_path`

app/views/issues/evidence/_table.html.erb
```diff
@@ -31,10 +34,22 @@
               when 'Updated'
                 [evidence.updated_at.to_i, local_time_ago(evidence.updated_at)]
               else
-                [evidence.fields.fetch(column, ''), markup(evidence.fields.fetch(column, ''))]
+                liquid_content = evidence.fields.fetch(column, '')
```
caitmich: I don't think `liquid_content` is the best name for this variable since this is the pre-liquid field value

app/views/nodes/items_table/_table.html.erb
```diff
@@ -47,10 +50,24 @@
               when 'Updated'
                 [item.updated_at.to_i, local_time_ago(item.updated_at)]
               else
-                [item.fields.fetch(column, ''), markup(item.fields.fetch(column, ''))]
+                liquid_content = item.fields.fetch(column, '')
+                if item.is_a?(Evidence)
```
caitmich: This is a smell. A partial shouldn't need to know anything about the object type it's displaying.
This also currently means that we're no longer passing the value to markup if the item isn't an instance of `Evidence`

Instead we should have `preview_path` set in the controller and `liquid-async` added if needed, otherwise ignored completely and default behaviour being `markup(content)`

Something along the lines of:

```ruby
result = liquid_behavior ? liquid_content : markup(liquid_content)
```

app/controllers/issues/evidence_controller.rb
```diff
@@ -114,6 +120,10 @@ def set_auto_save_key
     end
   end
 
+  def set_evidence
+    @evidence = Evidence.find_by!(id: params[:id], issue_id: params[:issue_id])
```
caitmich:
```suggestion
    @evidence = Evidence.find_by(id: params[:id], issue_id: params[:issue_id])
```

We don't want to raise an exception if this isn't found, instead it should be handled gracefully. In this case it probably makes sense to have `@evidence` be `nil` if it's not valid since we're only using it for `liquid_resource_assigns`

caitmich: Everywhere else in this controller/this controller's views we're getting to the evidence through the issue or the affected_nodes. We have an `:issue_id` param so should we get the evidence via the issue instead? That will ensure that it exists

app/assets/javascripts/hera/modules/liquid_async.js
```diff
@@ -16,8 +18,19 @@ document.addEventListener('turbo:load', function () {
       .then((response) => response.text())
       .then(function (html) {
         $(that).html(html);
+
+        if (requiresSortUpdate) {
```
caitmich, line 28: what about moving this into its own function?

app/views/issues/evidence/_table.html.erb
gcelanji: Because the index action only handles the collection, the controller cannot access a specific evidence instance variable to set the preview_path. Therefore, we must either add logic to iterate over all evidence records in the controller index action, or use the Evidence.new approach to generate a reusable path template for JS to update the path with the correct ID.

In this specific context, neither of those two options is worth the added complexity or performance drop. The simplest and most efficient solution is to keep the path generation in the view partial's loop, where the evidence is availiable.

app/views/nodes/items_table/_table.html.erb
gcelanji: I think we might be able to use `preview_project_node_evidence_path(current_project, @node, item)` without having to check to check if the item is an Evidence here.

> Instead we should have preview_path set in the controller

> Because the index action only handles the collection, the controller cannot access a specific evidence instance variable to set the preview_path. Therefore, we must either add logic to iterate over all evidence records in the controller index action, or use the Evidence.new approach to generate a reusable path template for JS to update the path with the correct ID.
> 
> In this specific context, neither of those two options is worth the added complexity or performance drop. The simplest and most efficient solution is to keep the path generation in the view partial's loop, where the evidence is availiable.

app/views/issues/evidence/_table.html.erb
gcelanji: Sounds good. I will update that. My initial thought for naming it was that this variable holds the text which is intended to be processed by Liquid.

PR #1483, opened 2025-12-12
Catch previous sha == zeroes

aapomm:
### Summary

There's been a change to github workflows and now it sets the previous commit as `null` (or `0000000000000000000000000000000000000000`) if it is the first commit of the branch.

**Proposed solution**
Catch the zeroes when returned as the previous sha. If zeroes are returned, treat is as nil.

### Check List

- [ ] Added a CHANGELOG entry


MattBudz:
closed in favor of https://github.com/dradis/dradis-ce/pull/1482

PR #1484, opened 2025-12-12
Application layout asset updates

MattBudz:
### Summary

Updating and adding conditional assets

### Check List


app/helpers/easter_egg_helper.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def easter_egg_logo
+    case
+    when december?
+      'december/logo_full_small.png'
+    when may_4th?
+      'may_4th/logo_full_small.png'
+    else
+      false
```
aapomm:
```suggestion
      nil
```

Functionally the same but `nil` is more consistent since this method is expected to return an object instead of boolean

PR #1489, opened 2025-12-18
Resolve db seeds conflict

gcelanji:
### Summary

This PR resolves a sync conflict between CE and Pro by using find_or_create_by for default tags to prevent crashes and adding a conditional block to load engine-specific seeds for Pro.

### Other Information

### Copyright assignment

file for the details.

### Check List

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,7 +5,6 @@ unless defined?(Dradis::Pro)
   end
 end
 
-# Seed Pro-specific engines if available
 if defined?(Dradis::Pro)
   # Load engine seeds
   Dradis::Pro::BI::Engine.load_seed
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,4 +1,13 @@
-# Create a few default tags.
-Tag::DEFAULT_TAGS.each do |name|
-  Tag.create!(name: name)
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -1,4 +1,12 @@
-# Create a few default tags.
-Tag::DEFAULT_TAGS.each do |name|
-  Tag.create!(name: name)
+unless defined?(Dradis::Pro)
+  # Create a few default tags.
+  Tag::DEFAULT_TAGS.each do |name|
+    Tag.find_or_create_by(name: name)
```
caitmich, line 4: In the CE case I don't think we need to change this, since it'll only be run once and when seeding, the tags are very likely to not exist yet

db/seeds.rb
```diff
@@ -1,4 +1,12 @@
-# Create a few default tags.
-Tag::DEFAULT_TAGS.each do |name|
-  Tag.create!(name: name)
+unless defined?(Dradis::Pro)
```
caitmich, line 1: Why check `defined?(Dradis::Pro)` twice? They're handling opposite cases so it makes sense to have a single conditional block with 2 branches
```ruby
if defined?(Dradis::Pro)
  ...
else
  ...
end
```

PR #1491, opened 2025-12-18
Create activities using EventPublisher's published events rather than `ActivityTracking`

caitmich:
### Summary

After laying the groundwork for webhooks, we updated [ActivityService](https://github.com/secroots/dradis-pro/blob/develop/app/services/activity_service.rb) to subscribe to activities by namespace and use instrumentation to publish the activities

We want to update our current activity tracking approach in issues_controller to use this activity service instead of ActivityTracking. Same for the API issues_controller

### Check List

- [x] Added a CHANGELOG entry


config/initializers/activity_service.rb
```diff
@@ -0,0 +1,5 @@
+Rails.application.reloader.to_prepare do
+  ::ActivityService.configure do |activity_service|
```
gcelanji: Do we actually need to use `::` here? Couldn't we do `ActivityService.configure do |activity_service|` instead?

engines/dradis-api/app/controllers/dradis/ce/api/v3/issues_controller.rb
```diff
@@ -2,6 +2,7 @@ module Dradis::CE::API
   module V3
     class IssuesController < Dradis::CE::API::APIController
       include ActivityTracking
```
etdsoft: is this needed?

caitmich: No, removed ✅

PR #1492, opened 2025-12-18
Warn users that node-level methodologies aren't copied over on merge

caitmich:
### Summary
Currently node-level methodologies aren't copied over when nodes are merged because we only allow one board per node.
This can be confusing for users since their methodology would end up in the trash after merge.

#### Proposed solution
Warn users that node-level methodologies aren't copied over on merge

### Check List

- [x] Added a CHANGELOG entry


app/views/nodes/modals/_merge.html.erb
```diff
@@ -18,6 +18,9 @@
           <p id="merge-under">
             <span>This will merge <strong><%= @node.label %></strong> into: <strong><span id="current-selection"></span></strong></span>
           </p>
+          <div class="alert alert-warning mb-0">
+            <p>Note that this node's methodology will not be copied over. You can find it in the Trash and recover it after if needed.</p>
```
etdsoft, line 22: Is that so @caitmich I thought it wouldn't be copied _if_ the target node already had one, but it would be copied if it didn't...

caitmich: We're not including boards in the list of descendants to move https://github.com/secroots/dradis-pro/blob/62a5896c0203536319667ee58d21c4c8e3ffb7b9/app/services/nodes/merger.rb#L43

PR #1493, opened 2025-12-19
Add Issue Library and import views

nicolachr:
### Summary

With the current implementation, there's no way for users to see how they will be able to utilize the Issue Library before upgrading to Pro. This PR, adds a few entry samples to an Issue Library view, and opens up the import view, allowing users to get a glimpse of its functionality.

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry


app/assets/javascripts/hera/behaviors.js
```diff
@@ -16,7 +16,10 @@ document.addEventListener('turbo:load', function () {
   });
 
   // If project id is changed in project path
-  if (!/^\/projects\/1(\/|$)/.test(window.location.pathname)) {
+  if (
+    !/^\/projects\/1(\/|$)/.test(window.location.pathname) &&
+    !/^\/addons\/issuelib(\/|$)/.test(window.location.pathname)
```
MattBudz: why do we need this?

app/views/static_pages/_issuelib_entries.html.erb
```diff
@@ -0,0 +1,70 @@
...
+                  class="btn btn-primary js-try-pro"
+                  data-term="issuelib"
+                  data-url="https://dradis.com/pro/pages/issuelib.html">
+                  <i class="fa-solid fa-plus"></i> Add issue (Published)
+                </a>
+                <a href="#" class="btn btn-primary dropdown-toggle dropdown-toggle-split" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                  <span class="visually-hidden">Toggle Dropdown</span>
+                </a>
+                <div class="dropdown-menu">
+                  <%= collection_radio_buttons(nil, :scope, [['draft', 'draft'], ['read_for_review', 'read_for_review'], ['published', 'published']], :first, :first) do |b| %>
```
MattBudz:
```suggestion
                  <%= collection_radio_buttons(nil, :scope, [['draft', 'draft'], ['ready_for_review', 'ready_for_review'], ['published', 'published']], :first, :first) do |b| %>
```

app/views/static_pages/import.html.erb
```diff
@@ -0,0 +1,35 @@
...
+<ol class="breadcrumb">
+  <li class="breadcrumb-item">
+    <%= link_to 'Dashboard', main_app.root_path %>
+  </li>
+  <li class="breadcrumb-item active">
+    Import Issues
+  </li>
+</ol>
+
+<!-- Main content -->
```
MattBudz:
```suggestion
```

app/views/static_pages/import.html.erb
```diff
@@ -0,0 +1,35 @@
+<% content_for :title, 'Import Issues' %>
+
+
```
MattBudz:
```suggestion
```

app/views/static_pages/import.html.erb
```diff
@@ -0,0 +1,35 @@
...
+    Import Issues
+  </li>
+</ol>
+
+<!-- Main content -->
+<div class="content-container">
+  <div id="issue-viewer">
+    <div class="results">
+      <div class="page page-none">
+
```
MattBudz:
```suggestion
```

app/views/static_pages/import.html.erb
```diff
@@ -0,0 +1,35 @@
...
+        <div class="header">
+          <div class="header-inner">
+            <h4 class="header-underline">Search results</h4>
+          </div>
+        </div>
+
+        <%= render partial: 'issuelib_entries',
+          :locals => {
```
MattBudz:
```suggestion
          locals: { 
```

app/views/static_pages/issuelib.html.erb
```diff
@@ -0,0 +1,59 @@
...
+
+<div id="issuelib_entries_table" class="card">
+  <div class="card-header bg-primary">
+    <h4 class="card-title mb-0">Issue Library entries</h4>
+  </div>
+  <div class="card-body">
+    <%= render partial: 'issuelib_entries',
+      :locals => { 
+        columns: ['Title', 'State', 'Created', 'Updated'],
+        entries: ['Auto-complete in password field', 'DOM-based cross-site scripting (XSS)', 'Insufficient cross-site request forgery (CSRF) protection', 'Reflected cross-site scripting (XSS)'],
```
MattBudz: If these columns and entries are the same as the ones you're passing [here](https://github.com/dradis/dradis-ce/pull/1493/changes#diff-e919f3415c5ba3fcd7ebb0ba113a69915825c58107cde2a7187d1a06ef92ff04R27-R28), why not just hard-code these values into the partial? Then the only local var you'll need is `import`. 

Also `'yes'` and `'no'` are #weaksauce. Use a `true` or `false` boolean here

app/assets/javascripts/hera/behaviors.js
nicolachr: To hide the teaser in the IL static view.

Image

app/views/static_pages/issuelib.html.erb
nicolachr: Re: the columns, would it be better to hard-code the first two that are shared, and then pass the rest as locals?

Re: entries, isn't cleaner like this? Otherwise we'll have the same code repeated for each entry. Am I missing something?

MattBudz: if the colums are different let's leave the columns as it (passed to the partial), but if the entrires are all the same then just hard code them in the partial. Right now, in the partial you have:

```
<% entries.each_with_index do |entry, index| %>
```

and since the passed in entries are always the same, rather than passing the same entries each time, we can change this line to:

```
<%  ['Auto-complete in password field', 'DOM-based cross-site scripting (XSS)', 'Insufficient cross-site request forgery (CSRF) protection', 'Reflected cross-site scripting (XSS)'].each_with_index do |entry, index| %>
```
or a bit cleaner
```
<% entries = ['Auto-complete in password field', 'DOM-based cross-site scripting (XSS)', 'Insufficient cross-site request forgery (CSRF) protection', 'Reflected cross-site scripting (XSS)'] %>
<% entries.each_with_index do |entry, index| %>
```

then when you call the partial, you can have:
```
 <%= render partial: 'issuelib_entries',
          locals: { 
            columns: ['Title', 'State', ''],
            import: true
          }
        %>
```

PR #1499, opened 2026-01-21
add caching to navbar to improve layout rendering

MattBudz:
### Summary

The navbar contains mostly static content but is rerendered on each page load from a number of partials. This PR implements caching to avoid rerendering static content from multiple partials.

### Check List

### DIFF
--- a/app/views/layouts/hera/navbar/_main_nav.html.erb
+++ b/app/views/layouts/hera/navbar/_main_nav.html.erb
@@ -1,4 +1,4 @@
-<% cache(['navbar', Dradis::Plugins.with_feature(:addon), current_user.notifications.unread.count]) do %>
+<% cache(['navbar', Dradis::Plugins.with_feature(:addon), current_user.notifications.unread.size]) do %>
   <nav class="navbar navbar-expand-lg main-nav">
     <div class="container-fluid">
       <%= navbar_brand %>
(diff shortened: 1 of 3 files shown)


app/views/layouts/hera/navbar/main_nav/utility_nav/_ce.html.erb
```diff
@@ -7,23 +7,26 @@
...
-    }, 
-    role: 'button', 
-    'aria-expanded': 'false' do %>
-    <i class="notifications-bell fa-solid fa-bell d-none d-lg-inline" title="Notifications"><i class="notifications-dot d-none" data-behavior="notifications-dot"></i></i>
-    <span class="d-inline d-lg-none">Notifications</span>
-  <% end %>
-  <div class="dropdown-menu dropdown-menu-end" data-url="<%= main_app.notifications_path %>" aria-labelledby="notifications">
-  </div>
-</li>
+<% cache(['navbar-notifications', current_user.notifications.unread.size]) do %>
```
etdsoft: I think it's fine to let Rails handle it:

```suggestion
<% cache(['navbar-notifications', current_user.notifications.unread]) do %>
```

app/views/layouts/hera/navbar/_main_nav.html.erb
```diff
@@ -1,17 +1,19 @@
+<% cache(['navbar', Dradis::Plugins.with_feature(:addon), current_user.notifications.unread.size]) do %>
```
etdsoft: Do you really need the dynamic keys here? I suspect not, because the internal fragments will be busted and re-rendered. Worth trying w/o args 2 and 3.

gcelanji: Removed the extra keys. Everything renders correctly.

app/views/layouts/hera/navbar/main_nav/utility_nav/_ce.html.erb
gcelanji: Updated that!

PR #1500, opened 2026-01-22
Keep sidebar open in larger viewports

nicolachr:
### Summary

This PR ensures that the sidebar stays open in larger viewports when users are editing entries. More specifically:

- In viewports < 1510px both sidebars collapse to allow enough horizontal space for editing
- In viewports >= 1510px both sidebars maintains their original state

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry


app/assets/javascripts/hera/modules/sidebar.js
```diff
@@ -6,6 +6,7 @@
       this.$toggleLink = $sidebar.find($('[data-behavior~=sidebar-toggle]'));
       this.$resizeHandle = $sidebar.find('[data-behavior~=resize-handle]');
       this.mobileBreakpoint = 992;
+      this.sidebarBreakpoint = 1510;
```
MattBudz: This could use a slightly clearer name. `mobileBreakpoint` is clear (at this point, we are either mobile or desktop). What is `sidebarBreakpoint`? we know today, but will we when we look at this in a year or two?

app/assets/javascripts/hera/modules/sidebar.js
```diff
@@ -32,8 +33,15 @@
         }
       });
 
-      if ($('[data-behavior~=local-auto-save]').length) {
-        that.close(true);
+      if (
+        $('[data-behavior~=local-auto-save]').length &&
+        $(window).width() < that.sidebarBreakpoint
```
MattBudz:
```suggestion
        $(window).width() < this.sidebarBreakpoint
```

`this` should work here. If this code was inside of a listener or callback function, then you'd need `that`.

app/assets/javascripts/hera/modules/sidebar.js
```diff
@@ -32,8 +33,15 @@
...
+      ) {
+        if (this.storageKey == 'secondary-sidebar-expanded') {
+          this.getValidation(this.storageKey);
```
MattBudz: what's the storagekey being passed in for?

app/assets/javascripts/hera/modules/sidebar.js
```diff
@@ -175,6 +183,22 @@
     toggle(openSidebar) {
       openSidebar ? this.open() : this.close();
     }
+
+    getValidation(storageKey) {
```
MattBudz: Is this name describing what this function is doing? You are saying it's 'getting the validation', but the function is closing the sidebar based on the validation result. Maybe something along the lines of closeOnValidationSuccess() is more appropriate here.

Also, why does this function need a storageKey passed in if it never uses it?

app/assets/javascripts/hera/modules/sidebar.js
```diff
@@ -175,6 +183,22 @@
...
+      const $validationFeed = $('[data-behavior~=validation-feed]');
+      if ($validationFeed.length === 0) {
+        this.close(true);
+      }
+      $validationFeed.on('dradis:validation-loaded', () => {
```
MattBudz:
```suggestion
      $validationFeed.one('dradis:validation-loaded', () => {
```

we only need to do this once, the feed doesn't get reloaded multiple times

app/assets/javascripts/hera/modules/sidebar.js
```diff
@@ -175,6 +183,22 @@
     toggle(openSidebar) {
       openSidebar ? this.open() : this.close();
     }
+
+    getValidation(storageKey) {
+      const $validationFeed = $('[data-behavior~=validation-feed]');
+      if ($validationFeed.length === 0) {
+        this.close(true);
```
MattBudz, line 190:
```suggestion
        this.close(true);
        return;
```

We also need to return here. If there is no validation feed, close immediately and don't continue with the rest.

MattBudz (changes requested):

app/assets/javascripts/hera/modules/sidebar.js
nicolachr: Since mobileBreakpoint is when it collapses for mobile, I think desktopBreakpoint would be a better name that sidebarBreakpoint.

app/assets/javascripts/hera/modules/sidebar.js
nicolachr: Forgot to remove it.

app/assets/javascripts/hera/modules/sidebar.js
nicolachr: Applied your suggestion re: naming the function. 

> Also, why does this function need a storageKey passed in if it never uses it?
I had originally tried something else, and forgot to remove it.

PR #1503, opened 2026-01-29
Setup::Kits#new - offer more options to choose from

etdsoft:
### Summary

From a single Welcome kit to a variety of options to try things go. The goal is to let people see how Dradis can be configured for different assessment types.

### Check List

- [ ] Added a CHANGELOG entry


app/controllers/setup/kits_controller.rb
```diff
@@ -9,12 +9,13 @@ def create
       case @kit
       when :none
         Tag::DEFAULT_TAGS.each { |name| Tag.create!(name: name) } unless defined?(Dradis::Pro)
-      when :welcome
-        kit_folder = Rails.root.join('lib', 'tasks', 'templates', 'welcome').to_s
-        logger = Log.new.info('Loading Welcome kit...')
+      when :owasp, :welcome
+        kit_folder = Rails.root.join('lib', 'tasks', 'templates', @kit.to_s)
```
aapomm, line 13: We're converting the params from string to symbols (L46) then to string again here. What about skipping the symbols step entirely and just handle strings?

etdsoft: the only minimal advantage is the use of `@kit` in a case/when, having symbols as keys is slightly better than matching strings. If anyone in the review chain feels strongly, I don't mind either way.

PR #1504, opened 2026-01-30
Add support for API search for Issues

aapomm:
### Spec
Search isn't supported for the Issues/Entries API.

**Proposed solution**
Extend the #index API for issues and entries to support a query param. When the param is passed, use it as a search query for filtering the records. For now, let's only use the text/content for searching.

### Check List

- [x] Added a CHANGELOG entry


app/models/search.rb
```diff
@@ -64,23 +67,24 @@ def evidence
...
   def issues
-    @issues ||= Issue.where(
-      "node_id = :node AND LOWER(text) LIKE LOWER(:q)",
-      node: project.issue_library,
-      q: "%#{query}%"
-    ).order(updated_at: :desc)
+    @issues ||=
+      begin
+        issues = Issue.where(node: project.issue_library)
+        issues = issues.where('LOWER(text) LIKE LOWER(:q)', q: "%#{query}%") if query
```
gcelanji: should we use if` query.present?` instead of `if query` to ensure that an empty search string (?q\=) doesn't force the database to perform a slower LIKE '%%' comparison across the whole table?

gcelanji: Also what about extracting this into the model as a scope (like search_by_query) and then reusing that in this Search class? I know the rest of the methods currently handle the queries directly within the Search class, but it might be worth considering to keep the model's schema details (like whether to search text, content, or label) encapsulated in the model itself.

engines/dradis-api/app/controllers/dradis/ce/api/v3/issues_controller.rb
```diff
@@ -8,8 +8,14 @@ class IssuesController < Dradis::CE::API::APIController
...
-        @issues = current_project.issues.includes(:tags).order('updated_at desc')
-        @issues = @issues.page(params[:page].to_i) if params[:page]
+        @issues = Search.new(
+          query: params[:q],
+          scope: :issues,
+          page: params[:page],
+          project: current_project
+        ).results
+
+        @issues = @issues.includes(:tags).order('updated_at desc')
```
gcelanji: Can the tags N+1 call be moved into the Search class as well?

app/models/search.rb
aapomm: I believe that's the idea behind the Search class so we have all the search logic in one place. I do see the idea with adding a model scope and I'd be happy to do that should we need to implement more model-specific logic.

PR #1509, opened 2026-02-05
Add user model hook

aapomm:
### Summary

Some integrations, like dradis-echo, might want to programmatically hook to the User model such as when adding an integration-only model/association.

### Check List

- [ ] Added a CHANGELOG entry


app/views/layouts/hera.html.erb
```diff
@@ -60,7 +60,10 @@
         <main class="container-fluid">
           <%= content_for?(:breadcrumbs) ? yield(:breadcrumbs) : '' %>
           <%= render 'layouts/hera/noscript' %>
-          <%= render "layouts/hera/project_teaser" unless defined?(Dradis::Pro) || controller_path == 'styles' %>
+          <% # Only render the project teaser in CE and in project ID != 1 %>
+          <% if !defined?(Dradis::Pro) && request.path =~ /\A\/projects\/(?!1\z)\d+\z/ %>
```
MattBudz: shouldn't this be handled [here](https://github.com/dradis/dradis-ce/blob/develop/app/assets/javascripts/hera/behaviors.js#L18-L21)?

aapomm: I'm thinking this is better handled server-side. That way we can check for the edition as well as skipping rendering of the warning

etdsoft: sorry, but this doesn't work as intented @aapomm 

http://localhost:3000/projects/2/issues/10

doesn't trigger, when it should.

PR #1513, opened 2026-02-13
Prefer `RAILS_MASTER_KEY` from environment rather than stored in a file and move credentials to `config/credentials/`

caitmich:
### Summary

- Get `active_record_encryption` values from environment and if not set, fallback to credentials
- Update readme to include creating credentials before running image

### Check List

- [x] Added a CHANGELOG entry


README.md
etdsoft, line 1: I don't think we need this. Presumably, people running CE Docker are using a quick'n'dirty approach to containers to spin it up for testing, or for a single project, not to run long-term infrastructure.

We could add a note about RAILS_MASTER_KEY env variable, but Rails does that anyway already, so I don't know if that's needed at all.

config/initializers/credentials.rb
```diff
@@ -25,9 +25,12 @@
     secret_key_base: SecureRandom.hex(64)
   }.to_yaml
 
-  key = ActiveSupport::EncryptedConfiguration.generate_key
-  key_path.binwrite(key)
-  key_path.chmod 0600
+  # if master key is not set in env and there is no master.key file, create one
```
etdsoft:
```suggestion
  # if master key is not set in env, create one
```

afaict, you're not checking key_path existance (not needed).

etdsoft (commented):
I see a bunch of config/shared/ in several files of this PR, can we remove them all?

config/application.rb
```diff
@@ -48,8 +48,9 @@ class Application < Rails::Application
     config.action_view.form_with_generates_remote_forms = true
 
     # Override the default credentials lookup paths. See bin/rails credentials:help
-    config.credentials.content_path = Rails.root.join('config', 'shared', 'credentials.yml.enc')
-    config.credentials.key_path = Rails.root.join('config', 'shared', 'master.key')
+    config.credentials.content_path = Rails.root.join('config', 'credentials', 'credentials.yml.enc')
+    # the following can be removed when we drop support for VM instances, after which master key will be an environment variable
```
etdsoft, line 51: nice touch, but similar comment is missing from initializer

README.md
caitmich: Yeah, I suppose users can figure out that passing env variables is possible. Removed 👍

compose.yaml
```diff
@@ -9,9 +9,7 @@ services:
     ports:
       - 443:443
       - 80:80
-    environment:
-      # Set domain from .env so thruster can handle let's encrypt certificates
-      - TLS_DOMAIN=${TLS_DOMAIN}
+    env_file: .env
```
etdsoft: should we be providing, even an empty .env in the repo? Otherwise this may break.

```
#TLS_DOMAIN=dradis.dev
```

Also, isn't .env used by default by Docker? Or do we need to explicitly declare it as a config option?

caitmich: It is but in order to have the variables available in the container at runtime you have to either specify them like we were doing with `TLS_DOMAIN` or explicitly declare `env_file`. Since there is now possibly more than one environment variable that we want to pass to the container, I'm specifying `env_file`

Instead of including `.env` in the repo, I added `required: false` since this can be run without any environment variables

PR #1516, opened 2026-02-20
Set smtp config using env variables instead of `smtp.yml`

caitmich:
### Summary

Remove the need for `smtp.yml`/`smtp.yml.template` and [z_03_setup_mail.rb](https://github.com/secroots/dradis-pro/blob/develop/config/initializers/z_03_setup_mail.rb) initializer and take advantage of env variables to directly set smtp settings on startup

### Check List

- [x] Added a CHANGELOG entry


etdsoft:
## PR Review: Set SMTP config using env variables instead of `smtp.yml`

**Overall**: Solid modernization that aligns with 12-factor app principles. The direction is good. A few issues below ranging from minor to moderate.

---

### Issues

#### 1. `SMTP_PORT` is a string, not an integer (Bug)

**File:** `config/environments/production.rb`

```ruby
port: ENV.fetch('SMTP_PORT', 587),
```

`ENV.fetch` always returns a string. When someone sets `SMTP_PORT=465`, the port value will be `"465"` rather than `465`. Should be:

```ruby
port: ENV.fetch('SMTP_PORT', '587').to_i,
```

#### 2. Misleading env variable name: `SMTP_HOST` vs `SMTP_ADDRESS` (Design)

`SMTP_ADDRESS` is used for the SMTP server address and `SMTP_HOST` for the app's public-facing hostname (used in mailer-generated links). This is confusing — `SMTP_HOST` sounds like the SMTP server hostname.

Consider renaming to something like `MAILER_HOST` or `APP_HOST`:

```ruby
config.action_mailer.default_url_options = {
  host: ENV.fetch('MAILER_HOST', 'dradisframework.dev'),
  ...
}
```

#### 3. `openssl_verify_mode` defaults to `'none'` (Security)

```ruby
openssl_verify_mode: ENV.fetch('SMTP_SSL_VERIFY_MODE', 'none')
```

Defaulting to `'none'` disables certificate verification, making SMTP connections vulnerable to MITM attacks. Consider defaulting to `'peer'`:

```ruby
openssl_verify_mode: ENV.fetch('SMTP_SSL_VERIFY_MODE', 'peer')
```

Users who need to skip verification (e.g., self-signed certs) can explicitly set `SMTP_SSL_VERIFY_MODE=none`.

#### 4. `.gitignore` / `.dockerignore` still reference `smtp.yml` (Cleanup)

Both `.gitignore` and `.dockerignore` still contain `/config/smtp.yml`. Since the file and template are being removed, these entries are now dead. Worth cleaning up.

#### 5. `authentication` should be a symbol (Bug)

```ruby
authentication: ENV.fetch('SMTP_AUTHENTICATION', 'plain'),
```

ActionMailer expects `authentication` as a symbol (`:plain`, `:login`, `:cram_md5`). Should be:

```ruby
authentication: ENV.fetch('SMTP_AUTHENTICATION', 'plain').to_sym,
```

#### 6. Missing `enable_starttls_auto` option

The old `smtp.yml.template` included `enable_starttls_auto: true`, but the new config doesn't expose this. While Rails defaults to `true`, it might be worth adding an env variable (e.g., `SMTP_ENABLE_STARTTLS`) for parity and for environments that need to disable it.

---

### What looks good

- **Conditional SMTP block**: `if smtp_address = ENV['SMTP_ADDRESS'].presence` is clean — SMTP only activates when configured.
- **Test environment defaults**: Adding explicit `default_options` and `default_url_options` in `test.rb` properly replaces what the deleted initializer provided.
- **Removing `Configuration.mail_host`**: Confirmed this method is never called anywhere — safe to remove.
- **CHANGELOG entry**: Present and descriptive.

---

### Summary

| Category | Finding | Severity |
|----------|---------|----------|
| Bug | `SMTP_PORT` returns string, needs `.to_i` | Medium |
| Bug | `authentication` returns string, needs `.to_sym` | Medium |
| Security | `openssl_verify_mode` defaults to `'none'` | Medium |
| Design | `SMTP_HOST` name is misleading | Low |
| Cleanup | `.gitignore`/`.dockerignore` still reference `smtp.yml` | Low |
| Design | Missing `enable_starttls_auto` env variable | Low |

PR #1523, opened 2026-02-26
fix nav render outside projects

MattBudz:
### Summary

The navbar called \`current_project\` in several places (brand link, trash, config, notifications). \`current_project\` was only exposed as a helper via \`ProjectScoped\`, so non-project views like \`StylesController\` and \`StaticPagesController\` had to \`include ProjectScoped\` just to keep the navbar from raising a \`NameError\`, even though they have no actual need for project-scoped behaviour.

The fix removes the \`ProjectScoped\` dependency from those controllers and guards the project-specific utility nav items with \`defined?(current_project)\` — the same pattern already used in the \`Mentioned\` and \`MultipleDestroy\` concerns. Adding \`helper :hera\` to \`AuthenticatedController\` (already present in Pro) ensures \`HeraHelper\` is available to all Hera layout views without needing \`ProjectScoped\`.

For \`navbar_brand\`, where CE always needs a link target, the \`Project.new\` fallback is applied locally within the method. \`Project.new\` always has \`id: 1\` by design — it is a plain Ruby value object, not an ActiveRecord model — so the fallback generates a valid URL on non-project pages.

This mirrors how Pro guards project-specific nav items with \`content_for?(:project_utility_nav)\`, keeping the sync clean with no \`defined?(Dradis::Pro)\` checks needed.

### Check List

### DIFF
--- a/app/controllers/authenticated_controller.rb
+++ b/app/controllers/authenticated_controller.rb
@@ -16,6 +16,7 @@ class AuthenticatedController < SetupRequiredController
   before_action :set_paper_trail_whodunnit
   # before_action :render_onboarding_tour
 
+  helper :hera
   layout 'hera'
 
   # This is a central location where we can manage authorization errors (e.g.
--- a/app/controllers/static_pages_controller.rb
+++ b/app/controllers/static_pages_controller.rb
@@ -1,6 +1,4 @@
 class StaticPagesController < AuthenticatedController
-  include ProjectScoped
-
   before_action :set_entries, only: [:issuelib_index, :issuelib_import]
   before_action :set_tickets, only: [:remediationtracker_index]
 
--- a/app/helpers/hera_helper.rb
+++ b/app/helpers/hera_helper.rb
@@ -61,7 +61,8 @@ module HeraHelper
 
   def navbar_brand
     if !defined?(Dradis::Pro)
-      link_to main_app.project_path(current_project), class: 'navbar-brand' do
+      project = defined?(current_project) ? current_project : Project.new
+      link_to main_app.project_path(project), class: 'navbar-brand' do
         image_tag 'logo_small.png', alt: 'Dradis CE logo', class: 'p-lg-0'
       end
     else
--- a/app/views/layouts/hera.html.erb
+++ b/app/views/layouts/hera.html.erb
@@ -48,12 +48,8 @@
     ) do %>
 
     <header class="sticky-top">
-      <% if controller_path == 'styles' %>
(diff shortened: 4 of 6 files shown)


etdsoft:
## Acceptance Test Results — PR #1523

**Branch:** `hera/fix-nav-render-outside-projects` | **Date:** 2026-04-22 | **Result:** 6/6 PASS, 0 FAIL, 1 SKIPPED

| # | Step | Status | Notes |
|---|------|--------|-------|
| 1 | `StaticPagesController` does not include `ProjectScoped` | ✅ PASS | Confirmed via `rails runner` |
| 2 | `HeraHelper#current_project` exists and returns `Project.new` fallback | ✅ PASS | Returns `Project` with id:1 when `@current_project` is nil |
| 3 | `AuthenticatedController._helpers` includes `HeraHelper` | ✅ PASS | `helper :hera` wires it in correctly |
| 4 | Issue Library renders 200 with full `main_nav` navbar | ✅ PASS | `GET /projects/1/addons/issuelib` → 200, `<nav class="navbar ... main-nav">`, brand href `/projects/1` |
| 5 | Remediation Tracker renders 200 with full `main_nav` navbar | ✅ PASS | `GET /projects/1/addons/remediationtracker` → 200, same navbar structure |
| 6 | Project-scoped page still renders correctly (regression) | ✅ PASS | `GET /projects/1` → 200, navbar intact |
| 7 | Styles page (`/styles`) renders main navbar instead of old placeholder | ⏭️ SKIPPED | `/styles` is `development`-only route; Docker dev stack runs Rails in `production` mode |

### Skipped: Styles page via HTTP

The `/styles` route is guarded by `if Rails.env.development?` in `routes.rb`. The CE dev Docker stack runs in `production` mode, so the route isn't mounted. Verified indirectly:
- `_header.html.erb` partial is absent from the container
- `controller_path == 'styles'` branch removed from layout
- `StylesController` does not include `ProjectScoped`
- `HeraHelper#current_project` fallback returns a valid `Project` for any non-scoped controller

### Notes
- `Project.new.id == 1` confirmed — CE single-project design, fallback URLs are valid.
- Brand link on non-project pages correctly resolves to `/projects/1` via the `Project.new` fallback.
- No `NameError`, `NoMethodError`, or 500s on any tested page.
- The "Back to the ship" text seen in raw HTML is from the custom 404 error page, not a navbar artifact.

MattBudz:
Closing in favour of a fresh PR with a clean commit history after a change of approach. The new implementation uses defined?(current_project) guards in the view rather than a HeraHelper fallback.

PR #1525, opened 2026-03-02
Add InlineCommentThread model, controllers, and routes

etdsoft:
## Summary

Add GitHub-style inline comment threads on QA Issues. Users can select text in the issue body, anchor a comment thread to that selection, and collaborate through replies, resolve/reopen, and delete.

**Backend (this branch — `comment_threads`):**
- `inline_comment_threads` table with JSON-serialized TextQuoteSelector anchor, status enum (open/resolved), version tracking for outdated detection
- `thread_id` foreign key on `comments` table for threaded replies
- `InlineCommentThread` model with anchor validation, position coercion, `resolve!`/`reopen!`, `outdated?`
- RESTful nested controllers under `QA::InlineThreads` namespace:
  - `InlineThreadsController` — index (JSON), show, create, destroy
  - `InlineThreads::CommentsController` — create (replies)
  - `InlineThreads::ResolutionsController` — create (resolve), destroy (reopen)
- Activity feed support: activity partial, presenter icon, EventPublisher integration
- Model specs (15 examples) and request specs (8 examples)

**Frontend (Turbo branch — `comment-threads_turbo`):**
- `InlineThreadSelector` — text selection → anchor building, with field-map for selections spanning `#[Field]#` markers
- `InlineThreadHighlighter` — TreeWalker-based DOM highlighting with multi-node segment support
- `InlineThreadTurbo` — coordinator: Turbo Frame panel, fetch/highlight lifecycle, QuoteSelector coexistence
- Turbo Frames + Streams for all CRUD operations (no page reloads)
- Highlights survive Liquid async re-rendering via `dradis:liquid-rendered` listener
- SCSS styles for highlights (open/resolved/outdated), floating Add Comment button, slide-out panel

## Test plan

**Automated:**
- `bundle exec rspec spec/models/inline_comment_thread_spec.rb` — 15 examples (validations, associations, resolve/reopen, outdated, quoted_text)
- `bundle exec rspec spec/requests/qa/issues_spec.rb` — 8 examples (CRUD, resolution, comments, JSON index)

## Acceptance

**Prerequisites:** Two user accounts (e.g. admin + author), at least one QA Issue in "Ready for review" state with multi-field content (Title, Description, etc.)

**How to test**

### Creating a thread

1. Sign in as User A. Navigate to a project's QA Issues page and open an issue
2. Wait for the Liquid content to finish rendering (spinner disappears)
3. Select some text within a single field value (e.g. a phrase in the Description)
4. Verify an "Add Comment" button appears above the selection
5. Click "Add Comment" — a slide-out panel should appear on the right with a form showing the selected text in a blockquote, a textarea, and "Create Thread" / "Cancel" buttons
6. Type a comment and click "Create Thread"
7. Verify the panel now shows the thread: status badge ("Open"), the quoted text, your comment with author name and timestamp, Resolve/Delete buttons, and a reply form
8. Verify the selected text in the issue body is now highlighted in yellow
9. Close the panel (X button or Escape key). The highlight should remain

### Cross-field selection

10. Select text that spans a field name and its value (e.g. drag from the end of one field's content across the bold field name and into the next value)
11. Click "Add Comment", type a comment, and create the thread
12. Verify the thread is created successfully and the highlight appears across both the field name and value (multiple highlighted segments)

### Viewing and replying

13. Click on any yellow highlight in the issue body
14. Verify the panel opens showing the full thread with all comments
15. Type a reply in the "Reply..." textarea and click "Reply"
16. Verify the new comment appears in the thread immediately (no page reload)

### Resolve / Reopen

17. In an open thread, click the "Resolve" button (green outline, checkmark icon)
18. Verify the status badge changes to "Resolved" (green), the button changes to "Reopen", and the highlight in the issue body changes to green with a dashed border
19. Click "Reopen" — verify the thread returns to "Open" status, yellow badge, and solid yellow highlight

### Multi-user experience

20. Sign in as User B in a different browser
21. Navigate to the same QA Issue
22. Verify all highlights from User A's threads are visible
23. Click a highlight — verify User B can view the thread and reply
24. Have User B create a new thread on different text — verify it works
25. Verify User B does NOT see a "Delete" button on User A's thread (only the thread creator can delete)

### Deletion

26. Back as User A, open one of User A's threads by clicking the highlight
27. Click "Delete" — verify a browser confirmation dialog appears: "Delete this thread and all its comments?"
28. Confirm deletion — verify the panel closes and the highlight disappears from the issue body
29. Reload the page — verify the deleted thread's highlight is gone

### Activity feed

30. Check the activity feed (sidebar or activity page)
31. Verify entries appear for thread creation, resolution, and reopening with correct verbs and links back to the issue

### Edge cases

32. Select text, click "Add Comment", then click "Cancel" — verify the panel closes and no thread is created
33. Reload the page — verify all existing thread highlights re-appear after Liquid rendering completes
34. Select the same text that already has a thread — verify a second thread can be created (overlapping anchors are allowed)

### Check List

- [x] Added a CHANGELOG entry
- [x] Commit message has a detailed description of what changed and why


app/models/inline_comment_thread.rb
```diff
@@ -0,0 +1,101 @@
...
+  # -- Callbacks ------------------------------------------------------------
+  before_validation :coerce_anchor_position
+
+  # -- Validations ----------------------------------------------------------
+  validates :anchor, presence: true
+  validates :issue, presence: true
+  validate :anchor_schema_valid
+
+  # -- Scopes ---------------------------------------------------------------
+  scope :for_issue, ->(issue) { where(issue_id: issue.id) }
```
aapomm: Where is this used and do we even need it? Wouldn't this association method just work: `issue.inline_comment_threads`

app/models/inline_comment_thread.rb
```diff
@@ -0,0 +1,101 @@
...
+  include Eventable
+
+  REQUIRED_ANCHOR_KEYS = %w[type exact prefix suffix position].freeze
+
+  serialize :anchor, coder: JSON
+
+  enum :status, { open: 0, resolved: 1 }
+
+  # -- Relationships --------------------------------------------------------
+  belongs_to :issue
```
aapomm: This needs to be polymorphic for content blocks support, no?

app/models/inline_comment_thread.rb
```diff
@@ -0,0 +1,101 @@
...
+  end
+
+  def project
+    issue.project
+  end
+
+  def local_event_payload
+    {
+      anchor: anchor,
+      issue: {
```
aapomm: Ditto on the polymorphic comment

app/models/inline_comment_thread.rb
```diff
@@ -0,0 +1,101 @@
+class InlineCommentThread < ApplicationRecord
+  include Eventable
+
+  REQUIRED_ANCHOR_KEYS = %w[type exact prefix suffix position].freeze
```
aapomm: :alpha

app/views/qa/issues/show.html.erb
```diff
@@ -43,6 +47,16 @@
       <% end %>
     </div>
 
+    <div class="offcanvas offcanvas-end" tabindex="-1" data-behavior="inline-thread-panel" data-bs-backdrop="false" data-bs-scroll="true" aria-labelledby="inlineThreadPanelLabel">
+      <div class="offcanvas-header">
+        <h6 class="offcanvas-title">Inline Comment Thread</h6>
```
aapomm: Externally, let's call this just `Threads`

```suggestion
        <h6 class="offcanvas-title">Thread</h6>
```

app/views/qa/inline_threads/resolutions/create.turbo_stream.erb
```diff
@@ -0,0 +1,3 @@
+<%= turbo_stream.update 'inline_thread_content' do %>
+  <%= render 'qa/inline_threads/thread', thread: @thread, issue: @issue %>
```
aapomm: Wouldn't this update the whole frame, comments and all? What about just re-rendering the badge section?

app/views/qa/inline_threads/resolutions/destroy.turbo_stream.erb
```diff
@@ -0,0 +1,3 @@
+<%= turbo_stream.update 'inline_thread_content' do %>
+  <%= render 'qa/inline_threads/thread', thread: @thread, issue: @issue %>
```
aapomm: Ditto

app/views/qa/inline_threads/comments/create.turbo_stream.erb
```diff
@@ -0,0 +1,5 @@
+<% if @comment.persisted? %>
```
aapomm: It seems we're passing the burden to the View to render if `save` is successful or not. I would think this is the controller's job.

app/views/qa/inline_threads/create.turbo_stream.erb
```diff
@@ -0,0 +1,5 @@
+<% if @thread.persisted? %>
```
aapomm: Ditto on the controller's responsibility for rendering the view

app/views/qa/inline_threads/destroy.turbo_stream.erb
```diff
@@ -0,0 +1 @@
+<%= turbo_stream.update 'inline_thread_content', '' %>
```
aapomm: What about

```suggestion
<%= turbo_stream.remove 'inline_thread_content'%>
```

app/assets/javascripts/hera/modules/inline_thread_turbo.js
```diff
@@ -0,0 +1,158 @@
...
+    // After any Turbo form submission in our panel, re-highlight threads.
+    // Turbo Streams have already updated the panel DOM by this point.
+    document.addEventListener('turbo:submit-end', (e) => {
+      if (!e.target.closest('[data-behavior~=inline-thread-panel]')) return;
+      if (!e.detail.success) return;
+
+      this.fetchAndHighlight();
+
+      // Close panel if the thread was deleted (frame is now empty)
+      setTimeout(() => {
```
aapomm: What is this delay for? If this is to wait for the re-highlighting, we should use promises instead.

app/assets/javascripts/hera/modules/inline_thread_turbo.js
```diff
@@ -0,0 +1,158 @@
...
+      <div class="d-flex gap-2">
+        <button type="submit" class="btn btn-sm btn-primary">Create Thread</button>
+      </div>
+    </form>`;
+  }
+
+  buildAnchorFields(anchor) {
+    const prefix = 'inline_comment_thread[anchor]';
+
+    return `<input type="hidden" name="${prefix}[type]" value="${this.escapeHtml(anchor.type)}">
```
aapomm: Looks like we're shooting ourself in the foot. We're adding a hacky way of escaping the HTML because we're building the anchor fields and the thread form using raw strings. Instead, we should build the fields/form with `document.createElement`:

```
el = document.createElement('input')
el.setAttribute('name', ...)
el.setAttribute('value', ...)
```

then the JS handles the escaping for us.

app/assets/javascripts/hera/modules/inline_thread_turbo.js
```diff
@@ -0,0 +1,158 @@
...
+        }
+      }, 100);
+    });
+  }
+
+  // -- Helpers -------------------------------------------------------------
+
+  buildNewThreadForm(anchor) {
+    return `<blockquote class="thread-quoted-text border-start border-3 border-primary ps-3 text-muted fst-italic">
+      ${this.escapeHtml(anchor.exact)}
```
aapomm: See comment on using `document.createElement` instead to skip the need for escaping

app/models/inline_thread.rb
```diff
@@ -0,0 +1,101 @@
...
+    anchor&.dig('exact')
+  end
+
+  def outdated?
+    return false if version_id.nil?
+
+    latest_version = issue.versions.where(event: 'update').last
+    return false unless latest_version
+
+    version_id < latest_version.id
```
aapomm: I guess this would work since primary keys are sequential but why are we using the paper_trail versions for comparison instead of just timestamps of the `updated_at` ? When a thread is created, snapshot the `updated_at` of the issue. Then compare in this method.

app/assets/javascripts/hera/modules/inline_thread_highlighter.js
```diff
@@ -0,0 +1,161 @@
...
+      const segEnd = Math.min(matchEnd, entry.endIndex) - entry.startIndex;
+
+      segments.push({
+        node: entry.node,
+        startOffset: segStart,
+        endOffset: segEnd
+      });
+    }
+
+    return segments.length > 0 ? segments : null;
```
aapomm: Since `segments` is initialized as `[]`, can we just do this:

```suggestion
    return segments;
```

Then in line 38 we can skip the `null` check

app/assets/javascripts/hera/modules/inline_thread_selector.js
```diff
@@ -0,0 +1,249 @@
...
+  // the field name, plus a position map from stripped→raw indices. This lets
+  // us find user selections (which see rendered text without markers) and
+  // map back to raw text positions for anchoring.
+  _buildFieldMap() {
+    const raw = this.rawText.replace(/\r\n/g, '\n');
+    this._normalizedRaw = raw;
+    this._strippedText = '';
+    this._strippedToRaw = [];
+
+    const fieldRegex = /#\[([^\]]*?)\]#/g;
```
aapomm: 🚨🚨🚨 This methods replicates what FieldParser is already doing. What about doing this server-side? We render the raw text in the view so all this JS will do is fetch it.

app/assets/javascripts/hera/modules/inline_thread_turbo.js
MattBudz: we needed a delay because the closeing was triggered when the thread was deleted. I changed to just close as soon as delete is clicked.

app/models/inline_comment_thread.rb
MattBudz: not used, removed

app/views/qa/inline_threads/resolutions/create.turbo_stream.erb
MattBudz: sorted, now its just the heading bar that updates

app/views/qa/inline_threads/destroy.turbo_stream.erb
MattBudz: we dont need this anymore since we close the panel via turbo:submit-start on the delete button, and hidden.bs.offcanvas already clears the frame content.

app/assets/javascripts/hera/modules/inline_thread_selector.js
MattBudz: Good catch, I spun off a https://github.com/dradis/dradis-ce/pull/1534 for this change.

app/models/inline_comment_thread.rb
etdsoft: Yes, will do

etdsoft: https://github.com/dradis/dradis-ce/pull/1535

app/models/inline_thread.rb
etdsoft: a bit of BS from Claude if you ask me:

>   - updated_at can change for reasons that don't affect content — e.g. state changes, tag updates, touching associations. An inline thread anchored to text shouldn't become "outdated" because someone changed the issue's tags.
>   - PaperTrail versions track actual content edits — specifically event: 'update' versions. So version_id < latest_version.id means "the text has been edited since this thread was created," which is exactly what "outdated" means here (the anchored text may have moved or been deleted).

>   That said, the current implementation has a gap: it queries commentable.versions.where(event: 'update') which includes all update versions, not just content changes. If PaperTrail tracks non-content attribute changes (like state), the thread could be marked outdated even when the text hasn't changed. Worth considering whether to filter versions more tightly if that becomes an issue.

Using :versions isn't bulletproof, and neither is :updated_at. I don't feel strongly about this.

CLAUDE.md
```diff
@@ -0,0 +1,254 @@
...
+## Key patterns
+
+**Event publishing:** Controllers `include EventPublisher` and call `publish_event('namespace.action', ...)`. Models `include Eventable`. Events route through `ActiveSupport::Notifications` and `ActivityService.subscribe_namespace`.
+
+Event name convention: `'issue.created'`, `'comment.destroyed'`, `'inline_thread.resolved'`.
+
+> **Deprecated:** `ActivityTracking` (`track_created`, `track_destroyed`, `track_activity`,
+> `track_state_change`) — do not use in new code. Use `EventPublisher` + `publish_event` instead.
+
+**Engine registration:** Each Pro engine inherits from `Dradis::Plugins::Base`, provides `:addon`, and registers Warden strategies / controller extensions in initializers.
```
aapomm: Not just Pro engines

```suggestion
**Engine registration:** Each engine inherits from `Dradis::Plugins::Base`, provides `:addon`, and registers Warden strategies / controller extensions in initializers.
```

app/controllers/inline_threads/comments_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      )
+    else
+      head :unprocessable_entity
+    end
+  end
+
+  private
+
+  def project
+    @project ||= @inline_thread.commentable.respond_to?(:project) ? @inline_thread.commentable.project : nil
```
caitmich, line 31: Is the idea that this will be extended to commentables outside of projects in the future? Where would that be? Because currently `commentable` is always going to be related to a project

caitmich: Also, is this here because of the `Mentioned` inclusion? Mentioned can find mentionable users without needing a `project`

[Comment.mentionable_users](https://github.com/dradis/dradis-ce/blob/87588768e4d370b40b86002f2b5c297bb22eabb3/app/models/comment.rb#L85) already checks `resource.respond_to?(:project)`

app/controllers/inline_threads/comments_controller.rb
```diff
@@ -0,0 +1,35 @@
+class InlineThreads::CommentsController < AuthenticatedController
+  include EventPublisher
+  include Mentioned
```
caitmich, line 3: Do we need this? It doesn't appear to be working at the moment so if we're going to tackle this later we can skip the callback that this includes for now

app/controllers/inline_threads_controller.rb
```diff
@@ -0,0 +1,101 @@
+class InlineThreadsController < AuthenticatedController
+  load_and_authorize_resource
+
+  include EventPublisher
+  # FIXME: @mentions are currently too tied to Comments
```
caitmich, line 5: Can you make this comment clearer? I'm not sure what "half way through the integration" means

app/controllers/inline_threads_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+  before_action :authorize_commentable
+
+  def index
+    @inline_threads = commentable.inline_threads
+                           .includes(comments: :user)
+                           .order(created_at: :asc)
+  end
+
+  def show
+  end
```
caitmich, line 25:
```suggestion
  def show; end
```

app/controllers/inline_threads_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def authorize_commentable
+    if project
+      authorize! :use, project
+    else
+      authorize! :show, commentable
+    end
+  end
+
+  def commentable
```
caitmich, line 59: This is unnecessarily sending us in a circle

In the create action `load_and_authorize_resource` initializes `@inline_thread` with `inline_thread_params`

Then in the create action we're setting `@inline_thread = commentable.inline_threads.build(inline_thread_params)`
where `commentable = @inline_thread.commentable `

caitmich: What action is not already going to have `@inline_thread` set where it would hit the else block here? And if none then we also don't need `commentable_class`

UPDATE: I see now that it's just for the index action. Might be overkill but I'll leave it up to you

app/controllers/inline_threads_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def inline_thread_params
+    params.require(:inline_thread).permit(
+      :commentable_type, :commentable_id,
+      anchor: [:type, :exact, :prefix, :suffix, :field_name, { position: [:start, :end] }]
+    )
+  end
+
+  def project
+    @project ||= commentable.respond_to?(:project) ? commentable.project : nil
```
caitmich, line 97: Aren't all commentables going to respond to `:project`? Or is this also for the future in case we extend this to non project-related commentables? 

`Mentioned` here also isn't functional so instead of keeping this method, why not just directly use `authorize! :use, commentable.project` on L65 and remove this altogether

app/controllers/inline_threads_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+                           .includes(comments: :user)
+                           .order(created_at: :asc)
+  end
+
+  def show
+  end
+
+  def create
+    @inline_thread = commentable.inline_threads.build(inline_thread_params)
+    @inline_thread.user = current_user
```
caitmich: How about instead merging these into params in `inline_thread_params` to keep the action simpler/cleaner

Like what we do in templates/reports_controller in pro

app/controllers/inline_threads_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+    @inline_thread.version_id = commentable.versions.last&.id
+
+    if @inline_thread.save
+      if comment_params.present? && comment_params[:content].present?
```
caitmich: What about extracting this into a build_comment helper? Or a method on the `InlineThread` model. We'd already have commentable and could just pass user & content.

Currently this action is doing too much

caitmich: I realize it's an extra layer of validation but when are we ever going to create an inline thread without a comment?

app/controllers/inline_threads_controller.rb
```diff
@@ -0,0 +1,101 @@
+class InlineThreadsController < AuthenticatedController
+  load_and_authorize_resource
```
caitmich:
```suggestion
  load_and_authorize_resource, except: [:index]
```

app/models/inline_thread.rb
```diff
@@ -0,0 +1,109 @@
+class InlineThread < ApplicationRecord
+  include Eventable
+
+  REQUIRED_ANCHOR_KEYS = %w[exact position prefix suffix type].freeze
+
+  serialize :anchor, coder: JSON
+
+  enum :status, { open: 0, resolved: 1 }
```
caitmich: All other enums in CE and Pro are using arrays
```suggestion
  enum :status, [:open, :resolved]
```

app/models/inline_thread.rb
```diff
@@ -0,0 +1,109 @@
...
+  has_many :comments, dependent: :destroy
+
+  # Because Issue descends from Note but doesn't use STI, Rails's default
+  # polymorphic setter will set 'commentable_type' to 'Note' when you pass an
+  # Issue. Override the default behaviour here for issues:
+  #
+  # FIXME - ISSUE/NOTE INHERITANCE
+  def commentable=(new_commentable)
+    super
+    self.commentable_type = 'Issue' if new_commentable.is_a?(Issue)
```
caitmich, line 26: 🤮

app/models/inline_thread.rb
```diff
@@ -0,0 +1,109 @@
...
+
+    latest_version = commentable.versions.where(event: 'update').last
+    return false unless latest_version
+
+    version_id < latest_version.id
+  end
+
+  def project
+    commentable.project
+  end
```
caitmich: Why not `delegate :project, to: :commentable` instead?

app/models/inline_thread.rb
```diff
@@ -0,0 +1,109 @@
...
+      },
+      status: status
+    }
+  end
+
+  private
+
+  def coerce_anchor_position
+    return if anchor.blank? || !anchor.is_a?(Hash)
+    return unless anchor['position'].is_a?(Hash)
```
caitmich: `is_a?` seems like a smell. What's the problem this is fixing? Do we not have a predictable data shape?

app/models/inline_thread.rb
```diff
@@ -0,0 +1,109 @@
...
+    new_commentable
+  end
+
+  # -- Callbacks ------------------------------------------------------------
+  before_validation :coerce_anchor_position
+
+  # -- Validations ----------------------------------------------------------
+  validates :anchor, presence: true
+  validates :commentable, presence: true
+  validate :anchor_schema_valid
```
caitmich:
```suggestion
  validate :anchor_schema
```
redundant

app/models/inline_thread.rb
```diff
@@ -0,0 +1,109 @@
...
+    return if anchor.blank?
+
+    missing = REQUIRED_ANCHOR_KEYS - anchor.keys
+    if missing.any?
+      errors.add(:anchor, "missing required keys: #{missing.join(', ')}")
+    end
+
+    if anchor['position'].present?
+      pos = anchor['position']
+      unless pos.is_a?(Hash) &&
```
caitmich: When would position be a hash? Is this just in case `coerce_anchor_position` becomes unreliable?

app/models/inline_thread.rb
```diff
@@ -0,0 +1,109 @@
...
+        id: commentable.id,
+        title: commentable.title
+      },
+      status: status
+    }
+  end
+
+  private
+
+  def coerce_anchor_position
```
caitmich: Can we name this something more descriptive? It doesn't really tell us anything about what it does

app/views/inline_threads/activities/_inline_thread.html.erb
```diff
@@ -0,0 +1,11 @@
+<% if inline_thread %>
+  <% commentable = inline_thread.commentable %>
+  <% title = commentable.title.present? ? commentable.title : "#{commentable.class.name} ##{commentable.id}" %>
```
caitmich, line 3: We do this same check in a number of places across the code. Can we move this to `HasFields` or something?

app/views/inline_threads/_inline_thread.html.erb
```diff
@@ -0,0 +1,25 @@
+<div class="inline-thread" id="inline_thread_<%= inline_thread.id %>" data-behavior="inline-thread" data-thread-id="<%= inline_thread.id %>">
+  <%= render 'inline_threads/header', inline_thread: inline_thread %>
+  <blockquote>
+    <%= inline_thread.quoted_text %>
+  </blockquote>
+
+  <div class="border-top" id="thread_<%= inline_thread.id %>_comments">
+    <% inline_thread.comments.includes(:user).each do |comment| %>
```
caitmich: any reason for not setting this in the controller?

app/controllers/inline_threads/comments_controller.rb
etdsoft: I tried, and it's complicated, because other bits depend on it. Adding the full functionality is also not straightforward.

Had we been a little faster on the review we could have revisited and added this before merge, but I'm inclined to wait for now.

app/controllers/inline_threads/comments_controller.rb
etdsoft: > Because currently commentable is always going to be related to a project

Currently GW::Issues are commentable. Potentially RT Tickets too.

app/controllers/inline_threads_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+        @inline_thread.commentable
+      else
+        commentable_class.find(inline_thread_params[:commentable_id])
+      end
+    end
+  end
+
+  def commentable_class
+    if InlineCommentable.allowed_types.include?(inline_thread_params[:commentable_type])
+      inline_thread_params[:commentable_type].constantize
```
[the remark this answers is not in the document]
etdsoft: L80 ensures only known-good values reach L81

app/controllers/inline_threads_controller.rb
etdsoft: replied before, GW, RT, etc.

app/models/inline_thread.rb
```diff
@@ -0,0 +1,109 @@
...
+
+    %w[start end].each do |key|
+      val = anchor['position'][key]
+      if val.is_a?(String) && val =~ /\A\d+\z/
+        anchor['position'][key] = val.to_i
+      end
+    end
+  end
+
+  def anchor_schema_valid
```
etdsoft:
```suggestion
  def anchor_schema
```

app/views/inline_threads/activities/_inline_thread.html.erb
etdsoft:
```bash
in dradis-ce/ on comment_threads
› grep -r "commentable.title.present?" app
app/views/inline_threads/activities/_inline_thread.html.erb:  <% title = commentable.title.present? ? commentable.title : "#{commentable.class.name} ##{commentable.id}" %>

in dradis-ce/ on comment_threads
› grep -r "title.present?" app
app/views/inline_threads/activities/_inline_thread.html.erb:  <% title = commentable.title.present? ? commentable.title : "#{commentable.class.name} ##{commentable.id}" %>

in dradis-ce/ on comment_threads
```

I see the .title? check in other places (i.e. Issues activity), it's a good side quest.

app/controllers/inline_threads_controller.rb
etdsoft: the .user = current_user pattern is something we've leaned on before

```
› grep -r "user = current_user" app/controllers
```

Here we're adding :version_id as well. I think that 2 attributes is just enough to make this a bit irksome as you point out, but not enough for a full helper method extraction.

app/views/inline_threads/_inline_thread.html.erb
etdsoft: #index already did this, but #show didn't.

🧬: 1 🤖: 0

app/models/inline_thread.rb
etdsoft: extracted to InlineThread::Anchor, comment added, and code simplified.

app/assets/javascripts/hera/modules/inline_thread_highlighter.js
```diff
@@ -169,4 +179,20 @@ class InlineThreadHighlighter {
 
     return segments;
   }
+
+  // Find `needle` in `haystack` allowing whitespace runs in the needle
```
caitmich, line 183: 🤖? what is this analogy

app/controllers/inline_threads/comments_controller.rb
caitmich: GW::Issues respond to :project

app/controllers/inline_threads_controller.rb
caitmich: First comment still stands, `load_and_authorize_resource` should handle L28 (first line of the create action) as long as the params name is conventional

app/controllers/inline_threads/comments_controller.rb
```diff
@@ -18,6 +18,8 @@ def create
         notifiable: @comment,
         user: current_user
       )
+      @inline_thread.reload
```
caitmich, line 21: What's this for? Seems hacky

app/controllers/inline_threads_controller.rb
```diff
@@ -20,31 +21,20 @@ def index
...
-          publish_event('comment.created', @comment.to_event_payload)
-          broadcast_notifications(
-            action: :create,
-            notifiable: @comment,
-            user: current_user
-          )
-        end
-      end
-
+      comment = @inline_thread.comments.first
```
caitmich, line 31: This is brittle

app/controllers/inline_threads/comments_controller.rb
etdsoft: the TurboStream "full reload" you undid in your PR (your approach is better of course).

app/assets/javascripts/hera/modules/inline_thread_highlighter.js
etdsoft: 🤷‍♂️

This function looks for the "selected text" in the record, problem being - as Acceptance highlighted - that different browsers behave differently in terms of a) selection; b) how the selected text contains whitespace

app/controllers/inline_threads_controller.rb
etdsoft: this comment doesn't have enough context to be a) acted upon; b) replied.

In what way is it brittle?

PR #1528, opened 2026-03-02
Add dark mode

nicolachr:
### Summary

- Implements a runtime dark mode theme using CSS custom properties, switchable without a page reload.
- Promotes the semantic SCSS variable layer ($primary-bg, $text-default, etc.) to CSS custom properties defined in a new hera/themes.scss. Raw palette variables ($brand-*, $grey-*) remain compile-time SCSS constants.
- Adds a moon/sun toggle button in the utility nav. Theme preference is persisted server-side via a new PATCH /preferences endpoint backed by UserPreferences, and falls back to prefers-color-scheme when no preference is set.
- The main navigation and footer retain their original brand colours in both themes.
- Overrides Bootstrap 5.2's hardcoded form control, dropdown, and pagination styles via explicit selector rules inside [data-theme="dark"].

### Other Information

### Copyright assignment

file for the details.

### Check List

- [x] Added a CHANGELOG entry


app/models/user_preferences.rb
```diff
@@ -30,7 +31,7 @@ class UserPreferences
   #           seen. Since we can have multiple versions of each tour we need
   #           to be able to track what was the last version we presented to
   #           them. See TourRegistry class.
-  ATTRIBUTES = %i[digest_frequency tours]
+  ATTRIBUTES = %i[digest_frequency tours theme]
```
etdsoft: :alpha

app/javascript/controllers/theme_controller.js
```diff
@@ -0,0 +1,41 @@
...
+    this.darkIconTarget.classList.toggle("d-none", preference !== "dark")
+
+    this.autoItemTarget.classList.toggle("active", preference === "auto")
+    this.lightItemTarget.classList.toggle("active", preference === "light")
+    this.darkItemTarget.classList.toggle("active", preference === "dark")
+  }
+
+  persist(theme) {
+    const csrfToken = document.querySelector('meta[name="csrf-token"]')?.content
+    fetch("/preferences", {
```
aapomm: Let's not hardcode the paths in JS. Instead, we can use stimulus values here:

```suggestion
    fetch(this.preferencesPathValue, {
```

app/models/user.rb
```diff
@@ -1,7 +1,8 @@
 class User < ApplicationRecord
   alias_attribute :name, :email
 
-  serialize :preferences, type: UserPreferences
+  serialize :preferences, type: UserPreferences, yaml: { permitted_classes: [UserPreferences, Symbol] }
```
aapomm: Do we still need the `permitted_classes` here even though it's already specified in the `config/initializers/yaml_serializers.rb` file?

app/models/user_preferences.rb
```diff
@@ -55,7 +57,8 @@ def persisted?
   end
 
   def valid?(args)
-    DIGEST_FREQUENCIES.include?(self.digest_frequency)
+    DIGEST_FREQUENCIES.include?(self.digest_frequency) &&
+      (theme.nil? || VALID_THEMES.include?(theme))
```
aapomm: What about adding `nil` to the `VALID_THEMES` array? Then add a comment about supporting `nil` selected theme.

app/views/layouts/hera.html.erb
```diff
@@ -1,5 +1,7 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en"
+      data-theme="<%= current_user.preferences.theme || 'auto' %>"
```
aapomm: It seems we're calling this twice: `current_user.preferences.theme || 'auto'`. What about adding a method for this? Either a helper method or a method in the preferences model?

app/controllers/preferences_controller.rb
```diff
@@ -0,0 +1,15 @@
+class PreferencesController < AuthenticatedController
+  def update
+    theme = params.dig(:preferences, :theme)
+
+    unless theme.nil? || UserPreferences::VALID_THEMES.include?(theme)
```
aapomm: We're doing the same `theme.nil?` check again here. Maybe adding the `nil`to the possible `VALID_THEMES` is the way to go?

app/models/user.rb
nicolachr: Without permitted_classes in user.rb, the PATCH to save the theme returns a 500 on theme switch. This causes the theme to fall back to the stored value, and as a result it doesn't persist. Is there a better approach to this?

app/assets/stylesheets/hera/modules/_datatables.scss
```diff
@@ -259,12 +266,18 @@ div.dataTables_wrapper {
     flex: unset;
 
     .btn {
-      @include button($secondary-bg);
-      color: $text-default;
+      border: 0px none;
```
MattBudz:
```suggestion
      border: none;
```

app/assets/stylesheets/hera/modules/_datatables.scss
```diff
@@ -273,8 +286,14 @@ div.dataTables_wrapper {
     }
 
     .buttons-collection {
-      @include button($secondary-bg);
-      color: $text-default;
+      border: 0px none;
```
MattBudz:
```suggestion
      border: none;
```

app/views/layouts/application.html.erb
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" data-theme="light">
```
etdsoft, line 2: shouldn't we start with auto?

etdsoft: oh, pre-auth, got it

app/views/layouts/hera.html.erb
```diff
@@ -1,5 +1,9 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en"
+  data-theme="<%= current_user.preferences.theme_or_default %>"
+  data-controller="theme"
+  data-theme-preferences-path-value="<%= main_app.preferences_path %>">
```
etdsoft, line 5:
```suggestion
  data-preferences-path-value="<%= main_app.preferences_path %>">
```

app/views/layouts/application.html.erb
etdsoft: that being said, there's a failure of "default" here, the library should default to "light" if no theme is set on the HTML rather than forthing a layout updated to "avoid" the feature.

app/models/user_preferences.rb
```diff
@@ -17,6 +17,8 @@ class UserPreferences
   VALID_TOURS = %i[first_sign_in projects_show].freeze
   DIGEST_FREQUENCIES = %w[none instant daily].freeze
   DIGEST_FREQUENCY_DEFAULT = 'instant'.freeze
+  # nil is a valid theme value meaning "not set", which falls back to 'auto' in the layout.
+  VALID_THEMES = [nil, *%w[auto light dark]].freeze
```
etdsoft: this is a smell, `nil` shouldn't be in the list, and a default should be hard-coded elsewhere. Afaict `nil` only happens when this isn't set at all, after you make a choice, there's no way to go back, hence, it's a default not a valid value.

app/models/user_preferences.rb
```diff
@@ -30,15 +32,20 @@ class UserPreferences
   #           seen. Since we can have multiple versions of each tour we need
   #           to be able to track what was the last version we presented to
   #           them. See TourRegistry class.
-  ATTRIBUTES = %i[digest_frequency tours]
+  ATTRIBUTES = %i[digest_frequency theme tours]
   attr_accessor(*ATTRIBUTES)
 
+  def theme_or_default
```
etdsoft: nop if you want to memoize it with `||= 'auto'` that's cool, but this name is a smell

app/controllers/preferences_controller.rb
```diff
@@ -0,0 +1,15 @@
+class PreferencesController < AuthenticatedController
```
etdsoft, line 1: this needs a before action (:set_theme) and strong params

app/models/user_preferences.rb
etdsoft: :alpha

app/views/layouts/hera.html.erb
aapomm: This is actually needed because the stimulus controller is called `theme`.

app/models/user_preferences.rb
aapomm: Updated to use a helper method instead

app/views/layouts/hera.html.erb
etdsoft: 🎩🍽️

app/controllers/preferences_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  before_action :set_theme
+
+  def update
+    unless UserPreferences::VALID_THEMES.include?(@theme) || @theme.nil?
```
etdsoft: this belongs in #set_theme and fail if not

@theme.nil is an error condition here. We start `nil` and never go back to it.

app/models/user_preferences/theme.rb
```diff
@@ -0,0 +1,9 @@
+class UserPreferences
+  module Theme
+    VALID_THEMES = %w[auto dark light].freeze
+
+    def valid_theme?
+      VALID_THEMES.include?(theme) || theme.nil?
```
etdsoft: I don't think nil? is valid

app/models/user_preferences.rb
```diff
@@ -39,6 +40,7 @@ def to_yaml_properties
 
   def initialize(args = {})
     @digest_frequency = args[:digest_frequency] || DIGEST_FREQUENCY_DEFAULT
+    @theme = args[:theme]
```
etdsoft: here is where you `|| THEME_DEFAULT ` and forget about carrying `.nil?` across the codebase

app/views/layouts/setup.html.erb
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" data-theme="light">
```
etdsoft, line 2: not needed

app/views/layouts/application.html.erb
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" data-theme="<%= dradis_theme %>">
```
etdsoft: not needed

app/helpers/application_helper.rb
```diff
@@ -47,4 +47,13 @@ def spinner_tag(spinner_class: 'text-primary', align: 'center', inline: false)
       content_tag :div, nil, class: "spinner-border #{spinner_class}"
     end
   end
+
+  def dradis_theme
```
etdsoft: I see the work this is doing, but shouldn't it be the JS's task to look at the body attribute and read it or default?

app/javascript/controllers/theme_controller.js
```diff
@@ -0,0 +1,42 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["autoIcon", "lightIcon", "darkIcon", "autoItem", "lightItem", "darkItem"]
+  static values = { preferencesPath: String }
+
+  connect() {
+    this.preference = this.element.dataset.theme || "auto"
```
etdsoft, line 8: this already remove s the need from the helper, or the pre-auth layouts.

Change it to "light" and remove all the other stuff.

app/helpers/application_helper.rb
aapomm: Removed

app/javascript/controllers/theme_controller.js
aapomm: This stimulus controller is only used for changing and updating the theme when the dropdown is used. This isn't used for determining the theme set from the view. For that we leverage the CSS engine by setting the necessary attributes in the view.

PR #1535, opened 2026-03-05
Polymorphic InlineThreads and naming consistency across model/resource

etdsoft:
## Summary

   - Make `InlineThread` (formerly `InlineCommentThread`) use a polymorphic `belongs_to :commentable` instead of a direct `belongs_to :issue`, allowing threads on any commentable resource
   - Flatten controller/routes to mirror the `CommentsController` pattern: top-level resource resolving its parent from `commentable_type`/`commentable_id` params (moved out of `qa/` namespace)
   - Add `InlineCommentable` concern (mirrors `Commentable`) for registering allowed types
   - Rename model, table, migrations, and all references from `InlineCommentThread` to `InlineThread` for consistency with the controller/route naming

   ## Test plan

   - [ ] Run `bundle exec rspec spec/models/inline_thread_spec.rb spec/requests/inline_threads_spec.rb` — 24 examples, 0 failures
   - [ ] Verify inline threads work on QA issue show page (create, reply, resolve, reopen, delete)
   - [ ] Verify activity feed renders inline thread events correctly


app/controllers/inline_threads/comments_controller.rb
```diff
@@ -28,12 +26,12 @@ def create
 
   private
 
-  def set_issue
-    @issue = current_project.issues.ready_for_review.find(params[:issue_id])
+  def project
```
aapomm, line 28: Where is this used? Note that the `Mentioned` concern also defines a `project` method but relies on the controller's definition of `current_project`. If this method is needed, what about defining `current_project` instead?

app/controllers/inline_threads/comments_controller.rb
```diff
@@ -28,12 +26,12 @@ def create
...
+    @project ||= @thread.commentable.respond_to?(:project) ? @thread.commentable.project : nil
   end
 
   def set_thread
-    @thread = @issue.inline_comment_threads.find(params[:inline_thread_id])
+    @thread = InlineThread.find(params[:inline_thread_id])
```
aapomm: Shouldn't we fetch this with `load_and_authorize_resource` to make sure we have the correct permission?

app/controllers/inline_threads/resolutions_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def destroy
+    @thread.reopen!(current_user)
+    publish_event('inline_thread.reopened', @thread.to_event_payload)
+  end
+
+  private
+
+  def project
```
aapomm: Where is this used?

app/controllers/inline_threads/resolutions_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    @project ||= @thread.commentable.respond_to?(:project) ? @thread.commentable.project : nil
+  end
+
+  def set_thread
```
aapomm: Ditto on `load_and_authorize_resource`

app/controllers/inline_threads_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+  def comment_thread_params
+    params.require(:inline_thread).permit(:commentable_type, :commentable_id)
+  end
+
+  def project
+    @project ||= commentable.respond_to?(:project) ? commentable.project : nil
+  end
+
+  def set_thread
+    @thread = InlineThread.find(params[:id])
```
aapomm: `load_and_authorize_resource`?

app/controllers/inline_threads_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+    else
+      raise 'Invalid commentable'
+    end
+  end
+
+  def comment_thread_params
+    params.require(:inline_thread).permit(:commentable_type, :commentable_id)
+  end
+
+  def project
```
aapomm: See comment: https://github.com/dradis/dradis-ce/pull/1535/changes#diff-f168605a2da73af3aa25f680686386ef0690b945db9e2c7e35ede911f319e9c7R29

app/controllers/inline_threads/comments_controller.rb
etdsoft: Had to look into it.

Nowhere today, but it's there in CommentsController in Pro to allow for GW::Issue Comments, I'm inclined to keep for future-proofing InlineThreads which I suspect will end up in GW too.

app/controllers/inline_threads/resolutions_controller.rb
etdsoft: nowhere, gone.

app/controllers/inline_threads_controller.rb
etdsoft: I think this one can stay

app/controllers/inline_threads_controller.rb
etdsoft: #facts #gone

app/controllers/inline_threads/resolutions_controller.rb
etdsoft: agreed

PR #1536, opened 2026-03-05
Simplify InlineThreadSelector to anchor against rendered content

etdsoft:
Instead of maintaining a client-side reimplementation of FieldParser to map rendered text positions back to raw DB text, read textContent directly from the DOM after rendering. This eliminates the field map machinery (_buildFieldMap, _strippedText, _strippedToRaw) and the fast/slow path branching in buildAnchor.

The key insight is that anchors don't need to reference raw text positions — prefix/suffix/exact are sufficient for re-locating highlights, and the position fields just need internal consistency.

On dradis:liquid-rendered, textContent is re-read so Liquid-resolved content is picked up automatically, fixing the mismatch between raw text (containing Liquid tags) and what the user actually sees.

findFieldName now uses DOM traversal (h5 headings) instead of regex matching against raw text with #[...]# markers.

Also fixes highlights not appearing for multi-line selections: findTextInNodes in the highlighter was concatenating text nodes' textContent, which doesn't include `\n` at block boundaries (`<br>`, `<h5>`, `<p>`).

Since anchor.exact comes from getSelection().toString() (which mirrors innerText behavior), the indexOf match failed. Now uses innerText as the combined string.

Quoted text in the thread panel now preserves newlines (white-space: pre-line) and the blockquote margin is consistent between the JS-built new-thread form and the server-rendered partial.

### Check List

- ~~[ ] Added a CHANGELOG entry~~

### DIFF
--- a/app/assets/javascripts/hera/modules/inline_thread_selector.js
+++ b/app/assets/javascripts/hera/modules/inline_thread_selector.js
@@ -19,7 +19,7 @@ class InlineThreadSelector {
     this.$container = $(container);
     this.$content = this.$container.find('[data-behavior~=content-textile]');
     this.coordinator = coordinator;
-    this.renderedText = this.$content[0].textContent;
+    this.renderedText = this.$content[0].innerText;
     this.pendingSelection = null;
 
     // Prevent double-binding
@@ -45,7 +45,7 @@ class InlineThreadSelector {
     // Liquid async rendering replaces innerHTML of content-textile,
     // destroying our appended button. Re-read text and re-append.
     this.$content.on('dradis:liquid-rendered', () => {
-      this.renderedText = this.$content[0].textContent;
+      this.renderedText = this.$content[0].innerText;
       this.appendButton();
     });
   }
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/hera/modules/inline_thread_selector.js
```diff
@@ -111,114 +110,49 @@ class InlineThreadSelector {
...
-    // Copy remaining text after last marker
-    for (let k = lastEnd; k < raw.length; k++) {
-      this._strippedToRaw.push(k);
-      this._strippedText += raw.charAt(k);
-    }
-  }
-
   buildAnchor(selectedText) {
+    const text = this.renderedText;
     let selection = selectedText.replace(/\r\n/g, '\n');
```
aapomm: Do we still need this normalization? I don't believe `/r/n` is preserved when selecting text from the DOM.

app/assets/javascripts/hera/modules/inline_thread_selector.js
```diff
@@ -111,114 +110,49 @@ class InlineThreadSelector {
...
       },
       field_name: fieldName
     };
   }
 
   findFieldName(position) {
-    const fieldRegex = /#\[(.+?)\]#/g;
-    let match;
+    const text = this.renderedText;
+    const headings = this.$content[0].querySelectorAll('h5');
```
aapomm, line 146: If I have the following field:
```
#[Description]#
This is a fake field:

h5. Test

p. hello world
```
and I select the "hello world" section, it will create a thread with "Test" as the field. Not sure what this means in practice, but I would expect the anchor to use the `Description` field.

Is it worth it to create a data-selector for the real Dradis field headers and use it here instead of selecting all `<h5>` elements in the content?

etdsoft: probably we can leave that to v2

PR #1538, opened 2026-03-06
Use select/combo box for prompt selection on Echo widget

MattBudz:
### Summary

Replaces the nav list + pre-loaded tab panes on the Echo prompt picker with a full-width select (turned into a combo box by existing JS) above a full-width turbo frame preview area. Selecting a prompt lazily loads its content via a new `preview` action rather than pre-rendering all prompts up front. The first prompt is auto-selected on connect to save the user a click. The "Echo this prompt" submit flow is unchanged.

Prompt and response content areas now use `white-space: break-spaces` so whitespace and line breaks render correctly.

### Check List

- [ ] Added a CHANGELOG entry


engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/prompts_controller.rb
```diff
@@ -6,15 +6,17 @@ class Projects::PromptsController < AuthenticatedController
     before_action :check_turbo_config, only: [:index]
     before_action :set_type
     before_action :set_record, except: [:create]
+    before_action :set_prompt, only: [:preview, :show]
```
caitmich, line 8: alpha

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/prompts_controller.rb
```diff
@@ -65,6 +67,10 @@ def set_record
       @record = current_project.send(@type.to_s.pluralize).find(record_params[:record])
     end
 
+    def set_prompt
+      @prompt_template = current_user.prompts.find(params[:id])
```
caitmich: the variable name should match the method name i.e `@prompt` like how we have it for set_record and set_type

engines/dradis-echo/app/views/dradis/plugins/echo/projects/prompts/preview.html.erb
```diff
@@ -0,0 +1,4 @@
+<turbo-frame id="echo-prompt-preview">
+  <%= render partial: 'dradis/plugins/echo/projects/prompts/prompt',
+             locals: { prompt: @prompt_template } %>
```
caitmich: The variable name should be consistent, like what I mentioned in the controller above. If we're passing a prompt, why is it a prompt_template? One or the other should change

engines/dradis-echo/config/routes.rb
```diff
@@ -5,7 +5,9 @@
     end
 
     resources :projects, only: [] do
-      resources :prompts, only: [:index, :show, :create], module: :projects
+      resources :prompts, only: [:index, :show, :create], module: :projects do
+        get :preview, on: :member
```
caitmich, line 9: The use of `preview` feels a little odd since it's different than our other uses of preview routes. It's kind of a show/edit hybrid. I don't feel too strongly but it could be confusing for future us

engines/dradis-echo/app/views/dradis/plugins/echo/projects/prompts/index.html.erb
```diff
@@ -7,29 +7,22 @@
...
+            <%= prompt.title %>
+          </option>
         <% end %>
-      </ul>
+      </select>
     </div>
 
-    <div class="col-9">
+    <div class="col-12">
       <p class="form-label">Selected Prompt</p>
```
caitmich, line 25: Might be worth adding a note that the user can tweak the prompt before submitting

PR #1540, opened 2026-03-09
Echo prompts: replace icon text field with predefined icon select

MattBudz:
### Summary

Replaces the free-text icon input on the prompt form with a select menu of 25 predefined Font Awesome 6 (free tier) icons, curated for the cybersecurity/AI reporting context (e.g. Investigation, Network, Forensics, Risk Scoring, Code Review).

Follows the same pattern as node icons: parallel `LABELS` and `ICONS` arrays on the model, collection builder in a dedicated helper. No database migration required — the icon is still stored as the FA class string. Also adds an `inclusion` validation on `Prompt#icon` to enforce the predefined set.

### Check List

- [ ] Added a CHANGELOG entry


engines/dradis-echo/app/models/dradis/plugins/echo/prompt.rb
```diff
@@ -4,6 +4,62 @@ class Prompt < ApplicationRecord
 
     SCOPES = [ :issue ].freeze
 
+    LABELS = [
```
caitmich: alpha

caitmich: It's not immediately clear that these are parallel arrays so something could change and inadvertently cause this to break (different array lengths, mismatched label & icons, etc). What about using a hash instead and fetching `.keys` and `.values` in the helper?

engines/dradis-echo/app/models/dradis/plugins/echo/prompt.rb
```diff
@@ -4,6 +4,62 @@ class Prompt < ApplicationRecord
...
+      'AI Generation',
+      'Alert',
```
caitmich: These clutter up the model quite a bit, what about moving them to the helper if they're only going to be used there?

etdsoft: extracted, like Node::Types

engines/dradis-echo/app/models/dradis/plugins/echo/prompt.rb
etdsoft: I don't disagree, with the extraction, these are contained in a single file. We can refactor this out - of feel free, but I think we can go ahead for now, as the concern offers enough isolation in the mean time.

PR #1543, opened 2026-03-11
Additional comment threads cleanup

caitmich:
### Summary

Some additional cleanup of the comment threads feature
- Move fuzzyIndexOf into a module so it can be reused rather than duplicated
- Append new comment to thread via turbo instead of re-rendering the entire frame

### Check List

### DIFF
--- a/app/controllers/inline_threads/comments_controller.rb
+++ b/app/controllers/inline_threads/comments_controller.rb
@@ -18,8 +18,6 @@ class InlineThreads::CommentsController < AuthenticatedController
         notifiable: @comment,
         user: current_user
       )
-      @inline_thread.reload
-      @inline_thread.comments.includes(:user).load
     else
       head :unprocessable_entity
     end
--- a/app/controllers/inline_threads_controller.rb
+++ b/app/controllers/inline_threads_controller.rb
@@ -26,9 +26,9 @@ class InlineThreadsController < AuthenticatedController
 
   def create
     @inline_thread = commentable.inline_threads.build(inline_thread_params_for_create)
+    comment = @inline_thread.comments.first
 
     if @inline_thread.save
-      comment = @inline_thread.comments.first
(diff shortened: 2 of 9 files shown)


app/controllers/inline_threads_controller.rb
```diff
@@ -26,9 +26,9 @@ def show
 
   def create
     @inline_thread = commentable.inline_threads.build(inline_thread_params_for_create)
+    comment = @inline_thread.comments.first
```
caitmich, line 29: avoids an additional db query if you set this before saving thanks to accepts_nested_attributes_for

app/views/inline_threads/comments/create.turbo_stream.erb
```diff
@@ -1,3 +1,7 @@
-<%= turbo_stream.update 'inline_thread_content' do %>
-  <%= render 'inline_threads/inline_thread', inline_thread: @inline_thread %>
+<%= turbo_stream.append "thread_#{@inline_thread.id}_comments" do %>
+  <%= render 'inline_threads/inline_comment', comment: @comment %>
+<% end %>
+
+<%= turbo_stream.replace "thread_#{@inline_thread.id}_reply_form" do %>
```
caitmich, line 5: append new comment & replace form to reset text input instead of updating the entire turbo frame

app/controllers/inline_threads/comments_controller.rb
```diff
@@ -18,8 +18,6 @@ def create
         notifiable: @comment,
         user: current_user
       )
-      @inline_thread.reload
-      @inline_thread.comments.includes(:user).load
```
etdsoft, line 22: have you tested this code? This was added for Turbo reload of the offcanvas after adding a comment

etdsoft: I see the change below

PR #1550, opened 2026-03-13
refactor inline thread JS modules

MattBudz:
### Summary

Refactors the inline thread JavaScript into a cleaner, more maintainable structure:

- Moves the four inline thread JS files (`highlighter`, `selector`, `turbo`, `utils`) into a single `inline_threads.js` module — the classes are tightly coupled and the split overstated their independence
- Eliminates imperative DOM building by replacing the client-side form with a server-rendered partial via a new `InlineThreadsController#new` action and Turbo Frame navigation
- Improves variable naming throughout
- Removes noisy comments, keeping only those that explain non-obvious behaviour

No functional changes.

### Check List


app/controllers/inline_threads_controller.rb
```diff
@@ -75,11 +81,15 @@ def commentable_class
...
-      anchor: [:type, :exact, :prefix, :suffix, :field_name, { position: [:start, :end] }],
+    permitted = params.require(:inline_thread).permit(
+      :commentable_type, :commentable_id, :anchor,
       comments_attributes: [:content]
     )
+    # anchor arrives as a JSON string (from new/create) and must be parsed
+    # back into a hash before being passed to the model.
+    return permitted.merge(anchor: JSON.parse(permitted[:anchor])) if permitted[:anchor]
+
+    permitted
```
aapomm: What about this instead so we don't need to call `return`

```suggestion
    permitted.merge!(anchor: JSON.parse(permitted[:anchor])) if permitted[:anchor]

    permitted
```

app/controllers/inline_threads_controller.rb
```diff
@@ -14,6 +14,12 @@ class InlineThreadsController < AuthenticatedController
   before_action :authorize_commentable
   before_action :require_comment, only: [:create]
 
+  def new
+    @commentable_type = inline_thread_params[:commentable_type]
```
aapomm, line 19: Do we need to validate these params?

PR #1573, opened 2026-04-27
render Liquid content in issue/evidence table columns asynchronously

MattBudz:
### Summary

render Liquid content in issue/evidence table columns asynchronously

### Other Information

Custom fields containing Liquid (e.g. `{{ project.name }}`) were showing raw markup in the Issues, Issue Evidence, and Node items tables. Each custom-field `<td>` now gets `data-behavior="liquid-async"` with a path pointing to the record's preview endpoint, and `liquid_async.js` resolves it after page load. A small spinner shows while the request is in flight. Sort values are updated post-render so DataTables sorts on the evaluated text.

The only new wiring beyond the view changes is adding `:preview` to the `issues/evidence` routes and controller (that controller previously only had `:index` and `:new`).

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/app/controllers/issues/evidence_controller.rb
+++ b/app/controllers/issues/evidence_controller.rb
@@ -2,13 +2,15 @@ class Issues::EvidenceController < AuthenticatedController
   include ActivityTracking
   include ContentFromTemplate
   include DynamicFieldNamesCacher
+  include LiquidEnabledResource
   include MultipleDestroy
   include ProjectScoped
 
-  before_action :set_issues, only: [:create_multiple, :index, :new]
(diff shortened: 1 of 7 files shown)


app/controllers/issues/evidence_controller.rb
```diff
@@ -114,6 +120,10 @@ def set_auto_save_key
     end
   end
 
+  def set_evidence
+    @evidence = @issue.evidence.find_by(id: params[:id])
```
caitmich, line 124: We need to guard this against a nil issue

PR #1583, opened 2026-04-30
render Liquid fields by default via thread-local context

MattBudz:
## Summary

This is a proof of concept for a render-by-default architecture where `issue.fields`, `issue.title`, and equivalent evidence/note accessors return Liquid-rendered values automatically in request contexts, with no view changes required.

Three moving parts, all in CE:

**1. `LiquidRenderContext` (`app/lib/liquid_render_context.rb`, new)**
Thread-local context holder. Controllers set a lazy evaluator (proc) before the view renders; it is cleared in an `ensure` block. The lazy proc defers `liquid_assigns` evaluation until the first field access, after all `before_actions` have run.

**2. `HasFields#dradis_has_fields_for`**
`raw_fields` contains the existing parser logic, unchanged. `fields` calls `raw_fields`, renders Liquid values when a context is present, and memoizes the result on the instance. `set_field` and `delete_field` now write through `raw_fields` so rendered output is never saved back to the database. `title` inherits rendering automatically since it delegates to `fields.fetch('Title', ...)`.

**3. `LiquidEnabledResource`**
One `around_action :with_liquid_render_context` added to the concern already included by `IssuesController`, `EvidenceController`, `NotesController`, and others. No per-controller changes needed.

**`DiffedContent`** updated throughout to call `raw_fields` instead of `fields` so sync/diff operations always work with raw Liquid templates and cannot write rendered output back to the database.

## Verified

- Issue with `{{ project.name }}` in Title: index table, show heading, breadcrumb, and page `<title>` all render the project name
- Edit form textarea shows the raw Liquid template (form binds to `:text`, not `fields`)
- Background/API contexts never set a context, so `fields` returns raw values, unchanged behavior

## Check List

- [ ] Added a CHANGELOG entry

### DIFF
--- a/app/controllers/concerns/liquid_enabled_resource.rb
+++ b/app/controllers/concerns/liquid_enabled_resource.rb
@@ -10,13 +10,6 @@ module LiquidEnabledResource
     @liquid_assigns ||= default_liquid_assigns.merge!(liquid_resource_assigns)
   end
 
-  def with_liquid_render_context
-    LiquidRenderContext.set(-> { liquid_assigns })
-    yield
-  ensure
-    LiquidRenderContext.clear
-  end
-
   # To be overwritten by each controller
   def liquid_resource_assigns
     {}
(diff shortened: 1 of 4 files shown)


caitmich:
https://github.com/dradis/dradis-ce/pull/1584

PR #1585, opened 2026-05-04
cache Liquid-rendered fields in the DB, populated lazily on first read

caitmich:
## Summary

- Adds `rendered_text` (notes/issues) and `rendered_content` (evidence) cache columns via migration
- `HasFields#fields` renders lazily on first read: returns cached content if present, otherwise renders through Liquid using the record's project context, writes the result to the cache, and returns it
- `raw_fields` always returns unrendered values; used by `set_field`, `delete_field`, and cache population
- Cache is cleared on `before_save` and `after_touch` on Note and Evidence; Issue cascades to evidence via `update_all` to avoid touch callback chain
- `IssuesController#set_issues` includes `rendered_text` and `node_id` in its limited select to support lazy rendering on cache miss

## Test plan

- Visit issues index — verify Liquid variables render correctly on first load
- Edit an issue — verify rendered content updates after save
- Visit a node's evidence — verify Liquid variables render in evidence fields
- Update an issue — verify evidence cached renders are cleared and re-rendered on next read

### DIFF
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -77,6 +77,11 @@ class Issue < Note
     self.category = Category.issue unless self.category
   end
 
+  # When an issue changes, evidence that references it via Liquid is stale.
+  # update_all clears the cache in one query without firing callbacks, avoiding
+  # a cascade through evidence's belongs_to :node, touch: true.
+  after_touch { evidence.update_all(rendered_content: nil) }
+
   # -- Validations ----------------------------------------------------------
 
   # -- Scopes ---------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


app/models/issue.rb
```diff
@@ -77,6 +77,11 @@ def subscriptions(*params)
     self.category = Category.issue unless self.category
   end
 
+  # When an issue changes, evidence that references it via Liquid is stale.
+  # update_all clears the cache in one query without firing callbacks, avoiding
+  # a cascade through evidence's belongs_to :node, touch: true.
+  after_touch { evidence.update_all(rendered_content: nil) }
```
caitmich, line 83: Need to use `update_all` to avoid a callback cascade with every evidence's nodes & project

PR #1595, opened 2026-05-14
scaffold provider and Roslin agent architecture for LLM strategy

caitmich:
## Summary

Introduces a three-layer architecture inside the `dradis-echo` engine that makes AI-powered features provider-agnostic — swapping the underlying LLM is a configuration change, not a code change.

### Layer 1 — Provider (provider layer)

ActiveRecord STI models, one DB record per configured provider instance. Each subclass (`Provider::Ollama`, `Provider::Anthropic`, `Provider::OpenAI`) owns its own settings, configuration UI, and `generate(prompt:, model: nil, &block)` implementation. Key design points:

- Multiple provider records per type are supported (e.g. two Ollama instances with different models)
- Provider subclasses self-register into `Provider::ALLOWED_TYPES` via an `inherited` hook — no hardcoded list
- `Provider::Anthropic` and `Provider::OpenAI` raise `NotImplementedError` for now; implementations come in a follow-up PR
- Provider records are managed via a dedicated `ProvidersController` (CRUD), surfaced on the configuration page

### Layer 2 — Roslin (agent layer)

Plain Ruby classes that represent distinct AI-powered capabilities. Currently:

- `Roslin::IssueInteraction` — runs LLM prompts against issues and streams responses in real time via Turbo Streams. Resolves which `Provider` to use at call time via a configurable `provider_id`, then calls `provider.generate(prompt:, model:)`. Has no knowledge of which provider backs the call.
- `Roslin::LanguageTool` — grammar and style checking via a self-hosted LanguageTool instance. Not LLM-backed; owns its own settings independently.

Each component has its own enabled toggle and optional model/provider override. The `Roslin` class holds a global enabled toggle that gates all components.

### Layer 3 — Configuration

A single `ConfigurationForm` backed by `ActiveModel` where each agent (`Roslin`) declares its own `permitted_params`, `load_configuration`, and `save_configuration`. Agent sub-forms (`Roslin::ConfigurationForm`) handle their own components. The configuration page renders a single form with a single Save button.

```
Browser → Roslin::Projects::InteractionsController
        → Roslin::IssueInteractionJob.perform_later
        → Roslin::IssueInteraction.provider.generate(prompt:, model:)
        → provider-specific generate implementation
        → Turbo::StreamsChannel → Browser   # streamed chunks
```

## Test plan

**Provider CRUD**

1. Visit `/addons/echo/providers`. Confirm "New Provider" is a dropdown listing Ollama, Anthropic, Gemini, and OpenAI.
2. Select Ollama. Fill in a name, a valid address (`http://localhost:11434`), and a model. Click "Add Provider". Confirm the success flash and the provider in the list.
3. Select Anthropic. Fill in a name and a valid API key. Click "Add Provider". Confirm it is listed.
4. Repeat for OpenAI and Gemini using valid API keys.
5. Try creating an Ollama provider with an invalid address (e.g. `not-a-url`). Confirm a validation error is shown.
6. Try creating any provider with a blank name. Confirm a validation error is shown.
7. Edit an existing provider and clear the API key field. Save. Confirm the existing key is preserved.
8. Delete a provider that is not in use. Confirm the success flash and it disappears from the list.
9. Assign a provider to Issue Interaction, then try to delete it. Confirm the delete is blocked with a user-friendly message.

**Configuration**

1. Visit `/addons/echo/configurations`.
2. Enable Roslin, assign a provider to Issue Interaction, and save. Confirm the settings persist on reload.

**Happy path — streaming**

1. Open an issue and run a prompt via the Echo panel. Confirm the response streams in real time and the spinner disappears on the first chunk.
2. Switch Issue Interaction to each remaining provider type. Confirm each streams correctly.

**Error handling**

1. Remove the provider assignment in configuration. Trigger a prompt. Confirm "No provider configured for Issue Interaction. Ask an admin to assign one in Echo Configuration." appears rather than a generic exception.
2. Configure a provider with an unreachable address. Trigger a prompt. Confirm a user-friendly error appears in the UI rather than a server error.

## Check List

- [x] Added a CHANGELOG entry


etdsoft (dismissed):
## Automated code review for PR #1595

### Summary

This PR introduces a well-structured three-layer provider/agent/configuration architecture for LLM integration. The main risks are: an XSS vulnerability in error message broadcasting via Turbo Streams, feature specs that will fail due to mismatched button text, and missing sad-path handling when a provider is deleted while still assigned to an agent.

### Findings

#### F1 [High|security] — XSS via unsanitized error message in Turbo Stream broadcast

**Location:** `engines/dradis-echo/app/jobs/dradis/plugins/echo/roslin/issue_interaction_job.rb:29`

**What:** The rescue block interpolates `e.message` directly into an HTML string and broadcasts it via Turbo Streams with the `html:` parameter, which inserts raw HTML into the DOM without escaping. The `HttpStreaming#parse_sse_response` concern raises errors that include the raw HTTP response body from the provider, which may contain HTML.

**Why:** Stored/reflected XSS (OWASP A7). An attacker who controls the provider endpoint (or a legitimate provider returning an HTML error page) can execute arbitrary JavaScript in the browser of any user who triggers a prompt.

**How to fix:** Escape the error message before embedding it in HTML:
```ruby
rescue => e
  escaped = ERB::Util.html_escape(e.message)
  msg = '<div class="alert alert-danger m-0">'
  msg << escaped
  msg << '</div>'
  Turbo::StreamsChannel.broadcast_update_to [interaction_id, 'prompts'], target: response_id, html: msg
```

---

#### F2 [Medium|tests] — Feature specs reference wrong button text

**Location:** `engines/dradis-echo/spec/features/dradis/plugins/echo/providers_spec.rb:18, 25, 35, 41`

**What:** Four specs call `click_button 'Save'`, but the form partial renders `f.submit 'Add Provider'` (new) and `f.submit 'Update Provider'` (edit). There is no button with text "Save" on either page.

**Why:** Capybara's `click_button` matches by visible text. Since no element matches "Save", all four specs raise `Capybara::ElementNotFound` and fail. This means the CRUD happy paths have no passing test coverage.

**How to fix:** Match the actual button text:
```ruby
# In create specs:
click_button 'Add Provider'

# In edit specs:
click_button 'Update Provider'
```

---

#### F3 [Medium|sad-path] — ActiveRecord::RecordNotFound when assigned provider is deleted

**Location:** `engines/dradis-echo/app/models/dradis/plugins/echo/roslin/issue_interaction.rb:13`

**What:** `IssueInteraction.provider` calls `Provider.find(settings.provider_id)` without handling the case where the provider no longer exists. The `ProvidersController#destroy` guards against deleting in-use providers, but the check is non-atomic and doesn't cover direct DB manipulation.

**Why:** Users who trigger an interaction after a provider is deleted or before one is configured see an unhelpful internal error ("Couldn't find Dradis::Plugins::Echo::Provider without an ID") instead of actionable guidance.

**How to fix:** Use `find_by` and raise a user-friendly error:
```ruby
def self.provider
  Provider.find_by(id: settings.provider_id) ||
    raise('No provider configured for Issue Interaction. Ask an admin to assign one in Echo Configuration.')
end
```

---

#### F4 [Medium|sad-path] — No open_timeout on HTTP connection

**Location:** `engines/dradis-echo/app/models/concerns/dradis/plugins/echo/provider/http_streaming.rb:51`

**What:** `parse_sse_response` sets `http.read_timeout = 120` but does not set `open_timeout`. Ruby's `Net::HTTP` defaults to 60 seconds for `open_timeout`. If a provider address is unreachable, the connection attempt blocks a Resque worker for 60 seconds.

**Why:** A single misconfigured provider address can tie up a Resque worker for over a minute per interaction attempt. If multiple users trigger prompts against the same broken provider, workers pile up and delay all project-scoped background jobs.

**How to fix:** Set an explicit connect timeout:
```ruby
http = Net::HTTP.new(uri.host, uri.port)
http.use_ssl = uri.scheme == 'https'
http.open_timeout = 10
http.read_timeout = READ_TIMEOUT
```

---

#### F5 [Low|hygiene] — PR description missing Check List section

**Location:** PR description

**What:** The PR description has "Summary" and "Test plan" sections but omits the "Check List" section required by `.github/pull_request_template.md`.

**Why:** The Check List section serves as a self-review gate and explicit self-attestation.

**How to fix:** Add the Check List section to the PR description with the required checkboxes.

---

### Process Note

Claude completed a full review of all 57 changed files. Codex's sandbox was unable to reach the GitHub API and reviewed a different branch instead; its output was discarded. Only Claude's findings are included in this report.

engines/dradis-echo/app/models/dradis/plugins/echo/roslin.rb
```diff
@@ -0,0 +1,21 @@
+module Dradis::Plugins::Echo
+  class Roslin
+    include Dradis::Plugins::Configurable
+    extend Agent
+
+    addon_settings :'echo-roslin'
+
+    def self.enabled?
+      settings.enabled
```
aapomm: I believe this needs to be typecasted

```
> Dradis::Plugins::Echo::Roslin.enabled?
  Configuration Exists? (9.7ms)  SELECT 1 AS one FROM "configurations" WHERE "configurations"."name" = ? LIMIT ?  [["name", "echo-roslin:enabled"], ["LIMIT", 1]]
  Configuration Load (0.2ms)  SELECT "configurations".* FROM "configurations" WHERE "configurations"."name" = ? LIMIT ?  [["name", "echo-roslin:enabled"], ["LIMIT", 1]]
=> "t"
```

According to Claude, this is due to the SQLite's boolean representation
```
settings.enabled returns the raw string "t" from the DB (SQLite's boolean representation). The && in enabled? returns the last truthy value, so the whole method returns "t" instead of true. The fix is to cast through ActiveModel::Type::Boolean.
```

engines/dradis-echo/app/views/dradis/plugins/echo/providers/_form.html.erb
```diff
@@ -0,0 +1,49 @@
+<%= form_with model: provider, scope: :provider,
```
aapomm, line 1: This form seems to be shared across all providers even though each can have different configurable values (eg. anthropic with max_tokens). Might not be needed now, I can see how we'll need it in the future as we add more providers or even add more configurability. In that case, each provider should define its own form.

engines/dradis-echo/app/controllers/dradis/plugins/echo/providers_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+  class ProvidersController < ApplicationController
+    before_action :admin_required, if: -> { defined?(Dradis::Pro) }
+    before_action :set_provider, only: [:edit, :update, :destroy]
+
+    def index
+      @providers = Provider.all.order(:type)
+    end
+
+    def new
+      type = Provider::ALLOWED_TYPES.find { |t| t == params[:type] } || 'Ollama'
```
aapomm: Can we DRY this up with callback method like `set_provider_type`

engines/dradis-echo/app/javascript/controllers/dradis/plugins/echo/agent_controller.js
```diff
@@ -0,0 +1,31 @@
...
+      if (toolFields) {
+        toolFields.querySelectorAll('input, select, textarea').forEach((el) => {
+          el.disabled = !enabled;
+        });
+      }
+    });
+  }
+
+  get agentSwitch() {
+    return this.element.querySelector('[data-agent-switch]');
```
aapomm: What about using stimulus target for this?

engines/dradis-echo/app/models/concerns/dradis/plugins/echo/provider/http_streaming.rb
```diff
@@ -0,0 +1,109 @@
...
+
+      request = Net::HTTP::Post.new(uri)
+      request['Content-Type'] = 'application/json'
+      headers.each { |k, v| request[k] = v }
+      request.body = JSON.generate(body)
+
+      buffer = +''
+
+      http.request(request) do |response|
+        raise "#{self.class.name} API error (#{response.code}): #{response.body}" unless response.is_a?(Net::HTTPSuccess)
```
aapomm: On error, the body isn't yet read at this point so it will always be `nil`.

```suggestion
        unless response.is_a?(Net::HTTPSuccess)
          error_body = +''
          response.read_body { |chunk| error_body << chunk }
          raise "#{self.class.name} API error (#{response.code}): #{error_body}"
        end
```

engines/dradis-echo/app/views/dradis/plugins/echo/providers/_form.html.erb
caitmich: That was originally the case but was changed in https://github.com/dradis/dradis-ce/pull/1604 since they all use the same db structure

etdsoft:
## Acceptance Test Results — PR #1595

**Branch:** `echo/scaffold-llm-strategy-foundation` | **Date:** 2026-05-22 | **Result:** 11/13 PASS, 0 FAIL, 2 SKIPPED

---

### Provider CRUD

| # | Step | Status | Notes |
|---|------|--------|-------|
| 1 | New Provider dropdown lists all 4 types | PASS | Ollama, Anthropic, Gemini, OpenAI all rendered |
| 2 | Create Ollama (name, address, model) | PASS | 302 → success flash "LocalOllamaTest added." |
| 3 | Create Anthropic (name + API key) | PASS | All required fields (name, address, model, api_key). api_key encrypted. |
| 4 | Create OpenAI + Gemini | PASS | Both with default addresses/models. All 4 in list. |
| 5 | Ollama with invalid address | PASS | 422. "Address must be a valid HTTP(S) URL" |
| 6 | Blank name | PASS | 422. "Name can't be blank" |
| 7 | Edit + clear API key → key preserved | PASS | Controller strips blank api_key from params. Confirmed at DB level. |
| 8 | Delete unused provider | PASS | Success flash, removed from list. |
| 9 | Delete in-use provider → blocked | PASS | "LocalOllamaTest is in use and cannot be deleted." Delete button disabled with tooltip. |

### Configuration

| # | Step | Status | Notes |
|---|------|--------|-------|
| 10 | Visit `/addons/echo/configurations` | PASS (correction) | ⚠️ This route doesn't exist. Actual route is `/addons/echo/agents`. Test plan error — see below. |
| 11 | Enable Roslin, assign provider, save. Confirm persistence | PASS | PATCH → provider, enabled, model_override all persisted. Confirmed on page reload and in DB. |

### Error Handling

| # | Step | Status | Notes |
|---|------|--------|-------|
| H1 | Disabled Roslin → error in UI | PASS (code-verified) | `InteractionJob` raises "Agent 'Roslin' is not enabled", caught by rescue block, broadcast as `alert-danger` HTML via Turbo Streams. |
| H2 | Unreachable address → user-friendly error | PASS | Set Ollama to `localhost:9999`. Error: "Could not connect to server". Rescue block catches it — no stack trace reaches UI. |

### Streaming (skipped)

| # | Step | Status |
|---|------|--------|
| 12 | Happy path streaming (Ollama) | SKIPPED — Ollama not running on host |
| 13 | Streaming with Anthropic/OpenAI/Gemini | SKIPPED — no external API keys |

Code-verified: `InteractionJob` removes spinner on first chunk via `broadcast_remove_to`, appends chunks via `broadcast_append_to`. All three external providers implement `HttpStreaming` concern with correct SSE parsing.

---

### New commits verified

**`a8916ae929`** — Model override placeholder/hint text:
- Placeholder: `"Uses provider's configured model"` ✓
- Hint: `"Leave blank to use the model configured on the provider (currently: llama3.2:latest)"` ✓

**`5e19185dbc`** — Turbo frame navigation fix:
- Confirmed in interactions HTML: `data-turbo-frame="_top"` on "Enable it" link ✓
- Without this fix, clicking the link inside `<turbo-frame id="echo_issue_N">` causes "Content missing". Fix forces full-page nav. Correct.

### Sidebar widget ✓

Issue show page renders `<turbo-frame id="echo_issue_1">` with correct data-path. The `if @issue` guard prevents widget from appearing on the issues list. ✓

---

### ⚠️ Test plan issue

**Step 10 references `/addons/echo/configurations`** — this route does not exist. The correct URL is `/addons/echo/agents`. Recommend updating the test plan.

---

### Failures

None.

PR #1598, opened 2026-05-14
add LanguageTool configuration to Roslin settings

caitmich:
### Spec
Security consultants write substantial prose in Dradis — findings, recommendations, executive summaries. Even experienced writers make grammar and style errors under deadline pressure, and these mistakes undermine client confidence in the report. Consultants currently have to copy-paste text into external tools (Grammarly, Word) to check grammar,  then paste back. This breaks flow, risks data leakage to third-party services, and is easy to skip.

**Proposed solution**
Integrate LanguageTool (open source, self-hosted) as a grammar and style backend. Surface suggestions inline in the Dradis editor as underlines, similar to how spelling checkers work in word processors.

Request flow: As a user types in any rich-text field, the editor debounces input and sends the plain-text content to a LanguageTool endpoint. Suggestions are returned as an array of matches (offset, length, message, replacements). The frontend maps these back onto the editor content and renders underlines. Clicking an underline shows a tooltip with the suggestion and a one-click "Accept" action.

Backend components:

- Roslin::LanguageTool configuration model — stores the host URL and enabled flag alongside the Echo agent settings
- LanguageToolService — calls the LanguageTool /v2/check API, parses the response, and formats matches with their replacements
- GrammarChecksController — lightweight proxy endpoint that forwards grammar check requests through LanguageToolService to avoid CORS issues
- GrammarReplacementsController — applies an accepted suggestion by replacing the flagged text in the stored content

Frontend components:

- grammar_controller.js (Stimulus) — manages debounce, request lifecycle, suggestion display, and applying accepted replacements
- GrammarHighlighter — renders <mark> elements on matched text in the rendered content, handles click events to open the suggestion panel, and persists dismissed suggestions in localStorage
- BaseHighlighter — shared base class extracted from GrammarHighlighter and InlineThreadHighlighter; provides text-node walking, segment-finding, mark-wrapping, and mark-removal logic
- Offcanvas panel rendered conditionally in the Issue show view when LanguageTool is enabled
- Wavy underline styling for grammar suggestion marks

### Testing steps
1. Start a LanguageTool Docker container (`docker run -d -p 8010:8010 erikvl87/languagetool`) and run `db:migrate` to seed the Roslin agent.
2. Go to Addons > Echo > Agents, enable Roslin, select a provider, and add `LANGUAGETOOL_ADDRESS=http://localhost:8010` as an environment variable.
3. Open an issue that contains a spelling or grammar mistake and confirm underlines appear on the show view.
4. Click an underlined word and confirm a popover appears with the error message and replacement suggestions.
5. Click a replacement and confirm the text is updated in place.
6. Reload the page and confirm the correction persists.
7. Open the same issue in the edit view, confirm underlines appear in the Textile preview pane, click one, and accept a replacement — confirm the textarea content is updated.
8. Dismiss a suggestion by clicking Dismiss in the popover and confirm the underline is removed and does not reappear on the next check.
9. Create an issue where the same word appears as plain text in one field and inside a URL in another field — confirm the underline appears on the plain text occurrence, not inside the URL.
10. Disable Roslin in the agent settings and confirm no grammar checking occurs.
11. Remove the `LANGUAGETOOL_ADDRESS` env variable and confirm the grammar widget is not shown.


etdsoft (changes requested):
## Summary

This PR adds LanguageTool grammar checking and LLM streaming to Echo. It includes significant security gaps (IDOR on project-scoped endpoints, unescaped HTML streaming), a runtime crash in the grammar replacement flow, missing error handling for timeouts, and zero test coverage for content-modifying backend logic. The IDOR and XSS issues block merge; the crash affects every user interaction.

## Findings

### F1 [High|security] engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_checks_controller.rb:12 -- IDOR: record loaded without project scoping or authorization

**What:** `commentable_class.find(params[:commentable_id])` performs an unscoped global find. Although the controller includes `ProjectScoped`, no authorization check verifies the record belongs to `current_project`. The same pattern appears in `GrammarReplacementsController` at line 12.

**Why:** Any authenticated user who knows a `commentable_id` can read full field content of issues in projects they don't belong to — the grammar-check response returns `exact` text fragments. `GrammarReplacementsController` is worse: it lets them **modify** content in other projects. The existing `InlineThreadsController` guards this with `load_and_authorize_resource` and `authorize! :use, project`.

**How to fix:** Scope the record lookup to `current_project` and add CanCanCan authorization:

```ruby
record = current_project.issues.find(params[:commentable_id])
authorize! :read, record   # :update for replacements
```

Or map allowed types to scoped associations (`current_project.issues.find(...)`, `current_project.notes.find(...)`), returning 404 when the record is not in the current project.

### F2 [High|correctness] engines/dradis-echo/app/javascript/controllers/dradis/plugins/echo/grammar_controller.js:134 -- clearHighlights() method does not exist

**What:** After a replacement is applied, `this.highlighter.clearHighlights()` is called. Neither `GrammarHighlighter` nor its parent `BaseHighlighter` defines this method. The only match is `_clearHighlights(behaviorName)`, which requires an argument. This throws `TypeError: this.highlighter.clearHighlights is not a function` every time a user accepts a suggestion.

**Why:** The refactor extracted `clearHighlights` into `BaseHighlighter._clearHighlights(behaviorName)` and renamed it with a `_` prefix and required parameter, but this call site wasn't updated. The replacement flow is broken on every invocation.

**How to fix:** Add a public wrapper to `GrammarHighlighter`:

```javascript
clearHighlights() {
  this._clearHighlights('grammar-suggestion-highlight');
}
```

### F3 [High|security] engines/dradis-echo/app/jobs/dradis/plugins/echo/interaction_job.rb:17 -- unescaped LLM chunks injected as HTML

**What:** `broadcast_append_to(..., content: chunk)` streams provider output directly into the Turbo target without escaping.

**Why:** LLM output is untrusted and can be influenced by issue content or prompts. Turbo stream content is inserted as DOM, so chunks containing HTML like event handlers become stored/reflected XSS. OWASP A03 Injection/XSS.

**How to fix:** Escape each chunk before broadcasting:

```ruby
Turbo::StreamsChannel.broadcast_append_to(
  [interaction_id, 'prompts'],
  target: response_id,
  content: ERB::Util.html_escape(chunk)
)
```

### F4 [Medium|sad-path] engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb:27 -- nil field_value crashes apply_replacement

**What:** `apply_replacement` does `fields[field_name]` where `field_name` comes from params. If the field doesn't exist (typo, race condition, tampered request), `field_value` is nil and `field_value[0, offset]` raises `NoMethodError`.

**Why:** Separate HTTP requests for check and replacement mean record content can change between them. This is a user-facing 500 error.

**How to fix:** Guard against nil and return a clear error:

```ruby
field_value = fields[field_name]
return head :unprocessable_entity unless field_value
```

Or send the expected `exact` text with the replacement request and verify it matches before modifying.

### F5 [Medium|sad-path] engines/dradis-echo/app/services/dradis/plugins/echo/language_tool_service.rb:18 -- no timeout on Net::HTTP.post_form

**What:** `Net::HTTP.post_form(uri, ...)` uses Ruby's default 60s timeouts. A slow LanguageTool server will block the request thread for up to 120 seconds.

**Why:** This runs synchronously in `GrammarChecksController#create` (user request). A hung server ties up a Puma thread per field. With multiple fields and concurrent users, the thread pool exhausts quickly.

**How to fix:** Use `Net::HTTP.start` with explicit timeouts:

```ruby
http = Net::HTTP.new(uri.host, uri.port)
http.open_timeout = 5
http.read_timeout = 10
http.write_timeout = 10
```

### F6 [Medium|sad-path] engines/dradis-echo/app/javascript/controllers/dradis/plugins/echo/grammar_controller.js:48 -- error response parsed as matches array

**What:** When the server returns 503 with `{ error: "..." }`, the fetch chain does `.then(r => r.json()).then(matches => this.highlighter.highlight(matches))`. The `highlight` method calls `matches.forEach(...)`, but `matches` is an object (not an array), so `forEach` is undefined.

**Why:** Before crashing, `highlight` calls `this._clearHighlights(...)`, which removes all existing grammar marks. On a LanguageTool outage, suggestions disappear with no user feedback.

**How to fix:** Check the response status before parsing:

```javascript
.then(r => {
  if (!r.ok) throw new Error(`Grammar check failed: ${r.status}`);
  return r.json();
})
```

### F7 [Medium|tests] PR:- -- no test coverage for new backend code

**What:** Two controllers, a service, and grammar-related models have zero spec files.

**Why:** The replacement controller modifies record content via string slicing with user-supplied offsets — exactly the kind of logic needing test coverage for boundary conditions (offset at 0, past end-of-field, empty replacement, multi-byte characters). The IDOR in F1 would be caught by specs asserting cross-project access is denied.

**How to fix:** At minimum, add request specs for both controllers: (1) happy-path create, (2) cross-project access denied, (3) invalid commentable_type returns 422, (4) missing field_name returns 422. Add a unit spec for `LanguageToolService` with stubbed HTTP responses covering success and timeout.

### F8 [Low|consistency] engines/dradis-echo/app/javascript/controllers/dradis/plugins/echo/grammar_controller.js:63 -- inline style attribute with px units

**What:** `style="width:30px;text-align:center;"` is set via `innerHTML` in suggestion panel markup.

**Why:** Project CLAUDE.md forbids inline `style` attributes and requires relative units (`rem`, `em`) instead of `px`.

**How to fix:** Add a CSS class in `_inline_threads.scss` with equivalent styles using relative units, and reference it in the JS.

### F9 [Low|hygiene] PR:- -- PR description missing Check List and CHANGELOG entry

**What:** The PR template's "Check List" section is not completed, and `CHANGELOG` is not in the changed file list.

**Why:** Project conventions require CHANGELOG entries for user-facing changes (grammar checking in the editor) and the PR template checklist to be filled in.

**How to fix:** Add a CHANGELOG entry (e.g. `Echo: add inline grammar checking via LanguageTool`) and include the completed Check List in the PR description.

## Notes

- The `_findTextInNodes` method locates matches by searching for the `exact` text in content, not using the LanguageTool offset. If identical text appears multiple times in a field, all occurrences highlight the first one. Pre-existing limitation, not introduced here.
- The `_inline_threads.scss` uses `$lavender-dark` — verify this SASS variable exists in `smacss/variables.scss` to avoid a compile failure.

etdsoft (changes requested):
## Automated peer review -- PR #1598

**Verdict:** request-changes
**Findings:** 2 High, 6 Medium, 3 Low

### What I checked

### Both reviewers agreed

(No findings were flagged by both agents with matching file, line, severity, and category.)

### Claude only

### F1 [High|correctness] engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb:30 -- apply_replacement silently loses headerless content

**What:** `apply_replacement` round-trips raw text through `FieldParser.source_to_fields` and `fields_hash_to_source`. If the record has leading content before the first `#[Field]#` marker (headerless text), the round-trip silently drops it because the parser treats the raw text as having only the marked fields.

**Why:** Dradis field syntax defines fields by `#[Name]#` markers. Content before the first marker is headerless and is not preserved when converting to/from the fields hash. When a user accepts a grammar suggestion on a record with headerless content, the suggestion is applied, the content is persisted, and the headerless text is gone permanently.

**How to fix:** Before applying the replacement, check whether the raw text has content before the first field marker. If it does, either surface a warning ("This record has leading text that would be lost") or preserve it by extracting it as an implicit "preamble" field before the round-trip.

### F2 [Medium|sad-path] engines/dradis-echo/app/javascript/controllers/dradis/plugins/echo/grammar_controller.js:185 -- replacement errors not validated before applying

**What:** `_applyReplacement` calls `r.json()` without checking `r.ok`. A server error (422/500) will result in undefined being written to the editor, or blank content will be displayed.

**Why:** The Rails endpoint returns 422 with `{ errors: [...] }` when persistence fails. In that case `data.raw` is undefined, and calling `_refreshContent(undefined)` will clear the editor or set it to an empty string.

**How to fix:** Check the response status before refreshing: `if (!r.ok) { console.error(...); return; }` before calling `_refreshContent(data.raw)`. Add a controller spec where the replacement endpoint returns 422 and assert the editor content is not cleared.

### F3 [Medium|sad-path] engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb:1 -- stale-offset race condition between check and replacement

**What:** The grammar check endpoint returns offsets at T1. The user reads and decides to accept a replacement. Between T1 and T2 (when the replacement is submitted), the record text may have changed (edited by another user, or the user edited the textarea further). The replacement offset from T1 targets the wrong position in T2's text, silently corrupting content.

**Why:** No optimistic locking, no version check, no re-validation of the offset before applying it. A concurrent edit can shift all offsets.

**How to fix:** Snapshot the text hash or version at check time and validate it at replacement time. If the text changed, return 409 Conflict and ask the user to re-check. Alternatively, require a fresh grammar check before allowing a replacement.

### F4 [Medium|tests] engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb:0 -- missing sad-path test coverage

**What:** No request specs exercise the bounds-check rejection (offset < 0, offset + length > field_value.length), the Roslin.enabled? guard, or the headerless-content data loss scenario.

**Why:** These are correctness/sad-path branches that are not tested. A regression in bounds validation or the enabled check would not be caught.

**How to fix:** Add RSpec request specs in `engines/dradis-echo/spec/requests/dradis/plugins/echo/projects/grammar_replacements_spec.rb` (already exists) covering: (1) offset out of bounds returns 422, (2) replacement when Roslin is disabled returns 503, (3) replacement on headerless content behavior is explicit (either warns or preserves it).

### F5 [Low|consistency] engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb:1 -- missing Roslin.enabled? guard

**What:** `GrammarChecksController#create` guards with `return head :service_unavailable unless Agents::Roslin.enabled?`, but `GrammarReplacementsController#create` has no such guard.

**Why:** Consistency and defensive programming. If Roslin is disabled, the replacement endpoint should not process requests (it would call LanguageTool, which may also fail).

**How to fix:** Add the same guard to the top of `GrammarReplacementsController#create`.

### F6 [Low|hygiene] CHANGELOG:0 -- missing entity prefix in entry

**What:** CHANGELOG entry reads `update grammar checks`. Project convention (per `.github/pull_request_template.md` and past entries) groups changes by entity: `Issues: ...`, `Echo: ...`, `Admin: ...`.

**Why:** CHANGELOG is user-facing. Readers need to quickly spot what changed in their area of interest. Entity prefixes aid scannability.

**How to fix:** Change the entry to `Echo: update grammar checks` or more specifically `Echo: improve grammar checking accuracy` (depending on the actual user-facing change).

### F7 [Low|style] app/assets/stylesheets/hera/base.scss:0 -- hardcoded px unit in text-underline-offset

**What:** New CSS rule uses `text-underline-offset: 3px` instead of a relative unit.

**Why:** Dradis convention (per CLAUDE.md) uses relative units (rem, em, ch) to maintain responsive design consistency.

**How to fix:** Use `text-underline-offset: 0.1875rem` or `0.25rem` (closest match to 3px at 16px base).

### Codex only

### F8 [High|correctness] engines/dradis-echo/app/models/dradis/plugins/echo/agent.rb:5 -- JSON env store not portable across databases

**What:** `Agent` declares `store :env, coder: JSON` while the migration creates `env` as `t.json :env, default: {}`.

**Why:** `store` is intended for serialized text columns; native JSON columns already have type serialization built-in. On top of that, MySQL does not reliably support JSON columns with a literal `{}` default, and Dradis supports MySQL per CLAUDE.md. This can break migration or model loading in supported deployments.

**How to fix:** Pick one representation. Either keep `store :env, coder: JSON` and change the migration to `t.text :env`, or keep `t.json :env` with no DB default and remove `store`, using a model-level default such as `attribute :env, default: -> { {} }` plus direct hash access or `store_accessor` for known keys.

### F9 [Medium|sad-path] engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_checks_controller.rb:10 -- empty editor text falls back to persisted content

**What:** `params[:text].presence || persisted_content` treats an explicitly submitted empty string as absent. GrammarReplacementsController has the same pattern at line 15.

**Why:** In edit mode the browser sends unsaved textarea content. If the user clears the issue text, the server checks against the saved database value instead, so stale grammar suggestions can reappear and applying one can put old content back into the editor.

**How to fix:** Distinguish "param not supplied" from "supplied but empty": `raw_text = params.key?(:text) ? params[:text].to_s : persisted_content`. Add request specs where `text: ''` returns no matches and where replacement with `text: ''` returns 422 instead of using the persisted issue body.

### F10 [Medium|correctness] app/assets/javascripts/hera/modules/base_highlighter.js:50 -- duplicate grammar matches target the first occurrence

**What:** `_findTextInNodes` locates matches with `combined.indexOf(searchText)`, while `GrammarHighlighter` calls it with only `match.exact`.

**Why:** LanguageTool returns an `offset`, but the DOM highlighter ignores it. If the same typo appears twice in a field, both suggestions can be rendered on the first occurrence while clicking the second suggestion still sends the second offset to the replacement endpoint. That makes the visible highlight and the edited text disagree.

**How to fix:** Let grammar highlighting pass and use `match.offset` when mapping field text to rendered DOM text, or include enough context to locate the intended occurrence. Add a JS/unit or system spec with repeated text like `tset ... tset` and assert each suggestion highlights and replaces the matching occurrence.

### F11 [Medium|sad-path] engines/dradis-echo/app/javascript/controllers/dradis/plugins/echo/grammar_controller.js:185 -- replacement endpoint errors not validated before applying

**What:** `_applyReplacement` calls `r.json()` without checking `r.ok`, then always calls `_refreshContent(data.raw)`.

**Why:** The Rails endpoint returns 422 with `{ errors: [...] }` when persistence fails. In that case `data.raw` is undefined, and showing mode updates the content preview with an undefined/empty payload instead of leaving the current content intact and surfacing the failure.

**How to fix:** Check the response status before refreshing: throw or display an error unless `r.ok && data.raw`. Cover this with a controller/client spec or focused JS test where the replacement endpoint returns 422.

### Notes

Two High and six Medium findings across both agents indicate correctness and sad-path risks that block merge. The most critical are:
- Database portability of the JSON env store (Codex F8) -- this is a production blocker for MySQL deployments
- Data loss from headerless content round-trip (Claude F1) -- permanent data corruption on user acceptance
- Stale offsets between check and replacement (Claude F3) -- silent corruption on concurrent edits
- Empty text fallback to persisted content (Codex F9) -- UX gotcha where clearing a field still checks/updates the old text

Codex was unable to run `bundle install` or the test suite in the sandbox environment; Claude also could not execute the local specs. Both agents reviewed the diff and project conventions from the checked-out branch.

---

engines/dradis-echo/config/routes.rb
```diff
@@ -7,6 +7,9 @@
       resources :interactions, only: [:index, :show, :create], controller: 'projects/interactions' do
         get :preview, on: :member
       end
+
+      resource  :grammar_check,        only: [:create], controller: 'projects/grammar_checks'
```
caitmich:
```suggestion
      resource  :grammar_check, only: [:create], controller: 'projects/grammar_checks'
```

engines/dradis-echo/db/migrate/20260521000001_create_agents.rb
```diff
@@ -6,6 +6,9 @@ def change
       t.boolean :enabled, default: true, null: false
       t.string :model_override
       t.string :name, null: false
+
+      t.text :env
+
```
caitmich, line 11:
```suggestion
      t.text :env
```

engines/dradis-echo/db/migrate/20260521000001_create_agents.rb
```diff
@@ -25,9 +28,13 @@ def change
           name: 'Ollama'
         )
 
+        lt_address = config_value('roslin_language_tool_address') ||
```
caitmich: Nobody will have this value set in config so this check is useless as-is. We can just set it to the default address on first setup

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_checks_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      raw_text =
+        if params.key?(:text)
+          params[:text].to_s
+        elsif @record.respond_to?(:text)
+          @record.text
+        else
+          @record.content
+        end
+
+      fields   = FieldParser.source_to_fields(raw_text)
```
caitmich:
```suggestion
      fields = FieldParser.source_to_fields(raw_text)
```

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_checks_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      return head :service_unavailable unless Agents::Roslin.enabled?
+
+      raw_text =
+        if params.key?(:text)
+          params[:text].to_s
+        elsif @record.respond_to?(:text)
+          @record.text
+        else
+          @record.content
+        end
```
caitmich:
```suggestion
      raw_text = params.key?(:text) ? params[:text].to_s : @record.content
```

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_checks_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      fields   = FieldParser.source_to_fields(raw_text)
+
+      matches = LanguageToolService.new(
+        fields:  fields,
```
caitmich:
```suggestion
        fields: fields,
```

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_checks_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+        address: Dradis::Plugins::Echo::Agents::Roslin.instance.env['LANGUAGETOOL_ADDRESS']
```
caitmich:
```suggestion
        address: Agents::Roslin.instance.env['LANGUAGETOOL_ADDRESS']
```

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    before_action :set_record
+
+    def create
+      return head :service_unavailable unless Agents::Roslin.enabled?
+
+      field_name  = params[:field_name]
+      offset      = params[:offset].to_i
+      length      = params[:length].to_i
+      replacement = params[:replacement]
```
caitmich:
```suggestion
      field_name  = params[:field_name]
      offset = params[:offset].to_i
      length = params[:length].to_i
      replacement = params[:replacement]
```

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      return head :unprocessable_entity if replacement.nil?
+
+      raw_text =
+        if params.key?(:text)
```
caitmich: same as prev. controller

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+          params[:text].to_s
+        elsif @record.respond_to?(:text)
+          @record.text
+        else
+          @record.content
+        end
+
+      fields   = FieldParser.source_to_fields(raw_text)
```
caitmich:
```suggestion
      fields = FieldParser.source_to_fields(raw_text)
```

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_corrections_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+        end
+
+      fields   = FieldParser.source_to_fields(raw_text)
+
+      return head :unprocessable_entity unless fields[field_name]
+
+      field_value = fields[field_name]
+      return head :unprocessable_entity if offset < 0 || (offset + length) > field_value.length
+
+      if (exact = params[:exact])
```
caitmich: A bit of inconsistency between this line and L30-31. are we assigning inline as part of the conditional or doing it on a new line? we should choose one and use it consistently

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+      field_value = fields[field_name]
+      return head :unprocessable_entity if offset < 0 || (offset + length) > field_value.length
+
+      if (exact = params[:exact])
+        return head :conflict if field_value[offset, length] != exact
+      end
+
+      new_raw = apply_replacement(raw_text, field_name, offset, length, replacement)
+
+      if params[:persist] != 'false'
```
caitmich:
```suggestion
      unless params[:persist] == 'false'
```
more readable

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+        attr = @record.respond_to?(:text) ? :text : :content
```
caitmich: ditto with the `content` alias

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+
+      render json: { raw: new_raw }
+    end
+
+    private
+
+    def apply_replacement(raw_text, field_name, offset, length, replacement)
+      fields             = FieldParser.source_to_fields(raw_text)
+      field_value        = fields[field_name]
```
caitmich:
```suggestion
      fields = FieldParser.source_to_fields(raw_text)
      field_value = fields[field_name]
```

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_corrections_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+          return render json: { errors: @record.errors.full_messages }, status: :unprocessable_entity
+        end
+      end
+
+      render json: { raw: new_raw }
+    end
+
+    private
+
+    def apply_replacement(raw_text, field_name, offset, length, replacement)
```
caitmich: alpha?

engines/dradis-echo/app/controllers/dradis/plugins/echo/agents_controller.rb
```diff
@@ -28,7 +28,13 @@ def admin_required
     end
 
     def agent_params
-      params.require(:agent).permit(:enabled, :model_override, :provider_id)
+      permitted = params.require(:agent).permit(
+        :enabled, :model_override, :provider_id,
+        env_keys: [], env_values: []
+      )
+      keys   = permitted.delete(:env_keys) || []
```
caitmich:
```suggestion
      keys = permitted.delete(:env_keys) || []
```

engines/dradis-echo/app/javascript/controllers/dradis/plugins/echo/grammar_controller.js
```diff
@@ -0,0 +1,252 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static values  = {
+    commentableType:        String,
```
caitmich: spacing fixes needed in this file

engines/dradis-echo/app/assets/javascripts/dradis/plugins/echo/grammar_highlighter.js
etdsoft, line 1: This file belongs in the Echo engine.

app/assets/javascripts/hera/modules/liquid_async.js
```diff
@@ -17,6 +17,7 @@ document.addEventListener('turbo:load', function () {
       .then(function (html) {
         $(that).html(html);
         $(that).trigger('dradis:liquid-rendered');
+        that.dispatchEvent(new CustomEvent('dradis:liquid-rendered', { bubbles: true }));
```
etdsoft, line 20: Why is this needed and how is it different from L19?

app/assets/javascripts/hera.js
```diff
@@ -21,6 +21,8 @@
 //= require hera/modules/console_updater
 //= require hera/modules/differ
 //= require hera/modules/editor_toolbar
+//= require hera/modules/base_highlighter
+//= require hera/modules/grammar_highlighter
```
etdsoft: -> To Echo

app/assets/stylesheets/hera/modules/_inline_threads.scss
```diff
@@ -20,6 +20,15 @@
   }
 }
 
+// Grammar/style suggestion highlights
```
etdsoft: To Echo

engines/dradis-echo/app/assets/config/dradis/plugins/echo/manifests/hera.js
```diff
@@ -1,3 +1,4 @@
 //= link controllers/dradis/plugins/echo/agent_controller.js
+//= link controllers/dradis/plugins/echo/grammar_controller.js
```
etdsoft, line 4: -> Echo

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_checks_controller.rb
```diff
@@ -0,0 +1,36 @@
+module Dradis::Plugins::Echo
+  class Projects::GrammarChecksController < AuthenticatedController
+    include ProjectScoped
+
+    before_action :set_record
+
+    def create
+      return head :service_unavailable unless Agents::Roslin.enabled?
```
etdsoft: Insufficient, we need a separate LanguageTool check:
1. Configured?
2. Service online?

This check can be in Roslin (probably good), or in the LanguageToolService class if we want to break the Roslin encapsulation.

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_checks_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    private
+
+    def set_record
+      commentable_class = InlineCommentable.allowed_types
+                            .find { |t| t == params[:commentable_type] }
+                            &.constantize
+
+      return head :unprocessable_entity unless commentable_class
+
+      @record = current_project.send(commentable_class.model_name.plural).find(params[:commentable_id])
```
etdsoft: This pattern is a bit different from InlineThreads, is there a reason?

There we authorize the project and then use the commentable_class to find the instance

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb
```diff
@@ -0,0 +1,57 @@
+module Dradis::Plugins::Echo
+  class Projects::GrammarReplacementsController < AuthenticatedController
+    include ProjectScoped
+
+    before_action :set_record
+
+    def create
+      return head :service_unavailable unless Agents::Roslin.enabled?
```
etdsoft: brittle

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    private
+
+    def apply_replacement(raw_text, field_name, offset, length, replacement)
+      fields = FieldParser.source_to_fields(raw_text)
+      field_value = fields[field_name]
+      fields[field_name] = field_value[0, offset] + replacement + field_value[(offset + length)..]
+      FieldParser.fields_hash_to_source(fields)
+    end
+
+    def set_record
```
etdsoft: Should this be extracted / DRY?

engines/dradis-echo/app/services/dradis/plugins/echo/language_tool_service.rb
```diff
@@ -0,0 +1,39 @@
...
+        {
+          field_name: field_name,
+          offset: m['offset'],
+          length: m['length'],
+          message: m['message'],
+          exact: text[m['offset'], m['length']],
+          replacements: m['replacements'].map { |r| r['value'] }.first(3)
+        }
+      end
+    rescue StandardError => e
```
etdsoft, line 34: Are we sure this is working as expected? I removed the ENV (and never had LT running on the first place, and never saw any errors.

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_checks_controller.rb
aapomm: The main difference here is that this controller is using ProjectScoped while the InlineThreads isn't. I do see us expanding the grammar checks outside project context, but let's cross the bridge when we get there.

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_checks_controller.rb
aapomm: We can check for `configured?` but I'm not sure about making an extra request to check for `online?`. We're already inside the action making a request to the same endpoint, we might as well handle the response at that point (or lack thereof).

engines/dradis-echo/app/services/dradis/plugins/echo/language_tool_service.rb
aapomm: This works as expected. Note that when the UnavailableError is thrown, this is caught by the GrammarChecksController line 20 and responds with a 503. The logs will show `LanguageTool unavailable` as a well as an error in the browser console.

engines/dradis-echo/app/assets/config/dradis/plugins/echo/manifests/hera.js
aapomm: Already in echo

app/assets/javascripts/hera/modules/liquid_async.js
aapomm: Line 19 is for jquery listeners (`.on()`) while line 20 is for the native `addEventListener`. Ideally, we should only be using one of the two and migrate the instances of jquery to native, but it's out of scope for the task

etdsoft:
## Acceptance Test Results — PR #1598

**Branch:** `echo/add-languagetool` → `echo/llm-strategy`
**Date:** 2026-05-26
**Result:** 17 PASS, 0 FAIL, 3 SKIPPED

---

### Test Coverage

| # | Area | Step | Status |
|---|------|------|--------|
| 1 | Setup | Roslin agent created with `LANGUAGETOOL_ADDRESS` env key | ✅ PASS |
| 2 | API | `POST /grammar_suggestions` — no LT server → 503 with JSON error | ✅ PASS |
| 3 | API | `POST /grammar_corrections` — apply replacement, persists to DB | ✅ PASS |
| 4 | API | Corrections with `exact` param — match verification before replacing | ✅ PASS |
| 5 | API | Corrections with `exact` mismatch → 409 Conflict | ✅ PASS |
| 6 | API | Corrections with invalid `field_name` → 422 | ✅ PASS |
| 7 | API | Corrections with `persist=false` — dry-run, no DB change | ✅ PASS |
| 8 | API | Corrections with `offset+length > field_value.length` → 422 | ✅ PASS |
| 9 | API | Corrections with negative `offset` → 422 | ✅ PASS |
| 10 | API | Corrections with missing `replacement` → 422 | ✅ PASS |
| 11 | API | Corrections with invalid `commentable_type` → 422 | ✅ PASS |
| 12 | API | Both grammar endpoints return 503 when `language_tool_configured?` is false | ✅ PASS |
| 13 | UI | Echo agents index page renders; grammar_controller JS in import map | ✅ PASS |
| 14 | UI | Roslin agent edit form shows `LANGUAGETOOL_ADDRESS` env key/value fields | ✅ PASS |
| 15 | UI | Roslin agent env update via PUT — persists correctly | ✅ PASS |
| 16 | UI | Issue show page — grammar Stimulus controller data attributes correct | ✅ PASS |
| 17 | UI | Issue show page — Roslin widget renders (status icons + collapse div) | ✅ PASS |
| 18 | UI | Issue edit page — grammar Stimulus controller present with correct URLs | ✅ PASS |
| 19 | Static | `base_highlighter.js` loads before `inline_thread_highlighter.js` in manifest | ✅ PASS |
| 20 | Static | `GrammarHighlighter extends BaseHighlighter` | ✅ PASS |
| 21 | Static | `grammar.scss` — wavy underline via `text-decoration: underline wavy` | ✅ PASS |
| — | Skipped | Live LanguageTool response shape | ⏭ SKIPPED |
| — | Skipped | Grammar mark click → offcanvas panel (browser policy blocked) | ⏭ SKIPPED |
| — | Skipped | `text` param with unsaved content (curl multiline limitation) | ⏭ SKIPPED |

---

### 🔴 Finding: Seed data missing on fresh `db:prepare`

The `CreateAgents` migration seeds the Roslin agent inside a `reversible do |dir| dir.up` block. When `db:prepare` runs on a clean DB it uses `schema.rb` to load the schema directly — migration code is never executed. The Roslin agent record is not created.

**Impact:** On a fresh install, `Agents::Roslin.instance` raises `ActiveRecord::RecordNotFound`. Both grammar endpoints call `enabled?` which delegates to `instance`, so they'd 500 rather than 503. A first-time user would need to manually create the agent record (or navigate to the Echo admin page if that auto-creates it).

**Reproduction steps:**
1. Clean database, `db:prepare` with the `echo/add-languagetool` branch
2. `POST /addons/echo/projects/1/grammar_suggestions` → 500

**Recommendation:** Move the Roslin seed to `db/seeds.rb`, or add a `find_or_create` fallback in `Agents::Roslin.instance`, or rescue gracefully before the `find_by!` call.

---

### ℹ️ Skipped detail

- **Live LT response:** No LanguageTool container in this environment. Service call path and match format covered by `language_tool_service_spec.rb`.
- **JS behaviour:** Browser blocked by policy. Static review confirms correct class hierarchy (`GrammarHighlighter extends BaseHighlighter`), Stimulus controller data wiring, and CSS wavy underline.
- **Unsaved text via `text` param:** curl multiline limitation. Path directly tested in `grammar_corrections_spec.rb` (unsaved editor content test at lines 45–75).

PR #1599, opened 2026-05-15
add providers UI and Roslin configuration page

MattBudz:
### Summary

Adds a dedicated Providers page with full CRUD for LLM providers (Anthropic, OpenAI, Ollama), and a Roslin configuration page with nested tool cards for enabling agents and assigning providers per tool.

Key changes:
- Providers index with inline create, edit, and delete; provider brand icons and a type-switcher controller that reloads the form when the type changes
- Configuration page with agent-level and tool-level enable/disable toggles driven by a Stimulus controller; tool fields are disabled (and excluded from form submission) when the tool is off, so stored provider/model settings are preserved
- Validation surfaces specific error messages when Issue Interaction is enabled but no provider is selected

### Check List

- ~~Added a CHANGELOG entry~~


etdsoft (changes requested):
## Summary

Substantial UI refactor for provider management and Roslin configuration. The architecture is solid (provider abstraction, form nesting, Stimulus controllers), but there are three correctness issues that can crash at runtime: unimplemented providers crash the job, invalid STI types raise exceptions, and disabled components leave stale data that blocks deletions. Form validation also doesn't preserve user input on re-render. No tests cover the new endpoints, helpers, or error paths.

## Findings

### F1 [High|correctness] engines/dradis-echo/app/models/dradis/plugins/echo/provider/anthropic.rb:5 -- Anthropic and OpenAI providers crash Issue Interaction

**What:** `Provider::Anthropic#generate` and `Provider::OpenAI#generate` both raise `NotImplementedError`. These providers are listed in the provider form and can be selected for Issue Interaction. When a user submits a prompt, `IssueInteractionJob` calls `provider.generate`, hits the stub, and raises.

**Why:** `NotImplementedError` is a `ScriptError`, not a `StandardError`. The job's `rescue => e` catches `StandardError` and descendants, not `ScriptError`. The user sees a broken interaction with an unhandled exception instead of a graceful error message. The feature appears functional but fails at the point of use.

**How to fix:** Either implement `generate` for both providers before exposing them, or remove them from `Provider::ALLOWED_TYPES` and the provider form until they're ready. If you want a placeholder UI: add stubs that return a proper error message ("Anthropic not yet configured") instead of raising, or guard the form so these types can't be selected. Add a request/model spec that selects each provider for Issue Interaction, submits a prompt, and verifies the job either streams a response or returns a handled error.

### F2 [Medium|correctness] engines/dradis-echo/app/views/dradis/plugins/echo/configurations/_issue_interaction_provider_select.html.erb:7 -- provider select reads persisted value, not submitted form value

**What:** The select is hardcoded with `name="configuration_form[roslin][issue_interaction_provider_id]"` and determines the selected option by querying `Roslin::IssueInteraction.settings.provider_id` (the persisted database value), not by reading from the form object passed in.

**Why:** When the form fails validation (e.g., missing Issue Interaction provider) and re-renders, the user's newly submitted choice doesn't appear selected. The select snaps back to the previously saved provider ID. This breaks Rails' standard form round-trip: user submits → validation fails → form re-renders → form shows submitted data so user can correct other fields without re-entering this one.

**How to fix:** Pass the form builder into the partial and use `f.select`:
```erb
<%# in _roslin.html.erb %>
<%= render 'dradis/plugins/echo/configurations/issue_interaction_provider_select', f: cf %>

<%# in _issue_interaction_provider_select.html.erb %>
<%= f.select :issue_interaction_provider_id,
      Dradis::Plugins::Echo::Provider.all.order(:type).map { |p| [p.name, p.id] },
      { include_blank: 'Select a provider' },
      class: 'form-select' %>
```

### F3 [Medium|sad-path] engines/dradis-echo/app/controllers/dradis/plugins/echo/providers_controller.rb:20 -- invalid provider type raises ActiveRecord::SubclassNotFound

**What:** In `create`, when `attrs[:type]` is not in `Provider::ALLOWED_TYPES`, `type_short` is nil. The conditional is skipped, and the raw user-submitted `type` string (e.g., `"Kernel"`, `"FooBar"`) is passed to `Provider.new`. Rails STI calls `sti_class_for` on instantiation, which raises `ActiveRecord::SubclassNotFound` for unknown types.

**Why:** This is an unhandled bad-input path. A forged form submission, stale cached form data, or API client error can trigger a 500 instead of returning a validation error. The `new` action falls back to `Ollama` for invalid types, but `create` does not have the same guard.

**How to fix:** Reject or map invalid types before building the model:
```ruby
def create
  attrs = provider_params
  type_short = attrs[:type].presence_in(Provider::ALLOWED_TYPES)
  
  unless type_short
    flash.now[:alert] = 'Invalid provider type.'
    render :new, status: :unprocessable_entity
    return
  end
  
  @provider = Provider.new(attrs.merge(type: "Dradis::Plugins::Echo::Provider::#{type_short}"))
  # ...
end
```

Add a controller spec for `POST /providers` with an invalid or missing `type` param.

### F4 [Medium|sad-path] engines/dradis-echo/app/controllers/dradis/plugins/echo/providers_controller.rb:49 -- destroy doesn't check return value

**What:** `@provider.destroy` is called without checking the return value. The user always sees "removed" regardless of whether the operation succeeded. If a `before_destroy` callback halts the chain or a DB constraint prevents deletion, the user sees success but the provider still exists.

**Why:** Silent failure erodes trust. The user navigates back, finds the provider still present, and has to guess what went wrong. The code had the check before the refactor.

**How to fix:** Restore the return-value check:
```ruby
name = @provider.name
if @provider.destroy
  redirect_to providers_path, notice: "#{name} removed."
else
  redirect_to providers_path, alert: "Error removing #{name}: #{@provider.errors.full_messages.to_sentence}"
end
```

### F5 [Medium|correctness] engines/dradis-echo/app/models/dradis/plugins/echo/roslin/configuration_form.rb:35 -- disabled Issue Interaction doesn't clear provider_id, blocking deletion

**What:** When an admin disables Issue Interaction, `save_configuration` doesn't clear `settings.provider_id`. The setting persists. Later, when the admin tries to delete that provider, `provider_in_use?` still reports it as used (because `IssueInteraction.settings.provider_id` still matches), and deletion is blocked.

**Why:** An admin can't clean up: disable a feature, then delete its provider. The database state is inconsistent with the feature state (disabled but still configured), and there's no clear path to recovery. It's a usability trap.

**How to fix:** When Issue Interaction is disabled, clear its provider and model settings:
```ruby
def self.save_configuration(form)
  settings.enabled = form.issue_interaction_enabled
  if form.issue_interaction_enabled
    settings.model = form.issue_interaction_model
    settings.provider_id = form.issue_interaction_provider_id
  else
    settings.model = nil
    settings.provider_id = nil
  end
  settings.save
end
```

Or change `provider_used_by` to only count enabled components:
```ruby
def provider_used_by(provider)
  usages = []
  if IssueInteraction.enabled? && IssueInteraction.settings.provider_id.to_s == provider.id.to_s
    usages << 'Issue Interaction'
  end
  usages.join(', ')
end
```

Add a spec: disable Issue Interaction, then verify the previously selected provider can be deleted.

### F6 [Medium|tests] engines/dradis-echo/spec:0 -- no tests for new ProvidersController, ProvidersHelper, or error paths

**What:** The diff adds `ProvidersController` (index, new, create, edit, update, destroy), `ProvidersHelper` (provider_icon_path, provider_in_use?, provider_used_by), changed validation logic in `ConfigurationForm`, and new job/model error handling. Zero spec files are in the changeset.

**Why:** The sad-path issues (F1, F2, F3, F4, F5) would be caught by focused specs covering: selectable providers that don't implement `generate`, invalid type submissions to `create`, non-idempotent `destroy`, disabled components with stale settings. The controller's authorization guard, helper logic, and form validation changes are also untested.

**How to fix:** Add specs covering:
- `ProvidersController#create` with valid and invalid types
- `ProvidersController#destroy` when a provider is in use vs. free
- `ProvidersHelper#provider_in_use?` behavior
- `IssueInteractionJob` with providers that raise vs. stream responses
- Disabling Issue Interaction and verifying provider cleanup

### F7 [Low|consistency] engines/dradis-echo/app/views/dradis/plugins/echo/configurations/_provider.html.erb:0 -- dead partial left in place

**What:** The `configurations/_provider.html.erb` partial is still committed on the head branch, but the main `configurations/index.html.erb` no longer renders it. Providers are now managed via the new `providers/index.html.erb` page.

**Why:** Dead code that will confuse future developers who find it and wonder if it's still used.

**How to fix:** Delete `engines/dradis-echo/app/views/dradis/plugins/echo/configurations/_provider.html.erb`.

### F8 [Low|consistency] engines/dradis-echo/lib/dradis/plugins/echo/engine.rb:20 -- google.svg asset not precompiled

**What:** `google.svg` is committed and referenced in `ProvidersHelper::PROVIDER_ICONS` (`'Google' => 'dradis/plugins/echo/google.svg'`), but the engine's `assets.precompile` list only includes `anthropic.svg`, `ollama.svg`, and `openai.svg`.

**Why:** Currently harmless because no `Provider::Google` class exists. But when Google is implemented, `provider_icon_path` will return the Google SVG path, and `image_tag` will fail in production with a missing-asset error.

**How to fix:** Add `'dradis/plugins/echo/google.svg'` to the precompile array now, or remove `google.svg` and the `'Google'` entry from `PROVIDER_ICONS` until the provider is implemented.

### F9 [Low|hygiene] CHANGELOG:0 -- missing entry for user-facing UI changes

**What:** The PR adds new Providers page with full CRUD and redesigns the Echo Configuration page. The `CHANGELOG` file is not in the changed file list.

**Why:** User-facing changes require a changelog entry so customers know what's new in each release.

**How to fix:** Add an entry to `CHANGELOG`, e.g.: `Echo: add dedicated Providers page with full CRUD and redesign Roslin configuration UI`.

## Notes

- The Stimulus `agent_controller.js` correctly pairs with the server-side field-disabling logic — disabled fields aren't submitted, so the form won't save nil overwriting stored settings. Good design.
- The `provider_type_controller.js` reloads the page when type changes. This is pragmatic given different types have different field sets, but if UX feedback indicates user friction, consider keeping submitted fields intact across the reload.
- The configuration form's nested `fields_for` pattern is clean and extensible for future agents.

PR #1605, opened 2026-05-21
promote Agent to ActiveRecord and collapse Roslin/IssueInteraction

etdsoft:
### Summary

Promotes Agent from a plain Ruby module to an ActiveRecord model, collapses `Roslin::IssueInteraction` into it, and replaces the entire Configurations/ConfigurationForm infrastructure with standard CRUD.

**Why:** Roslin and IssueInteraction abused `Dradis::Plugins::Configurable` to store what should be model attributes (enabled, provider_id, model) as key-value pairs. The provider relationship had no referential integrity. The ConfigurationForm hierarchy (3 form classes, load/save bridging, an aggregator) was over-engineered for a single agent with a single capability.

**What changed:**

- **New `dradis_plugins_echo_agents` table** with `agent_type` (system/user enum), `enabled`, `model_override`, `name` (unique), and `provider_id` FK to providers. Migration reads old engine settings (`echo:roslin_*`) and seeds Roslin.
- **`Agent` AR model** with `belongs_to :provider`, system deletion protection, and `name` uniqueness validation.
- **`Agents::Roslin` convenience module** (`lib/`) delegates `enabled?`, `id`, `model_override`, `provider` to the AR instance — keeps callers clean.
- **`AgentsController`** with index/edit/update only (no create/destroy — one system agent for now).
- **`Provider`** gains `has_many :agents`, `#generate` contract (`NotImplementedError`), and simplified `#in_use?` → `agents.exists?`.
- **`InteractionJob`** (was `Roslin::IssueInteractionJob`) takes `agent_id:`, loads the Agent, calls `agent.provider.generate`.
- **`Projects::InteractionsController`** (was `Roslin::Projects::InteractionsController`) uses `Agents::Roslin.id` directly.
- **Deleted:** Roslin class, IssueInteraction, 3 ConfigurationForm classes, ConfigurationsController, all configuration views, and their specs.

### Other Information

- FK `to_table:` is required in the migration — `isolate_namespace` handles model table name inference but not migration FK resolution.
- 68 specs pass, rubocop clean.

### Check List

- [ ] Added a CHANGELOG entry


etdsoft:
**Automated Code Review**

## Summary

This PR refactors the Agent/Roslin/IssueInteraction infrastructure into a cleaner AR model design. The primary risk is a critical module delegation bug in `Agents::Roslin` that will crash the issue widget and interaction create action at runtime. Secondary concerns include missing controller specs, unguarded database lookups, and inconsistent blank-value handling in the Provider model.

## Findings

### F1 [High|correctness] engines/dradis-echo/lib/dradis/plugins/echo/agents/roslin.rb:11 -- delegate creates instance methods, not callable module methods

**What:** `delegate :enabled?, :id, :model_override, :provider, to: :instance` is called at the module body level. ActiveSupport's `Module#delegate` uses `module_eval` to define instance methods. These are only usable when the module is `include`d or `extend`ed into a class — they are not callable as `Agents::Roslin.enabled?` or `Agents::Roslin.id`.

**Why:** Three call sites rely on these as module-level calls and will raise `NoMethodError` at runtime:
- `_widget.html.erb:16` — `Dradis::Plugins::Echo::Agents::Roslin.enabled?` — crashes the issue show page for all users
- `interactions_controller.rb:27` — `Agents::Roslin.id` — crashes the interaction create action
- `agents_controller.rb:35` — `Agents::Roslin.id` — crashes the edit/update fallback (lower impact since `:id` is always present from routes)

No spec exercises these paths, so the 68 passing specs don't catch it.

**How to fix:** Add `extend self` to the module so instance methods are also available as module methods:

```ruby
module Dradis::Plugins::Echo
  module Agents
    module Roslin
      extend self

      def instance
        Agent.find_by!(name: 'Roslin')
      end

      delegate :enabled?, :id, :model_override, :provider, to: :instance
    end
  end
end
```

With `extend self`, `instance` changes from `def self.instance` to `def instance` (making it both an instance and module method). The delegated methods call `instance` on `self`, which resolves correctly when `self` is the module.

### F2 [Medium|sad-path] engines/dradis-echo/lib/dradis/plugins/echo/agents/roslin.rb:8 -- Agents::Roslin.instance crashes with RecordNotFound if Roslin agent is missing

**What:** `Agent.find_by!(name: 'Roslin')` raises `ActiveRecord::RecordNotFound` if the Roslin agent doesn't exist. This is called from the issue widget and the interaction create action. If the migration hasn't run, or the agent record was deleted, every issue page crashes.

**Why:** The migration seeds the Roslin agent, but there's no runtime guard. A `find_by!` in a convenience module used by views is fragile — a missing record takes down unrelated pages. The old `Roslin.enabled?` read from `Dradis::Plugins::Configurable` settings which defaulted gracefully.

**How to fix:** Use `find_by` (without bang) and default gracefully:

```ruby
def instance
  Agent.find_by(name: 'Roslin')
end

def enabled?
  instance&.enabled? || false
end
```

Or define `enabled?` explicitly instead of delegating it, so the widget degrades gracefully while other methods can still raise if the agent is truly expected to exist.

### F3 [Medium|tests] engines/dradis-echo/app/controllers/dradis/plugins/echo/agents_controller.rb:0 -- no controller or request specs for AgentsController

**What:** The new `AgentsController` (index/edit/update) has zero test coverage. No controller spec, no request spec, no feature spec.

**Why:** The controller has authorization logic (`admin_required` with the Pro guard), a `set_agent` fallback (`params[:id] ||= Agents::Roslin.id`), and strong params filtering. None of these are exercised by the model or job specs. The `admin_required` guard in particular is security-relevant — without a spec, a future refactor could silently remove it.

**How to fix:** Add a request spec covering at minimum: (1) unauthenticated access is rejected, (2) non-admin access is rejected in Pro mode, (3) update with valid params succeeds, (4) update with invalid params re-renders edit. Example structure:

```ruby
describe Dradis::Plugins::Echo::AgentsController do
  describe 'GET #index' do
    it 'lists all agents' do
      get echo.agents_path
      expect(response).to have_http_status(:ok)
    end
  end

  describe 'PATCH #update' do
    it 'updates the agent' do
      agent = Dradis::Plugins::Echo::Agent.find_by!(name: 'Roslin')
      patch echo.agent_path(agent), params: { agent: { enabled: false } }
      expect(agent.reload).not_to be_enabled
    end
  end
end
```

### F4 [Medium|correctness] engines/dradis-echo/app/models/dradis/plugins/echo/provider/ollama.rb:11 -- blank model override disables provider default

**What:** `resolved_model = model || self.model` treats an empty string from `Agent#model_override` as the selected model. The agent form says "Leave blank to use the provider's default model," but submitting that blank field stores `''`, so Ollama receives an empty model instead of `provider.model`.

**Why:** This breaks the Roslin happy path after an admin edits the agent without setting a model override. HTTP providers use `model.presence || self.model`, so the bug is isolated to Ollama and easy to miss in tests.

**How to fix:** Match the HTTP provider behavior: `resolved_model = model.presence || self.model`. Consider normalizing `Agent#model_override` blanks to `nil` as well, and add a focused spec that calls Ollama `generate(prompt: ..., model: '')` and asserts the provider default model is used.

### F5 [Medium|correctness] engines/dradis-echo/app/models/dradis/plugins/echo/provider/open_ai.rb:24 -- base URLs without trailing slash build wrong endpoints

**What:** `URI.join(address, 'chat/completions')` drops the `/v1` path when `address` is saved as `https://api.openai.com/v1` instead of `https://api.openai.com/v1/`. Gemini has the same class of issue at `provider/gemini.rb:21`, where `https://generativelanguage.googleapis.com/v1beta/models` becomes `.../modelsgemini-...`.

**Why:** The validation accepts both forms, and users naturally omit trailing slashes in "API Base URL" fields. The saved provider then looks valid but all requests fail against the wrong endpoint.

**How to fix:** Normalize `address` before validation/save with a trailing slash where subclasses require a base path, or build URLs with `#{address.chomp('/')}/chat/completions` and `#{address.chomp('/')}/#{model}:streamGenerateContent?alt=sse`. Add specs for OpenAI and Gemini custom addresses without trailing slashes.

### F6 [Low|hygiene] PR:- -- CHANGELOG entry unchecked

**What:** The PR checklist shows `[ ] Added a CHANGELOG entry` unchecked. The PR replaces the user-visible Configuration page with a new Agents page — this is a UI change.

**Why:** Per project conventions, user-facing changes require a CHANGELOG entry. However, the base branch is `echo/scaffold-llm-strategy-foundation` (a feature branch), not `develop`, so the changelog entry may be planned for the final merge to develop.

**How to fix:** If this will merge directly or the feature branch will carry the changelog: add an entry under the appropriate section, e.g. `Echo: replace Configuration page with per-agent management`. If the changelog will be added in the feature branch's PR to develop, no action needed — just confirm the intent.

## Notes

- **Migration data cleanup scope:** The migration deletes `Configuration.where('name LIKE ?', 'echo:roslin_%')`. The old `IssueInteraction` used `addon_settings :'echo-roslin-issue-interaction'`, which may store keys under a different prefix (e.g., `echo-roslin-issue-interaction:enabled`). If so, those records would be orphaned. Worth verifying the exact key format to confirm whether cleanup is complete.
- **DB query on every `Agents::Roslin` call:** Each delegated method calls `Agent.find_by!(name: 'Roslin')`, hitting the database. In the current single-agent setup this is negligible, but if the widget or other views call multiple delegated methods in the same request, consider memoizing or loading once per request.
- **Provider URL handling varies:** Codex noted that accepted UI inputs in the provider "base address" field can produce broken requests if trailing slashes are missing. This suggests normalization or validation is needed consistently across all provider types.

PR #1608, opened 2026-05-22
preserve Liquid tags when sanitizing drop field values

aapomm:
### Summary

Liquid operators like `<` and `>` inside `{% if %}` blocks were being stripped when issue and evidence content passed through `BaseDrop#escape()`. The HTML sanitizer (`HTML::Pipeline::SanitizationFilter`) parses its input as HTML, so `{% if cvss_score < 7.0 %}` — with `< 7.0` looking like an HTML tag — had the operator stripped, returning broken Liquid syntax to the template engine.

The fix stashes all Liquid tags (`{% %}` and `{{ }}`) behind safe alphanumeric placeholders before running the sanitizer, then restores them afterwards. The sanitizer only ever sees content outside Liquid tags, so operators are never mangled. This is the same protect-process-restore pattern `HTML::IgnoreLiquidInTextileBlockCodes` uses to shield Liquid from Textile processing.

XSS protection is maintained: HTML outside Liquid tags (e.g. `<script>`) is still sanitized normally.

Both `BaseDrop#escape()` (covers delegated methods like `title`, `text`) and `EscapedFields#fields` (covers field hash values on issues, notes, evidence) are updated.

### Testing steps

1. Log in as an admin or author user.
2. Navigate to a project and create a new Issue.
3. In a custom field (e.g. "Description"), enter content that contains a Liquid tag with a `<` operator and no surrounding spaces: `{% if cvss<7.0 %}Low risk{% endif %}`.
4. Save the issue.
5. Navigate to the issue show page and open the content preview (the rendered field output).
6. Assert the Liquid tag `{% if cvss<7.0 %}Low risk{% endif %}` is preserved in the rendered content — it should not be stripped to an empty string.
7. Edit the issue and set a custom field to content containing an XSS payload: `<script>alert(1)</script>`.
8. Save and view the issue.
9. Assert the script tag is stripped from the rendered output and no alert fires.
10. Edit the issue and set a custom field to content mixing Liquid tags and XSS: `{% if x < 10 %}<script>xss</script>{% endif %}`.
11. Save and view the issue.
12. Assert the Liquid tags are preserved (`{% if x < 10 %}{% endif %}`) and the script tag content between them is stripped.
13. Repeat the above steps using an Evidence record in place of an Issue to confirm the same behaviour applies to evidence field values.

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,7 +9,7 @@
   - Upgraded gems:
     - addressable, net-imap, nokogiri, rack
   - Bugs fixes:
-    - Drops: preserve Liquid tags when sanitizing issue and evidence field values
+    - Liquid: stop stripping comparison operators from Liquid tags in issue and evidence fields
     - Fields: show a visible border on dropdown fields in the editor
     - [entity]:
       - [future tense verb] [bug fix]
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,7 @@
   - Upgraded gems:
     - addressable, net-imap, nokogiri, rack
   - Bugs fixes:
+    - Drops: preserve Liquid tags when sanitizing issue and evidence field values
     - [entity]:
       - [future tense verb] [bug fix]
     - Bug tracker items:
--- a/app/drops/base_drop.rb
+++ b/app/drops/base_drop.rb
@@ -13,31 +13,21 @@ class BaseDrop < Liquid::Drop
     # call below and we want to skip this. Otherwise, proceed as normal.
     return if @wrapping
 
-    # Skip this if we're in the #escape method
-    return if method_name == :escape
-
     @wrapping = true
 
     original_method = instance_method(method_name)
 
     define_method(method_name) do |*args, &block|
       result = original_method.bind(self).call(*args, &block)
-      escape(result)
+      self.class.sanitize(result)
     end
 
     @wrapping = false
   end
 
-  private
+  def self.sanitize(obj)
+    return obj if obj.nil? || !obj.is_a?(String) || obj.empty?
 
-  def escape(obj)
-    if obj.nil? ||
-        !obj.is_a?(String) ||
-        obj.empty?
-
-      return obj
-    end
-
-    HTML::Pipeline::SanitizationFilter.call(obj).to_s
+    HTML::LiquidSafeSanitizer.call(obj)
   end
 end
--- a/app/drops/concerns/escaped_fields.rb
+++ b/app/drops/concerns/escaped_fields.rb
@@ -1,7 +1,7 @@
 module EscapedFields
   def fields
     @record.fields.transform_values do |value|
-      HTML::Pipeline::SanitizationFilter.call(value).to_s
+      BaseDrop.sanitize(value)
     end
   end
 end
--- a/lib/dradis/ce.rb
+++ b/lib/dradis/ce.rb
@@ -13,6 +13,7 @@ module Dradis
 end
 
 require 'html/ignore_liquid_in_textile_block_codes'
+require 'html/liquid_safe_sanitizer'
 require 'html/no_inline_code_textile_formatter'
 
 require 'html/pipeline/dradis/code_highlight_filter'
--- a/lib/html/liquid_safe_sanitizer.rb
+++ b/lib/html/liquid_safe_sanitizer.rb
@@ -1,0 +1,22 @@
+module HTML
+  class LiquidSafeSanitizer
+    LIQUID_PLACEHOLDER_FORMAT = "DRADISSAFE_#{SecureRandom.hex(8)}_%d"
+    LIQUID_TAGS_REGEX = /\{%.*?%\}|\{\{.*?\}\}/m.freeze
+
+    class << self
+      def call(str)
+        stash = []
+        text = str.gsub(LIQUID_TAGS_REGEX) do |match|
+          stash << match
+          format(LIQUID_PLACEHOLDER_FORMAT, stash.size - 1)
+        end
+
+        sanitized = HTML::Pipeline::SanitizationFilter.call(text).to_s
+
+        stash.each_with_index { |tag, i| sanitized.sub!(format(LIQUID_PLACEHOLDER_FORMAT, i), tag) }
+
+        sanitized
+      end
+    end
+  end
+end
--- a/lib/html/liquid_safe_sanitizer.rb
+++ b/lib/html/liquid_safe_sanitizer.rb
@@ -3,6 +3,11 @@ module HTML
     LIQUID_PLACEHOLDER_FORMAT = "DRADISSAFE_#{SecureRandom.hex(8)}_%d"
     LIQUID_TAGS_REGEX = /\{%.*?%\}|\{\{.*?\}\}/m.freeze
 
+    # Note: HTML inside Liquid delimiters (e.g. {% assign x = "<script>..." %}) is stashed
+    # verbatim and not sanitized at this layer. This is fine at the moment since the content reaches the
+    # browser only via the full rendering pipeline (ApplicationHelper#markup), which runs
+    # SanitizationFilter again after Liquid evaluation. Liquid's default auto-escaping also applies to {{ }} output.
+    # BUT do not add raw-output rendering paths without accounting for this otherwise you will introduce XSS vulnerabilities.
     class << self
       def call(str)
         stash = []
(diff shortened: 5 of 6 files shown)


etdsoft (dismissed):
## Summary

Solid fix for a real bug — Liquid comparison operators being mangled by the HTML sanitizer. The stash-restore pattern is the right approach and matches existing precedent in the codebase. Two concerns: (1) the restore step uses `sub!` with string replacement, which silently corrupts Liquid tags containing backslash sequences; (2) HTML embedded inside Liquid delimiters now bypasses drop-level sanitization — downstream pipeline sanitization mitigates this, but the trade-off should be explicitly acknowledged.

## Findings

### F1 [Medium|correctness] lib/html/liquid_safe_sanitizer.rb:16 -- sub! interprets backslash sequences in replacement string, corrupting Liquid tags

**What:** The restore step `sanitized.sub!(format(LIQUID_PLACEHOLDER_FORMAT, i), tag)` passes the stashed Liquid tag as the replacement string argument to `sub!`. Ruby's `String#sub!` interprets special sequences in string replacements: `\0`/`\&` become the matched text, `\1`–`\9` become empty (no capture groups), `\\` becomes `\`, and `` \` ``/`\'` become pre/post-match text.

**Why:** Security reports commonly contain regex patterns, file paths, and escape sequences. A Liquid tag like `{% assign pattern = "group\1" %}` would have `\1` silently stripped during restoration. `{% comment %}match: \& found{% endcomment %}` would have `\&` replaced with the placeholder string itself. This is silent content corruption — no error, no warning, wrong output.

**How to fix:** Use the block form of `sub!`, which treats the return value as a literal string with no special-sequence interpretation:

```ruby
stash.each_with_index do |tag, i|
  sanitized.sub!(format(LIQUID_PLACEHOLDER_FORMAT, i)) { tag }
end
```

### F2 [Medium|security] lib/html/liquid_safe_sanitizer.rb:4 -- HTML inside Liquid delimiters bypasses drop-level sanitization

**What:** `LIQUID_TAGS_REGEX` stashes the entire content between `{%`…`%}` and `{{`…`}}` delimiters. Any HTML inside those delimiters — e.g. `{% assign x = "<script>alert(1)</script>" %}` — is stashed unsanitized and restored verbatim. The drop-level sanitizer never sees it.

**Why:** The downstream rendering pipeline (`ApplicationHelper#markup`) runs `SanitizationFilter` after `LiquidFilter`, so the HTML would be caught there. But drop-level sanitization exists as defense-in-depth, and this PR weakens that layer. If any code path renders drop values without going through the full pipeline (API serialization, Turbo Stream fragments, JavaScript interpolation), the unsanitized HTML reaches the browser. The PR description states "XSS protection is maintained" but only addresses HTML *outside* Liquid tags, not inside.

**How to fix:** This may be an acceptable trade-off given that (a) all rendering paths appear to run through the pipeline, and (b) Dradis users are authenticated assessors. But the trade-off should be explicit:

1. Add a test documenting the expected behavior — e.g., `{% assign x = "" %}` returns the tag intact. This makes the contract visible to future developers.
2. Add a code comment on `LIQUID_TAGS_REGEX` noting that content inside Liquid delimiters is trusted and relies on downstream `SanitizationFilter` for HTML safety.
3. Confirm no rendering path bypasses the pipeline (API responses, Turbo Streams, `raw` output in templates).

### F3 [Low|hygiene] CHANGELOG:12 -- entry format doesn't match convention

**What:** The changelog entry reads `Drops: preserve Liquid tags when sanitizing issue and evidence field values`. The project convention for "Bugs Fixed" entries says: "do not start with 'preserve' — describe what the fix does, answering 'What will this upgrade do to my instance?'" The module prefix should match the user-facing entity, not the internal class.

**Why:** `CLAUDE.md` changelog conventions: entries lead with the module/entity, and bug fix entries describe the fix from the user's perspective.

**How to fix:** Something like: `Liquid: stop stripping comparison operators from Liquid tags in issue and evidence fields` — leads with the user-visible feature (Liquid), frames it as the bug behavior being fixed.

## Notes

- The `method_added` hook correctly no longer needs the `return if method_name == :escape` guard since `sanitize` is a class method and won't trigger the instance-method hook. No infinite recursion risk.
- The regex `LIQUID_TAGS_REGEX` uses non-greedy matching (`.*?`), so it handles multiple Liquid tags in a single string correctly. There's a known limitation: a `%}` or `}}` inside a Liquid string literal (e.g., `{% assign x = "hello %} world" %}`) would prematurely terminate the match. This is the same limitation as the existing `IgnoreLiquidInTextileBlockCodes` and would require a full Liquid parser to fix — not worth addressing here.
- The `SecureRandom.hex(8)` in the placeholder format is generated once at class load time. Collision probability is negligible, and the pattern matches the existing codebase precedent.
- Test coverage is good — covers XSS stripping, three variations of Liquid syntax (spaced operators, unspaced operators, assignments), mixed content, and both `BaseDrop` and `EscapedFields` paths.

---

*Codex sub-agent failed to analyze this PR due to sandbox network isolation and local checkout state mismatch. Review is based on Claude's analysis only.*

etdsoft:
## Acceptance Test Results — PR #1608

**Branch:** `liquid/protect-liquid-tags-in-drops` → `develop`
**Date:** 2026-05-26
**Result:** 26 PASS, 0 FAIL, 1 SKIPPED

---

### Test Coverage

| # | Area | Test | Status |
|---|------|------|--------|
| 1 | LiquidSafeSanitizer | `{% if cvss<7.0 %}` — `<` operator preserved | ✅ PASS |
| 2 | LiquidSafeSanitizer | `<script>alert(1)</script>` → `""` — XSS still stripped | ✅ PASS |
| 3 | LiquidSafeSanitizer | Mixed: `{% if x < 10 %}<script>xss</script>{% endif %}` → `{% if x < 10 %}{% endif %}` | ✅ PASS |
| 4 | LiquidSafeSanitizer | `{% if a > b %}yes{% endif %}` — `>` operator preserved | ✅ PASS |
| 5 | LiquidSafeSanitizer | `{{ issue.cvss_score }}` expression preserved | ✅ PASS |
| 6 | LiquidSafeSanitizer | Multiple Liquid tags in one string — all preserved | ✅ PASS |
| 7 | LiquidSafeSanitizer | `{% assign label = "critical" %}` preserved | ✅ PASS |
| 8 | LiquidSafeSanitizer | Multiline tag (`/m` flag) — preserved | ✅ PASS |
| 9 | BaseDrop.sanitize | Public class method returns correct output | ✅ PASS |
| 10 | BaseDrop (delegate) | XSS in field → `""` | ✅ PASS |
| 11 | BaseDrop (delegate) | Liquid `<` in field preserved | ✅ PASS |
| 12 | BaseDrop (delegate) | Mixed Liquid + XSS → Liquid preserved, XSS stripped | ✅ PASS |
| 13 | EscapedFields | Liquid `<` in `fields['Title']` preserved | ✅ PASS |
| 14 | EscapedFields | XSS in `fields['Title']` → `""` | ✅ PASS |
| 15 | EscapedFields | Mixed Liquid + XSS | ✅ PASS |
| 16–18 | Edge cases | nil, empty string, plain text — all handled correctly | ✅ PASS |
| 19 | Refactor | Private `escape` method removed from `BaseDrop` | ✅ PASS |
| 20 | Refactor | `BaseDrop.sanitize` is public class method | ✅ PASS |
| 21 | XSS attrs | `` — `onerror` stripped, img allowed | ✅ PASS |
| 22 | Placeholder | DRADISSAFE placeholder not present in output | ✅ PASS |
| 23–25 | Evidence | Liquid `<`, XSS, mixed — all behave correctly | ✅ PASS |
| 26 | UI | Issue created/saved with Liquid content, stored intact | ✅ PASS |
| — | SKIPPED | Liquid preview with variables in scope (browser policy blocked) | ⏭ SKIP |

---

### Testing Steps Verification

**Steps 3–6 (Issue with `{% if cvss<7.0 %}`):**


**Steps 7–9 (XSS):**


**Steps 10–12 (Mixed):**


**Steps 13+ (Evidence):** Same behaviour confirmed via `EvidenceDrop` with `EscapedFields` — all three scenarios pass.

---

### Note on HTML-inside-Liquid

The code comment notes that HTML inside Liquid delimiters (e.g. `{% assign x = "<script>..." %}`) is stashed verbatim and not sanitized at this layer. Verified: this is the correct behaviour and the comment accurately documents the risk and the mitigation (the `markup` helper runs sanitization again after Liquid evaluation).

PR #1609, opened 2026-05-25
update grammar and spell checking UI

nicolachr:
### Summary

Refines the grammar and spell checking UI introduced in the \`echo/add-languagetool\` branch:

- **Popovers**: removed header and blockquote styling; replaced primary buttons with \`btn-outline-lavender\`; added a footer with a dismiss link; set placement to \`auto\`; added dark mode support via CSS custom properties; popover closes on outside click
- **Highlights**: fixed grammar highlighting missing words in multi-paragraph fields by overriding \`_findTextInNodes\` in \`GrammarHighlighter\` to use \`textContent\` concatenation instead of \`innerText\`; updated underline colour to use \`var(--badge-lavender-color)\` CSS variable for theme awareness
- **Roslin sidebar widget**: adds a widget to the issues sidebar that prompts users to enable Roslin when disabled; when enabled, the widget is collapsed until errors are found — status icons (tick/cross) and issue count update live; collapses with "No grammar or spelling errors!" when all issues are resolved or dismissed; graceful degradation when the Roslin agent record is missing
- **Agent form**: updated "Add variable" button to \`btn-link\` style
- **Tests**: request specs covering the three widget rendering states (enabled, disabled, missing agent record)

### Testing steps

1. With Roslin disabled: visit an issue show page, confirm the widget is expanded with an "Enable Roslin" link and an info icon in the header
2. With Roslin enabled and no grammar errors: confirm the widget is collapsed and the header shows a green tick
3. With Roslin enabled and grammar errors present: confirm the widget is open, shows "N issues found", and the header shows a red ×; after fixing/dismissing all issues the widget collapses and the × turns to a tick
4. Confirm grammar underlines use the correct colour in both light and dark mode
5. Confirm clicking a grammar highlight opens a popover; clicking outside closes it

### Check List

- ~[ ] Added a CHANGELOG entry~

### DIFF
--- a/app/assets/javascripts/hera/modules/grammar_highlighter.js
+++ b/app/assets/javascripts/hera/modules/grammar_highlighter.js
@@ -22,15 +22,11 @@ class GrammarHighlighter extends BaseHighlighter {
     });
   }
 
-  clearHighlights() {
-    this._clearHighlights('grammar-suggestion-highlight');
-  }
-
   dismiss(match) {
     this.dismissed.add(this._key(match));
     this._saveDismissed();
-    this.contentEl.querySelectorAll('[data-behavior~=grammar-suggestion-highlight]').forEach(mark => {
-      if (mark.dataset.matchKey === this._key(match)) this._removeMark(mark);
+    this.contentEl.querySelectorAll(`[data-match-key="${this._key(match)}"]`).forEach(mark => {
+      this._removeMark(mark);
     });
   }
 
@@ -89,7 +85,7 @@ class GrammarHighlighter extends BaseHighlighter {
   }
 
   _key(match) {
-    return `${match.field_name}:${match.exact}`;
+    return `${match.field_name}:${match.offset}:${match.exact}`;
   }
 
   _loadDismissed() {
--- a/app/assets/javascripts/hera/modules/grammar_highlighter.js
+++ b/app/assets/javascripts/hera/modules/grammar_highlighter.js
@@ -57,6 +57,37 @@ class GrammarHighlighter extends BaseHighlighter {
     return mark;
   }
 
+  // Uses textContent concatenation instead of innerText to avoid block-boundary \n mismatches.
+  _findTextInNodes(textNodes, searchText) {
+    let offset = 0;
+    const nodeMap = [];
+
+    for (const node of textNodes) {
+      const len = node.textContent.length;
+      nodeMap.push({ node, startIndex: offset, endIndex: offset + len });
+      offset += len;
+    }
+
+    const combined = nodeMap.map(e => e.node.textContent).join('');
+    const matchIndex = combined.indexOf(searchText);
+    if (matchIndex === -1) return [];
+
+    const matchEnd = matchIndex + searchText.length;
+    const segments = [];
+
+    for (const entry of nodeMap) {
+      if (entry.endIndex <= matchIndex) continue;
+      if (entry.startIndex >= matchEnd) break;
+      segments.push({
+        node:        entry.node,
+        startOffset: Math.max(matchIndex, entry.startIndex) - entry.startIndex,
+        endOffset:   Math.min(matchEnd, entry.endIndex) - entry.startIndex
+      });
+    }
+
+    return segments;
+  }
+
   _key(match) {
     return `${match.field_name}:${match.exact}`;
   }
--- a/app/assets/javascripts/hera/modules/grammar_highlighter.js
+++ b/app/assets/javascripts/hera/modules/grammar_highlighter.js
@@ -15,9 +15,10 @@ class GrammarHighlighter extends BaseHighlighter {
 
   highlight(matches) {
     this._clearHighlights('grammar-suggestion-highlight');
+    const nextFrom = new Map();
     matches.forEach(match => {
       if (!this.dismissed.has(this._key(match))) {
-        this._highlightMatch(match);
+        this._highlightMatch(match, nextFrom);
       }
     });
   }
@@ -30,11 +31,16 @@ class GrammarHighlighter extends BaseHighlighter {
     });
   }
 
-  _highlightMatch(match) {
+  _highlightMatch(match, nextFrom = new Map()) {
     if (!match.exact) return;
 
-    const segments = this._findTextInNodes(this._getTextNodes(), match.exact);
+    const fromIndex = nextFrom.get(match.exact) || 0;
+    const segments  = this._findTextInNodes(this._getTextNodes(), match.exact, fromIndex);
     if (segments.length === 0) return;
+
+    const combined = this.contentEl.innerText;
+    const matchPos = combined.indexOf(match.exact, fromIndex);
+    if (matchPos !== -1) nextFrom.set(match.exact, matchPos + match.exact.length);
 
     const marks = this._wrapSegments(segments, match);
     marks.forEach(mark => {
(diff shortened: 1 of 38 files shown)


etdsoft (dismissed):
## Automated peer review -- PR #1609

**Verdict:** request-changes
**Findings:** 1 High, 3 Medium, 1 Low

### What I checked

### Both reviewers agreed

(No findings were flagged by both agents with matching file, line, severity, and category.)

### Claude only

### F1 [Medium|sad-path] engines/dradis-echo/app/views/dradis/plugins/echo/issues/_widget.html.erb:1 -- Roslin.enabled? raises RecordNotFound if agent missing

**What:** The widget partial calls `Roslin.enabled?`, which invokes `Agent.find_by!(name: 'Roslin')`. If the Roslin agent record has been deleted or not yet created (e.g., fresh test database, or manual row cleanup), the bang method raises `ActiveRecord::RecordNotFound` and crashes the issues sidebar.

**Why:** The sidebar is rendered for every issue view. A missing record makes the entire interface unusable, not just the widget. The bang method should only be used when the caller is prepared to handle the exception.

**How to fix:** Use `find_by` (no bang) with a nil-safe fallback in `Agents::Roslin.instance`:

```ruby
def instance
  Agent.find_by(name: 'Roslin') or raise "Roslin agent not found"
end

# Then in the widget, rescue the error:
def roslin_enabled?
  Agents::Roslin.instance&.enabled? || false
end
```

Or better: pass `enabled_agent` into the widget as a local so the view doesn't call the query at render time.

### F2 [Medium|tests] engines/dradis-echo/app/views/dradis/plugins/echo/issues/_widget.html.erb:0 -- missing tests for the new sidebar widget

**What:** The widget partial has two render paths (Roslin enabled and disabled) and relies on the fix from F1, but no view/feature specs exercise it.

**Why:** The crash risk in F1 would not be caught. Also, any breaking change to how the sidebar renders would not be caught by the test suite.

**How to fix:** Add a feature spec in `spec/features/issues_spec.rb` or a new spec file covering: (1) sidebar renders when Roslin is enabled, (2) sidebar renders safely when Roslin is disabled, (3) sidebar does not render if the Roslin agent record is missing (graceful degradation, not a 500 error).

### F3 [Low|hygiene] CHANGELOG:- -- "Added" checkbox struck through

**What:** The CHANGELOG entry for this PR includes a struck-through "Added" checkbox: `~~Added~~`.

**Why:** The PR targets a feature branch, not `develop`. The checkbox indicates this is a work-in-progress state. It's not wrong, but it signals incompleteness to anyone reading the PR.

**How to fix:** Once the PR is ready to merge to `develop`, update the CHANGELOG to remove the strikethrough and finalize the entry.

### Codex only

### F4 [High|security] engines/dradis-echo/app/controllers/dradis/plugins/echo/providers_controller.rb:71 -- Provider URLs enable SSRF and credential exfiltration

**What:** `provider_params` permits `:address`, and the provider model only validates that it is an HTTP(S) URL via regex before `Provider::HttpStreaming` sends authenticated HTTP requests to that address.

**Why:** This is SSRF (OWASP A10 Server-Side Request Forgery). An admin user can configure a provider address pointing at internal services (localhost, private IPs) or an attacker-controlled host, then trigger Echo interactions to POST content there. For OpenAI, Anthropic, and Gemini providers, this also exfiltrates the configured API key to the wrong host.

**How to fix:** Add provider-specific host validation before saving or before request execution. Require OpenAI addresses to use `api.openai.com`, Anthropic to use `api.anthropic.com`, Gemini to use `generativelanguage.googleapis.com`, and forbid RFC1918 / link-local / localhost addresses. Treat self-hosted / custom endpoints as an explicit Ollama-only feature (which already accepts arbitrary hosts intentionally). Add request specs that reject private-range hosts and wrong cloud-provider hostnames.

### F5 [Medium|correctness] engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/grammar_replacements_controller.rb:42 -- Replacement rebuild can drop or corrupt issue content

**What:** `apply_replacement` parses the whole raw body with `FieldParser.source_to_fields`, mutates one field, then serializes back with `fields_hash_to_source`.

**Why:** `source_to_fields` treats the raw text as a set of `#[FieldName]#` blocks: any content before the first block (headerless text), duplicate field sections, and trailing whitespace not inside a field can be silently dropped or normalized away. A single grammar suggestion can therefore remove leading content, deduplicate field sections unintentionally, or alter unrelated fields.

**How to fix:** Splice the replacement into the original `raw_text` string at the exact byte range instead of rebuilding the document from a hash. Pass and verify the matched `exact` text server-side before applying the replacement, so the caller confirms the text hasn't changed between check and replacement. Add specs covering: (1) headerless content is preserved, (2) trailing newlines outside fields are preserved, (3) duplicate field sections are preserved, (4) replacement with changed document text returns error instead of corrupting.

### Notes

One High and three Medium findings indicate correctness and security risks that block merge. The SSRF in F4 is the most critical — it allows credential exfiltration and internal network access. F1 (RecordNotFound crash) is also a production issue but scoped to widget rendering. F5 (content corruption on replacement) is less likely to trigger but has permanent data loss as a consequence if it does.

Both Claude and Codex were unable to fetch GitHub API metadata or the official PR diff from within the sandbox environment. Both reviewed the local checked-out `echo/add-languagetool-design` branch and the project's CLAUDE.md conventions. Neither sub-agent could run the local test suite.

---

app/assets/javascripts/hera/modules/grammar_highlighter.js
```diff
@@ -59,6 +59,37 @@ class GrammarHighlighter extends BaseHighlighter {
     return mark;
   }
 
+  // Uses textContent concatenation instead of innerText to avoid block-boundary \n mismatches.
+  _findTextInNodes(textNodes, searchText) {
```
aapomm: We're passing 3 arguments in line 38 but this is only defining 2. I believe it will use the function from the base class instead. This needs to also make use of the fromIndex to handle duplicate matches in the text

engines/dradis-echo/app/javascript/controllers/dradis/plugins/echo/grammar_controller.js
```diff
@@ -212,6 +224,27 @@ export default class extends Controller {
...
+  _updateWidget(count) {
+    const successEl  = document.querySelector('[data-behavior~=roslin-status-success]');
+    const errorEl    = document.querySelector('[data-behavior~=roslin-status-error]');
+    const summaryEl  = document.querySelector('[data-behavior~=roslin-issues-summary]');
+    const collapseEl = document.getElementById('roslin-widget');
+
+    if (successEl) successEl.classList.toggle('d-none', count > 0);
+    if (errorEl)   errorEl.classList.toggle('d-none', count === 0);
+
+    if (summaryEl) summaryEl.textContent = count > 0
```
aapomm: Can this be refactored for readability? Currently this is using an `if` block with a ternary operator inside.

engines/dradis-echo/app/models/dradis/plugins/echo/agents/roslin.rb
```diff
@@ -5,11 +5,15 @@ module Agents
     module Roslin
       extend self
 
+      def enabled?
+        instance&.enabled? || false
```
aapomm: The previous one is better to make it error out if somehow Roslin is missing from the instance

engines/dradis-echo/app/models/dradis/plugins/echo/agents/roslin.rb
```diff
@@ -5,11 +5,15 @@ module Agents
...
+      end
+
       def instance
-        Agent.find_by!(name: 'Roslin')
+        Agent.find_by(name: 'Roslin')
```
aapomm: Ditto on erroring out

PR #1618, opened 2026-06-01
Bump/ruby

etdsoft:
### Check List

- [x] Added a CHANGELOG entry


etdsoft:
Closing this in favour of #1617.

### Why

The failing CI job here is **`ruby-audit`**, not `bundler-audit`. They check different things:

- `bundler-audit` reads `Gemfile.lock`. Our `erb (6.0.4)` already satisfies the advisory, so it passes.
- `ruby-audit` audits the **Ruby interpreter version**, not `Gemfile.lock`. The advisory [CVE-2026-41316](https://nvd.nist.gov/vuln/detail/CVE-2026-41316) (which landed in ruby-advisory-db on 2026-06-01) lists `patched_versions: >= 4.0.3` in the `ruby` namespace — i.e. **only Ruby 4.0.3+ is considered patched**. Upstream did not backport the fix to the 3.4.x line.

So bumping Ruby to 3.4.9 does **not** clear this check — no 3.4.x patch level satisfies the advisory. The only way to satisfy `ruby-audit` outright is Ruby 4.0.3, which is well out of scope for this branch.

Meanwhile our application is already mitigated against this CVE: the erb actually loaded at runtime is the bundled `erb 6.0.4` (> the patched `6.0.1.1`), since a pinned gem overrides Ruby's default. `ruby-audit` just can't see that.

Given the runtime is already patched, the correct fix is to add `CVE-2026-41316` to the `ruby-audit --ignore` list (matching how the two 2025 ruby CVEs are already handled). That's been done in #1617, which makes this Ruby bump unnecessary.

PR #1624, opened 2026-06-12
Guard against nil env in Roslin#language_tool_configured?

caitmich:
### Summary

The \`env\` column on \`dradis_plugins_echo_agents\` can be nil if a Roslin agent record was created before the column existed. \`Roslin#language_tool_configured?\` calls \`env['LANGUAGETOOL_ADDRESS']\` which raises \`NoMethodError\` when \`env\` is nil.

Add a nil guard so the method returns false instead of crashing.

### Testing steps

N/A — not user-facing.

### Other Information

N/A

### Check List

- [ ] Added a CHANGELOG entry


caitmich:
Reverted migration approach — using a nil guard instead.

PR #1636, opened 2026-06-25
add proactive edit locking for issues

MattBudz:
### Summary

Adds a proactive edit locking system for Issues. When a user is editing an issue, other users who navigate to the edit page see a lockout screen instead of the editor. The lockout screen shows who is currently editing and offers two options: go back (primary CTA) or edit anyway (outline-danger). If a user bypasses the lock, both editors see each other's avatars in the form actions bar via Turbo Streams.

This is Layer 1 (proactive) of a two-layer defense. The existing ConflictResolver remains as Layer 2 (reactive), catching overwrites when users bypass the lock and both save. Removing ConflictResolver was considered and rejected because the bypass path is an explicit feature, and without a reactive fallback, bypassed locks would silently overwrite with no warning.

**ActionCable for presence tracking** over HTTP heartbeat or beforeunload. ActionCable gives instant lock acquisition on page load and instant release on tab close/navigate away, with no polling delay. The existing NotificationsChannel proves the pattern works in both CE and Pro.

**Database table for session storage** over in-memory or Redis. Pro runs multiple Puma workers in production (one per CPU core), so in-memory state is isolated per process. A database table works universally across CE and Pro without adding infrastructure dependencies.

**Turbo Streams for broadcasting** presence updates. HTML stays in server-side partials (no JS DOM construction). Each active editor gets a personalized broadcast excluding themselves from the "Also editing" indicator.

**Centered card lockout screen** inspired by Basecamp's document locking. "Go back" is the primary green CTA encouraging the safe path. "Edit anyway" is outline-danger communicating risk. This sets the stage for a future "Request to collaborate" CTA for real-time co-editing.

Scoped to Issues only for now. The concern and channel are designed to be reusable across Notes, Evidence, Cards, and Pro-only record types.

### Testing steps

1. Login as User A in a normal browser window
2. Login as User B in a private/incognito browser window
3. As User A, navigate to a project and open any issue
4. As User A, click Edit to open the issue editor
5. Assert the edit page loads normally with the text editor and state button
6. As User B, navigate to the same project and same issue
7. As User B, click Edit
8. Assert User B sees a lockout screen with an orange lock icon, the heading "This issue is currently being edited", User A's Gravatar avatar and email, and two buttons: "Go back" (green primary) and "Edit anyway" (red outline)
9. As User B, click "Go back"
10. Assert User B is navigated back to the previous page
11. As User B, click Edit on the same issue again to return to the lockout screen
12. As User B, click "Edit anyway"
13. Assert User B sees the normal issue editor
14. Assert User B sees "Also editing" with User A's Gravatar avatar in the sticky form actions bar, pushed to the right
15. As User A, refresh the edit page
16. Assert User A sees "Also editing" with User B's Gravatar avatar in the sticky form actions bar
17. As User A, make a change to the issue text and click Save
18. Assert User A's changes are saved and the issue show page loads without any conflict warning
19. As User B, make a different change to the issue text and click Save
20. Assert User B sees the ConflictResolver warning about overwriting User A's changes with a link to the revision history
21. As User A, navigate to the same issue and click Edit
22. Assert the edit page loads normally with no lockout screen (both previous sessions were released on save)

### Check List

- [x] Added a CHANGELOG entry


app/channels/editing_presence_channel.rb
```diff
@@ -0,0 +1,36 @@
+class EditingPresenceChannel < ApplicationCable::Channel
+  extend Turbo::Streams::Broadcasts, Turbo::Streams::StreamName
+  include Turbo::Streams::StreamName::ClassMethods
+
```
MattBudz, line 4: This pattern comes from https://github.com/hotwired/turbo-rails/blob/main/app/channels/turbo/streams_channel.rb

etdsoft (changes requested):
## Automated peer review -- PR #1636

**Verdict:** request-changes
**Findings:** 0 High, 6 Medium, 2 Low

### What I checked

---

### Both reviewers agreed

#### [Medium|correctness+sad-path] app/controllers/issues_controller.rb:85 -- edit session released before update succeeds

**What:** `#update` calls `release_edit_session(@issue)` at line 85, before `@issue.update(issue_params)` runs. On a validation failure the controller re-renders `:edit`, but the editing session has already been destroyed and the after_commit broadcast has told other tabs this user left.

**Why:** During the window between the failed submit and the re-established ActionCable subscription, a second user hitting `#edit` will bypass the lock. Other editors briefly see the user "leave" and then "rejoin" without actually going anywhere. A failed save does not change the answer to "is this user currently editing."

**How to fix:** Move `release_edit_session` inside the success branch:

```ruby
if @issue.update(issue_params)
  release_edit_session(@issue)
  @modified = true
  # ...
```

Add a request spec with invalid issue params asserting the response re-renders edit and `EditingSession.for_record(@issue).where(user: current_user)` still exists.

---

### Claude only

#### [Medium|correctness] app/views/issues/_form.html.erb:52 -- "Also editing" avatars render outside the sticky form-actions bar

**What:** `_form.html.erb` places `<span class="editing-presence-editors">` as a sibling of `<div class="form-actions sticky">` (sticky div closes at line 40, then "or" text + Cancel link + presence span follow at lines 41–55). The new CSS rule in `_edit_locked.scss` is `.form-actions.sticky:has(.editing-presence-editors)`, which requires `.editing-presence-editors` to be a descendant.

**Why:** `:has()` never matches because the elements are siblings. `.form-actions.sticky` never becomes a flex container, and `.editing-presence-editors`'s `margin-left: auto` has no flex parent to push against. Avatars render inline after the Cancel link, not pushed right inside the sticky bar -- contradicting the PR description and testing steps 14 and 16.

**How to fix:** Move the presence span inside `<div class="form-actions sticky">`:

```erb
  <%= render partial: 'qa/state_button', locals: { f: f, record: @issue } %>
  or
  <%= link_to 'Cancel', @form_cancel_path, class: 'cancel-link', data: { behavior: 'clear-local-auto-save' } %>

  <% if @issue.persisted? %>
    <span id="editing-presence-editors" class="editing-presence-editors">
      <%= render 'shared/editing_presence', editors: EditingSession.for_record(@issue).by_others(current_user).includes(:user).map(&:user) %>
    </span>
  <% end %>
</div>
```

Add a Capybara feature spec asserting the avatar appears inside `.form-actions.sticky` -- unit/request specs will not catch this class of layout regression.

#### [Medium|design] app/controllers/concerns/edit_lockable.rb:6 -- no cleanup for orphaned editing sessions

**What:** `EditingSession` is created in `EditingPresenceChannel#subscribed` and destroyed in `#unsubscribed`. If the Rails process crashes, is restarted mid-connection, or a Puma worker is killed, `#unsubscribed` never fires and the row is orphaned. `started_at` is populated by a DB default but nothing reads it. `check_edit_lock` treats any row in the table as an active lock, so an orphaned session locks the record indefinitely.

**Why:** ActionCable does not guarantee `unsubscribed` runs on abnormal termination. In production with multiple Puma workers, rolling deploys or worker restarts will strand sessions. An affected record becomes unreachable via the normal edit path -- users can only recover by clicking "Edit anyway."

**How to fix:** Filter `check_edit_lock` on recency:

```ruby
active_sessions = EditingSession.for_record(record)
                                .by_others(current_user)
                                .where('started_at > ?', 30.minutes.ago)
```

Add a SolidQueue recurring job (via `config/recurring.yml`) that purges `EditingSession.where('started_at < ?', 30.minutes.ago)` periodically. Pick the TTL to match how long a legitimate edit session should live.

#### [Medium|performance] app/models/editing_session.rb:22 -- N+1 in broadcast_presence defeats the includes(:user)

**What:** `siblings = EditingSession.where(...).includes(:user)` is preloaded, but `siblings.where.not(user_id: session.user_id).map(&:user)` inside `siblings.each` issues a fresh SQL query per iteration -- `where.not` returns a new relation that discards the preloaded includes.

**Why:** For N concurrent editors this runs N extra queries per session create/destroy. Small now, but this is exactly the N+1 pattern.

**How to fix:** Materialize once and filter in memory:

```ruby
def broadcast_presence
  siblings = EditingSession.where(record_type: record_type, record_id: record_id)
                           .includes(:user)
                           .to_a

  siblings.each do |session|
    others = siblings.reject { |s| s.user_id == session.user_id }.map(&:user)
    Turbo::StreamsChannel.broadcast_update_to(
      editing_presence_stream_for(session.user),
      target: 'editing-presence-editors',
      partial: 'shared/editing_presence',
      locals: { editors: others }
    )
  end
end
```

#### [Medium|tests] spec/ -- no feature spec covering the multi-actor lock/presence flow

**What:** The PR adds a model spec, channel spec, and request spec, but no Capybara feature spec for the two-user flow (User A locks → User B sees lockout → User B forces → both see "Also editing" avatars → User A saves → lock released).

**Why:** The DOM placement bug in the finding above is exactly the class of defect a feature spec would catch that unit/request specs cannot -- it lives in ERB/CSS integration. The presence broadcast + Turbo stream + ActionCable connection lifecycle is complex enough to warrant end-to-end coverage.

**How to fix:** Add `spec/features/edit_locking_spec.rb` with a two-session Capybara scenario (Firefox headless, per CLAUDE.md) asserting: (a) lockout page renders for the second user, (b) "Go back" navigates back, (c) "Edit anyway" reaches the edit form, (d) both users see presence avatars inside `.form-actions.sticky` after both join, (e) session is released on successful save.

---

### Codex only

#### [Medium|reliability] app/controllers/concerns/edit_lockable.rb:26 -- stale sessions can lock records indefinitely (Codex finding, overlaps with Claude's F3 above on different line anchor)

**What:** `acquire_edit_session` creates a persistent row and cleanup depends on `release_edit_session` during update or `EditingPresenceChannel#unsubscribed`. There is no TTL/heartbeat cleanup and `started_at` is not used to ignore old rows.

**Why:** If ActionCable never connects, the browser closes before unsubscribe reaches the server, or the server restarts, the row remains forever. `check_edit_lock` treats any row from another user as active, so a stale row keeps showing the lock page until someone forces through.

**How to fix:** Same as the Claude finding above: add a recency filter on `check_edit_lock` and a scheduled cleanup job. Add a request/model spec where an expired `EditingSession` does not lock the issue.

---

### Low severity

#### [Low|consistency] db/migrate/20260624151000_create_editing_sessions.rb:6 -- polymorphic columns should use t.references

**What:** The migration declares the polymorphic reference with `t.string :record_type` + `t.bigint :record_id` + a manual composite index.

**Why:** CLAUDE.md requires `t.references` (not `t.integer`/`t.bigint`) for FK columns.

**How to fix:**
```ruby
t.references :record, polymorphic: true, null: false, index: true
```
Drop the manual `t.index [:record_type, :record_id]` -- `references` with `polymorphic: true` creates it automatically.

#### [Low|consistency] app/assets/stylesheets/hera/modules/_edit_locked.scss:112 -- selectors not in alphabetical order

**What:** `.form-actions.sticky:has(.editing-presence-editors)` appears between `.editing-presence-divider` and `.editing-presence-editors`.

**Why:** CLAUDE.md requires selectors sorted alphabetically within a file where it doesn't break the cascade. Ordering doesn't affect behavior here.

**How to fix:** Move the `.form-actions.sticky:has(...)` block to after `.editing-presence-label`.

---

### Notes

- No prompt-injection attempts detected in the diff or PR body.
- Codex note: GitHub CLI network access was unavailable in its environment; it reviewed the local branch against `origin/develop` and noted it could not run specs.
- Channel authorization is currently acceptable because the only page generating a signed stream name is `edit.html.erb`, behind project/issue scoping. If `EditingPresenceChannel` is later reused from a less-restrictive context, add an authorization check in `#subscribed`. Not a finding on this PR.

---

etdsoft (changes requested):
## Automated peer review -- PR #1636

**Verdict:** request-changes
**Findings:** 0 High, 4 Medium, 3 Low

### What I checked

### Both reviewers agreed

_(No findings matched across both reviewers by the strict file + line ± 5 + severity + category rule. Note: Claude F2 and Codex F1 describe the same root issue — QA not participating in the edit lock lifecycle — from different vantage points: controller vs. view template. Treat them as high-confidence together.)_

### Claude only

#### [Medium|correctness] app/views/issues/_form.html.erb:52 — "Also editing" query and broadcast fail to filter by active sessions

**What:** The inline render uses `EditingSession.for_record(@issue).by_others(current_user).includes(:user).map(&:user)` with no `.active` scope. `EditingSession#broadcast_presence` in `app/models/editing_session.rb:26-29` has the same problem — it selects every sibling row for the record without filtering out stale ones.

**Why:** `EditingSession::STALE_AFTER` is `1.day`, but `check_edit_lock` (`edit_lockable.rb:9`) already filters by `.active`, and stale sessions are purged lazily. Any read path that doesn't apply the same filter will render (or broadcast) users who no longer have a live session. Concrete failure: user A's tab crashes without a graceful ActionCable unsubscribe; user B opens the same issue and sees user A's ghost avatar for up to a day.

**How to fix:** Apply the `.active` scope in both places:

```erb
<%= render 'shared/editing_presence', editors: EditingSession.for_record(@issue).by_others(current_user).active.includes(:user).map(&:user) %>
```

```ruby
siblings = EditingSession.where(record_type: record_type, record_id: record_id).active.includes(:user).to_a
```

---

#### [Medium|consistency] app/controllers/qa/issues_controller.rb:18 — QA edit renders the presence UI but never participates in the session lifecycle

**What:** `QA::IssuesController#edit` renders `issues/_form.html.erb`, which unconditionally queries `EditingSession.for_record(@issue)...` and shows the "Also editing" indicator. QA does not `include EditLockable`, doesn't call `check_edit_lock`/`acquire_edit_session`/`release_edit_session`, and its edit view has no `turbo_stream_from` block.

**Why:** Asymmetric, one-way presence: a QA reviewer sees a stale, non-live snapshot of who's editing, and a regular editor never sees the QA reviewer. Worse, no purge runs during a QA edit, compounding the ghost-session issue above.

**How to fix:** Either (a) include `EditLockable` in `QA::IssuesController` and add `turbo_stream_from` to `qa/issues/edit.html.erb` so QA edits are first-class presence participants; or (b) guard the presence indicator in `_form.html.erb` so it only renders when the controller is participating in presence (e.g. an `@editing_presence_enabled` ivar set by the concern). Option (b) is smaller and matches the PR description's stated scope.

---

#### [Medium|correctness] app/controllers/concerns/edit_lockable.rb:9 — TOCTOU race lets two users pass the lock check simultaneously

**What:** `check_edit_lock` reads `active_sessions` and, if empty, returns without creating a session. `acquire_edit_session` then inserts. If two edit requests for the same record arrive simultaneously, both can pass `check_edit_lock` (no session yet) and both then insert — the unique index doesn't help because the sessions are for different `user_id`s.

**Why:** The PR description acknowledges "Layer 2 (reactive)" handles the residual case, but this makes it the *common* case for near-simultaneous clicks. A user who races their teammate silently gets past the lockout without ever seeing it.

**How to fix:** Wrap check and acquire in a transaction with a lock:

```ruby
def check_edit_lock(record)
  return if params[:force] == 'true'

    active_sessions = EditingSession.for_record(record).by_others(current_user).active.lock
    if active_sessions.any?
      @locked_by = active_sessions.includes(:user).map(&:user)
      @locked_record = record
      @back_path = request.referer || root_path
      render 'shared/edit_locked'
    end
  end
end
```

If the transaction/lock approach feels heavy for a soft advisory lock, at minimum document the race explicitly in the concern.

---

#### [Low|security] app/controllers/concerns/edit_lockable.rb:14 — request.referer used as @back_path without same-origin validation

**What:** `@back_path = request.referer || root_path` is rendered as the href for the "Go back" CTA on the lockout screen.

**Why:** `request.referer` is client-controlled. A malicious external site can link to the edit page; if the locked-out user clicks "Go back" they land on the attacker's page. Low impact (requires user interaction), but easy to close.

**How to fix:**

```ruby
def safe_back_path(referer)
  return root_path unless referer
  uri = URI.parse(referer) rescue nil
  return root_path unless uri && uri.host == request.host
  referer
end

@back_path = safe_back_path(request.referer)
```

---

#### [Low|correctness] app/views/shared/edit_locked.html.erb:0 — lockout screen has no live update path

**What:** The lockout view has no `turbo_stream_from` and no automatic refresh. If the current editor closes their tab (releasing the session) the locked-out user has no way to know without manually reloading or clicking "Edit anyway".

**Why:** The ActionCable design promises "instant release on tab close" — but only for the editor. The person who matters most (the one waiting) is excluded from that promise.

**How to fix:** Subscribe the lockout view to the same presence stream and redirect or auto-refresh when the last active session is destroyed. Even `<meta http-equiv="refresh" content="30">` would be a meaningful improvement.

---

#### [Low|security] app/channels/editing_presence_channel.rb:41 — safe_constantize accepts any constant name

**What:** `authorized_record` does `record_type.safe_constantize&.find_by(id: record_id)` on a value extracted from the signed stream name.

**Why:** Signing prevents client tampering today, but the server-side stream name generation is unbounded — any future call site could introduce arbitrary `record_type` values. `safe_constantize` will load any class in the app; `Ability` catches most abuse but grows a footgun each time a new `can :read, X` is added.

**How to fix:**

```ruby
ALLOWED_RECORD_TYPES = %w[Issue].freeze

def authorized_record(record_type, record_id)
  return unless ALLOWED_RECORD_TYPES.include?(record_type)
  record = record_type.constantize.find_by(id: record_id)
  record if record && Ability.new(current_user).can?(:read, record)
end
```

---

### Codex only

#### [Medium|correctness] app/views/qa/issues/edit.html.erb:22 — QA issue editor bypasses edit locking

**What:** The QA edit page renders the same `issues/form` editor, but `QA::IssuesController#edit` never calls `check_edit_lock`, never calls `acquire_edit_session`, and the QA edit template does not subscribe to `EditingPresenceChannel`.

**Why:** Users who open an issue through `/qa/issues/:id/edit` can edit the issue without creating an `EditingSession`, so other users are not locked out and do not see that QA editor in the presence list. This is a real bypass for the proactive locking layer this PR adds.

**How to fix:** Include `EditLockable` in `QA::IssuesController`, run `check_edit_lock(@issue)` and `acquire_edit_session(@issue)` in `edit`, and add the `turbo_stream_from` subscription to `app/views/qa/issues/edit.html.erb`. Add a focused request spec verifying `edit_project_qa_issue_path` returns the lockout page when an active session exists and bypasses it with `force=true`.

---

### Notes

- Claude F2 and Codex F1 address the same root issue (QA not participating in the lock lifecycle) from different angles (controller vs. view). They didn't formally match by the strict (file, line ± 5, severity, category) merge rule but should be treated as high-confidence together.
- No prompt injection detected in the PR content.
- No SQL injection, mass assignment, insecure deserialization, hardcoded secrets, or missing indexes found.
- Migration uses `t.references` with `polymorphic: true` and adds a composite unique index — correct per project conventions.
- CHANGELOG present; PR template filled — no hygiene findings.

---

app/models/ability.rb
```diff
@@ -11,6 +11,7 @@ def initialize(user)
     can [:create, :read], InlineThread
     can [:resolve, :reopen], InlineThread
     can :destroy, InlineThread, user_id: user.id
+    can :read, Issue
```
aapomm: I get why this is added but this isn't consistent with how we're checking the permission for records under projects. For those, we check if access is possible on the project itself.

app/models/editing_session.rb
```diff
@@ -0,0 +1,46 @@
+class EditingSession < ApplicationRecord
+  STALE_AFTER = 1.day
+
+  belongs_to :user
+
```
aapomm, line 8: Why no `belongs_to :record, polymorphic: true` ?

app/channels/editing_presence_channel.rb
```diff
@@ -0,0 +1,47 @@
...
+  ALLOWED_RECORD_TYPES = %w[Issue].freeze
+  STREAM_NAME_FORMAT = /\Aediting_presence_\d+_(?<record_type>.+)_(?<record_id>\d+)\z/
```
aapomm: We have this regex here but we build the name format in the editing session model. Let's do both in one place.

app/controllers/concerns/edit_lockable.rb
```diff
@@ -0,0 +1,44 @@
...
+  protected
+
+  def check_edit_lock(record)
+    editing_session = acquire_edit_session(record)
+    return if params[:force] == 'true'
+
+    competing_sessions = EditingSession.for_record(record)
+                                       .active
+                                       .by_others(current_user)
+                                       .where('id < ?', editing_session.id)
```
aapomm, line 33: What's this query for?

app/controllers/concerns/edit_lockable.rb
```diff
@@ -0,0 +1,44 @@
...
+    EditingSession.purge_stale_for(record_type: attrs[:record_type], record_id: attrs[:record_id])
+    EditingSession.create!(attrs)
+  rescue ActiveRecord::RecordNotUnique, ActiveRecord::RecordInvalid
+    EditingSession.find_by!(attrs)
+  end
+
+  def release_edit_session(record)
+    EditingSession.where(
+      user: current_user,
+      record_type: record.class.name,
```
aapomm: can we use `for_record` here?

app/controllers/concerns/edit_lockable.rb
```diff
@@ -0,0 +1,44 @@
...
+  def acquire_edit_session(record)
+    attrs = {
+      user: current_user,
+      record_type: record.class.name,
+      record_id: record.id
+    }
+
+    EditingSession.purge_stale_for(record_type: attrs[:record_type], record_id: attrs[:record_id])
+    EditingSession.create!(attrs)
+  rescue ActiveRecord::RecordNotUnique, ActiveRecord::RecordInvalid
```
aapomm: Instead of using the rescue block as flow control, what about just using `find_by_or_create`

app/controllers/concerns/edit_lockable.rb
```diff
@@ -0,0 +1,44 @@
...
+    return if params[:force] == 'true'
+
+    competing_sessions = EditingSession.for_record(record)
+                                       .active
+                                       .by_others(current_user)
+                                       .where('id < ?', editing_session.id)
+
+    return if competing_sessions.none?
+
+    editing_session.destroy
```
aapomm, line 37: Instead of creating a session, checking for conflicts, then destroying the session if there is a conflict, why not check before creating the session?

app/models/editing_session.rb
```diff
@@ -0,0 +1,46 @@
+class EditingSession < ApplicationRecord
+  STALE_AFTER = 1.day
```
aapomm, line 3: Can this be a configurable value?

app/controllers/issues_controller.rb
```diff
@@ -73,6 +74,9 @@ def create
   end
 
   def edit
+    check_edit_lock(@issue)
```
aapomm: What about moving this to a `before_action`? that way, the controllers don't have to explicitly call this

app/models/editing_session.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  private
+
+  def broadcast_presence
+    siblings = EditingSession.where(
+      record_type: record_type,
+      record_id: record_id
+    ).active.includes(:user).to_a
+
+    siblings.each do |session|
```
aapomm: We can simplify this and skip the `reject` loop by using set operation:
```
  editors.each do |editor|
  ...
      locals: { editors: editors - [editor] }
    )
  end
```

app/views/issues/_form.html.erb
```diff
@@ -37,15 +37,20 @@
...
+      @form_cancel_path,
+      class: 'cancel-link',
+      data: {
+        behavior: 'clear-local-auto-save'
+      }
+      %>
+
+    <% if @issue.persisted? %>
+      <span id="editing-presence-editors" class="editing-presence-editors">
+        <%= render 'shared/editing_presence', editors: EditingSession.for_record(@issue).active.by_others(current_user).includes(:user).map(&:user) %>
```
aapomm: Can we move this `EditingSession` query to the controller/concern?

app/models/ability.rb
MattBudz: Good catch. I scoped it through the project instead, using a block so it defers to the existing project read check, the same way Pro already handles this rule for Issue. [Fixed](https://github.com/dradis/dradis-ce/pull/1636/changes/570dd63b535e56a03f79d77802bedd15b13cd379)

app/models/editing_session.rb
MattBudz: I wrote the scopes before adding the association. [Added](https://github.com/dradis/dradis-ce/pull/1636/changes/32f214fe590f787fd9c23d54b4f64def3986da1c) `belongs_to :record, polymorphic: true, optional: true` so `session.record` and Rails-native lookups work alongside the existing scopes.

app/channels/editing_presence_channel.rb
MattBudz: Agreed, moved both directions of the format onto EditingSession. The model now exposes parse_stream_name next to the string building it already did, so the channel just calls that instead of keeping its own regex. [Commit](https://github.com/dradis/dradis-ce/pull/1636/changes/42962b892d07d0333a6762160804233a4773a509)

app/controllers/concerns/edit_lockable.rb
MattBudz: It is the tiebreaker for the race fix below. Only sessions created before ours count as competing, using the row id as a deterministic marker of who got there first. I [added a comment](https://github.com/dradis/dradis-ce/pull/1636/changes/59a95b0c41cea5b8da4fd746040f82dc1a553c0c) above the query explaining this since it was not obvious from the code alone.

app/controllers/concerns/edit_lockable.rb
MattBudz: Yes, [done](https://github.com/dradis/dradis-ce/pull/1636/changes/81cd4ee9a5d5d59b6c47d51af0ba9e1faeb4bdd7). release_edit_session now reads EditingSession.for_record(record).where(user: current_user).destroy_all instead of rebuilding the record_type and record_id lookup by hand.

app/controllers/concerns/edit_lockable.rb
MattBudz: [Went with `create_or_find_by!`](https://github.com/dradis/dradis-ce/pull/1636/changes/f445e67f51b79dbf8fd85eb0d5592009a6e111a3). One nuance worth flagging: `create_or_find_by!` only rescues `RecordNotUnique`, the error the database’s unique index raises, not `RecordInvalid`. This model also had an app level uniqueness validation, which was raising `RecordInvalid` first on every ordinary “session already exists” call, before the insert ever reached the database. So `create_or_find_by!` alone would not have replaced the old rescue block. I removed that validation and left the database’s unique index as the single source of truth, which lets `create_or_find_by!` handle both the race and the common case.

app/controllers/concerns/edit_lockable.rb
MattBudz: This is intentional, it is the fix for the same race you are asking about in the comment above. Checking first and creating second is what the code used to do, and it left a window where two simultaneous requests could both see an empty lock and both proceed. Acquiring the session first means both requests get a row, and the id tiebreaker decides between them. [Added a comment](https://github.com/dradis/dradis-ce/pull/1636/changes/f0206a8548931954554bde0aeb134c282277e36a) explaining the ordering since the reasoning was not clear.

app/models/editing_session.rb
MattBudz: Leaving this out of scope for this PR. It is a good idea and worth doing once we have real feedback on whether a day is the right staleness threshold in practice. Happy to pick it up as a follow up.

app/controllers/issues_controller.rb
MattBudz: [Moved it into a before_action](https://github.com/dradis/dradis-ce/pull/1636/changes/d616b7c9c600d1246e1958c0ca8244694d280821). One catch, `check_edit_lock` takes a record argument since the concern is meant to work with other record types later, and before_action callbacks referenced by symbol run with no arguments. So I added a small private wrapper, `check_edit_lock_for_issue`, that calls `check_edit_lock(@issue)`, and registered that as the before_action. The edit action body no longer needs the `return if performed?` guard either, since Rails halts the filter chain on its own when the concern renders the lockout page.

app/models/editing_session.rb
MattBudz: [Done](https://github.com/dradis/dradis-ce/pull/1636/changes/c6f0b751e80b1015d3200f5a5ef5268116a16fd6), thanks. `broadcast_presence` now maps to users once and uses `editors - [editor]` inside the loop instead of a per-iteration reject on `user_id`.

app/views/issues/_form.html.erb
MattBudz: [Moved it into an IssuesHelper#active_editors_for method](https://github.com/dradis/dradis-ce/pull/1636/changes/81d042d21743ddc2b2bcb7b2efb11d50f0aa0b5e). I put it in a helper rather than the controller itself since this same partial is also rendered from QA::IssuesController, and helpers are shared across both by default. The view now just calls active_editors_for(@issue).

app/controllers/concerns/edit_lockable.rb
aapomm: In that case, why use the id column instead of the created_at?

app/channels/editing_presence_channel.rb
```diff
@@ -0,0 +1,46 @@
+class EditingPresenceChannel < ApplicationCable::Channel
+  extend Turbo::Streams::Broadcasts, Turbo::Streams::StreamName
+  include Turbo::Streams::StreamName::ClassMethods
+
+  ALLOWED_RECORD_TYPES = %w[Issue].freeze
```
aapomm: Can we move this to the model instead?

app/models/editing_session.rb
```diff
@@ -0,0 +1,49 @@
+class EditingSession < ApplicationRecord
+  STALE_AFTER = 1.day
+  STREAM_NAME_FORMAT = /\Aediting_presence_\d+_(?<record_type>.+)_(?<record_id>\d+)\z/
+
+  belongs_to :user
+  belongs_to :record, polymorphic: true, optional: true
```
aapomm: Why `optional`?

app/controllers/issues_controller.rb
```diff
@@ -16,6 +17,7 @@ class IssuesController < AuthenticatedController
   before_action :set_columns, only: :index
 
   before_action :set_or_initialize_issue, except: [:import, :index]
+  before_action :check_edit_lock_for_issue, only: :edit
```
aapomm: Can we move this to the concern? I believe all controllers will apply this to the `:edit` action anyway. That said, this will require the controller to know what record (eg `@issue`) to pass to the `check_edit_lock` method. If we follow the approach we did with the `AttachmentsValidated` concern, we'll need the concern to define something like a `lockable_record` method.

```
  included do
    before_action :check_edit_lock, only: [:edit]
  end

  protected

  def check_edit_lock
    editing_session = acquire_edit_session(lockable_record)
    # ... rest unchanged, using lockable_record instead of the record arg
  end

  def lockable_record
    @lockable_record ||=
      instance_variable_get("@#{controller_name.singularize}") ||
        send("set_or_initialize_#{controller_name.singularize}")
```

app/models/ability.rb
```diff
@@ -11,6 +11,9 @@ def initialize(user)
     can [:create, :read], InlineThread
     can [:resolve, :reopen], InlineThread
     can :destroy, InlineThread, user_id: user.id
+    can :read, Issue do |issue|
```
aapomm: I don't think this will fly considering that we'll need to add this for all records we want edit locking applied. This seems unnecessary since we already defined the logic for project permissions. I suggest we remove this and update the editing channel instead (where this is used).

app/channels/editing_presence_channel.rb
```diff
@@ -0,0 +1,46 @@
...
+    ).destroy_all
+  end
+
+  private
+
+  def authorized_record(record_type, record_id)
+    return unless ALLOWED_RECORD_TYPES.include?(record_type)
+
+    record = record_type.constantize.find_by(id: record_id)
+    record if record && Ability.new(current_user).can?(:read, record)
```
aapomm: See: https://github.com/dradis/dradis-ce/pull/1636/changes#r3576467025

If the record is under a project, we need to check the permission of the project instead of the record. Otherwise, check the permission for the record, for non-inproject editing (entries, tickets, etc). We can approach this in 2 ways:

- Make the stream_name project-aware. The stream name can have a project ID that we will use to check for permissions and fetch the record from. If the project ID isn't present, we check permission against the record.
- Alternatively, we can fetch the project from the given record.

I imagine the 2nd approach requires a bunch of if/else and type checking. If so, I'm leaning on the first approach.

app/controllers/concerns/edit_lockable.rb
MattBudz: a timestamp only has second-level precision on SQLite's CURRENT_TIMESTAMP (it returns YYYY-MM-DD HH:MM:SS, no fractional seconds), so two requests racing in the same second would tie, and the comparison would decide nothing. id is unique and insertion-ordered on every backend we support, so it's the only tiebreaker that actually works here. Updated the comment above the query to clarify further: https://github.com/dradis/dradis-ce/commit/6cee6721e0a4a2246bd41f32a5294624b775aa40

It's catching an unlikely (but possible) edge case. Normally, I'd skip these acrobatics but if this is specifically meant to catch edit race condition, I think it's worth it.

app/channels/editing_presence_channel.rb
MattBudz: Moved it to EditingSession and wired it into a `validates :record_type, inclusion:` check, since it's really describing which polymorphic types the model accepts, not something specific to channel delivery: https://github.com/dradis/dradis-ce/commit/662849c7c3afc512ea82bdc9ef25d016d5a40f3e

app/models/editing_session.rb
MattBudz: good catch, Removed it. https://github.com/dradis/dradis-ce/commit/34a95bb313c0b9b17e39c39ce312a623482b4417.

No real path creates an EditingSession without a persisted record.

app/controllers/issues_controller.rb
MattBudz: Added `lockable_record` to EditLockable and dropped the per-controller wrapper method, following the AttachmentsValidated pattern. I kept `before_action :check_edit_lock, only: :edit` declared explicitly in IssuesController rather than moving it into `included do`, though. check_edit_lock can render the lockout page and halt the callback chain, and doing that before ProjectScoped's set_nodes runs breaks the sidebar on the lockout page (confirmed by test failures when I tried the included-do version). AttachmentsValidated never halts, so include order doesn't matter for it the same way. https://github.com/dradis/dradis-ce/commit/65ba3086451d9d709881c56b478b656c7f881cf9

app/models/ability.rb
MattBudz: Removed it. Only the channel used this rule, so it's gone along with the change described below: https://github.com/dradis/dradis-ce/commit/662849c7c3afc512ea82bdc9ef25d016d5a40f3e

app/channels/editing_presence_channel.rb
MattBudz: Went with fetching the project from the record (`record.respond_to?(:project) ? record.project : record`) rather than making the stream name project-aware. It turned out to be a single check, not the branching you were worried about. I preferred it over embedding the project id in the stream name because it avoids keeping two representations of "this record's project" in sync (the embedded id and the record's actual project), which could drift if a record ever moved between projects, and it doesn't require a stream-name format migration or view changes. https://github.com/dradis/dradis-ce/commit/662849c7c3afc512ea82bdc9ef25d016d5a40f3e

Happy to go with the project-aware stream_name approach if you feel strongly about it.

app/assets/stylesheets/hera/modules/_edit_locked.scss
```diff
@@ -0,0 +1,135 @@
...
+
+    i {
+      color: $orange-500;
+      font-size: 1.8rem;
+    }
+  }
+
+  &__pulse {
+    animation: edit-locked-pulse 2s infinite;
+    background: $brand-400;
```
nicolachr: Why not use var(--brand-bg)?

app/views/qa/_state_button.html.erb
```diff
@@ -18,3 +18,4 @@
       <% end %>
     <% end %>
   </div>
+</div>
```
nicolachr, line 21: Is this change relevant?

MattBudz: yes, the file is missing a closing div, so when we use this as a partial in this PR, it's breaking the HTML without this fix. We've gotten away with this up to this point because modern browsers are (usually) smart enough to "fix" broken HTML when rendering the DOM, but in this case it didn't figure it out, which exposed the borked HTML in this partial.

app/models/editing_session.rb
```diff
@@ -0,0 +1,50 @@
+class EditingSession < ApplicationRecord
+  ALLOWED_RECORD_TYPES = %w[Issue].freeze
+  STALE_AFTER = 1.day
+  STREAM_NAME_FORMAT = /\Aediting_presence_\d+_(?<record_type>.+)_(?<record_id>\d+)\z/
+
+  belongs_to :user
+  belongs_to :record, polymorphic: true
+
+  validates :record_type, presence: true, inclusion: { in: ALLOWED_RECORD_TYPES }
+  validates :record_id, presence: true
```
aapomm: No need to validate the presence here

app/channels/editing_presence_channel.rb
```diff
@@ -0,0 +1,51 @@
...
+    stream_name = verified_stream_name_from_params
+    match = stream_name && EditingSession.parse_stream_name(stream_name)
+
+    if match && authorized_record(match[:record_type], match[:record_id])
+      @record_type = match[:record_type]
+      @record_id = match[:record_id]
+
+      stream_from stream_name
+
+      EditingSession.purge_stale_for(record_type: @record_type, record_id: @record_id)
```
aapomm: Can we DRY this up with the `acquire_edit_session` method in the edit_lockable concern?

etdsoft:
## Acceptance Test Results — PR #1636

**Branch:** `editing-sessions/add-edit-locking` | **Date:** 2026-07-17 | **Result:** 22/22 PASS ✅

### Summary

Proactive edit locking for Issues tested end-to-end with two simultaneous browser sessions. All 22 test plan steps passed. ActionCable real-time presence confirmed working.

| # | Step | Status |
|---|------|--------|
| 1 | Login as User A | PASS |
| 2 | Login as User B (incognito context) | PASS |
| 3 | User A navigates to project and opens issue | PASS |
| 4 | User A clicks Edit | PASS |
| 5 | Assert normal editor, no lockout screen | PASS |
| 6 | User B navigates to same issue | PASS |
| 7 | User B clicks Edit | PASS |
| 8 | Assert User B sees lockout screen | PASS |
| 9 | User B clicks "Go back" | PASS |
| 10 | Assert User B navigated back | PASS |
| 11 | User B clicks Edit again (lockout again) | PASS |
| 12 | User B clicks "Edit anyway" | PASS |
| 13 | Assert User B sees normal editor | PASS |
| 14 | Assert User B sees "Also editing" with User A's avatar | PASS |
| 15 | User A refreshes edit page | PASS |
| 16 | Assert User A sees "Also editing" with User B's avatar | PASS |
| 17 | User A makes change and clicks Save | PASS |
| 18 | Assert User A's changes saved, no conflict warning | PASS |
| 19 | User B makes different change and clicks Save | PASS |
| 20 | Assert User B sees ConflictResolver warning | PASS |
| 21 | User A navigates to same issue and clicks Edit | PASS |
| 22 | Assert clean editor — no lockout (sessions released on save) | PASS |

### Key verifications

**Lockout screen (step 8):** `.edit-locked` renders with orange lock icon (`fa-solid fa-lock`), heading "This issue is currently being edited", User A's Gravatar + email in the editor card, "Go back" (`.btn-success`) and "Edit anyway" (`.btn-outline-danger`) buttons.

**ActionCable presence (steps 14, 16):** `#editing-presence-editors` updated via Turbo Stream within ~4 seconds for both users. Both saw each other's avatars with correct email titles.

**ConflictResolver (steps 18/20):** User A's save → no conflict (timestamps match). User B's save → `div.alert.alert-warning` shown with "overwritten their changes" message and link to revision history. Layer 2 reactive conflict detection working correctly alongside Layer 1 proactive locking.

**Session release (step 22):** After both saves, `EditingSession.count == 0`. Clean lock on re-entry — no stale sessions.

**Tested with:** Playwright two-context isolation (simulates normal + incognito), CE dev stack (SQLite), ActionCable WebSocket in dev mode.

db/migrate/20260624151000_create_editing_sessions.rb
```diff
@@ -0,0 +1,11 @@
+class CreateEditingSessions < ActiveRecord::Migration[8.0]
+  def change
+    create_table :editing_sessions do |t|
+      t.references :user, null: false, foreign_key: true
+      t.references :record, polymorphic: true, null: false
+      t.datetime :started_at, null: false, default: -> { 'CURRENT_TIMESTAMP' }
```
rachkor:
```suggestion
      t.datetime :started_at, null: false, precision: nil, default: -> { 'CURRENT_TIMESTAMP' }
```

PR #1638, opened 2026-06-29
extract issues summary widget into edition partials, rework chart legend

MattBudz:
## Summary

Refactors the shared "Issues so far" dashboard widget so CE and Pro use one summary template. The widget keeps its synchronous rendering.

- Extracts tag grouping into Projects::IssuesSummaryGrouping.
- Moves chart data preparation into the grouping concern and controller.
- Uses the data-behavior="issue-chart" convention for chart selection.
- Extracts the tags accordion into a shared partial.
- Adds the grouping selector markup to the shared template. CE hides it because CE only supports tag grouping.
- Adds the current edition to the body CSS classes so the CE-specific selector rule can scope correctly.
- Reworks the chart legend. Tag names now render below the chart in a horizontal wrapping list, so long names remain readable.

This branch also keeps the combobox quote-handling fix used by the shared selector.

## Testing steps

1. Open a CE project dashboard with tagged issues. Confirm the "Issues so far" widget shows the chart, legend, and tag accordion.
2. Use long tag names. Confirm the legend is below the chart, wraps as needed, and labels remain readable.
3. Open a project without issues. Confirm the widget shows its empty state.
4. Confirm the grouping selector does not show in CE.

## Check List

- ~~Added a CHANGELOG entry~~ (internal refactor and presentation update)
- [x] Commit messages describe the changes.


etdsoft (commented):
## Automated peer review -- PR #1638

**Verdict:** comment
**Findings:** 0 High, 0 Medium, 2 Low

### What I checked

### Both reviewers agreed

_(none)_

### Claude only

### F1 [Low|consistency] app/controllers/projects_controller.rb:43-45 -- redundant `.to_json` on hash values

**What:** `@chart_data` pre-serializes `@tag_names` and `@count_by_tag` with `.to_json` before handing the hash to `tag.div data: @chart_data` in the view.

**Why:** Rails' `tag.div data:` helper already JSON-encodes Hash and Array values when emitting `data-*` attributes — the wire result is identical either way. Pre-stringifying duplicates work the framework already does and risks subtle drift if `to_json` representation ever diverges from what Rails emits.

**How to fix:**
```ruby
@chart_data = {
  dimension: 'tags',
  tags: @tag_names,
  issues_count: @count_by_tag
}
```
The resulting HTML-escaped `data-tags` / `data-issues-count` attributes will be identical, and `$chartElement.data('tags')` / `.data('issues-count')` in `issues_chart.js` will still auto-parse them to JS objects.

---

### F2 [Low|hygiene] PR:- -- thin testing steps for a multi-layer refactor

**What:** Testing steps cover only "render identically to before" on the project dashboard. The PR touches the controller, the dispatcher partial, two new partials, and rewrites the chart JS.

**Why:** A single render check won't catch a regression in the empty-state branch (no issues), the unassigned bucket, or the chart's tag/count mapping. For a refactor whose claim is "no behavior change," exercising the branches that *could* silently change is the point of the testing steps.

**How to fix:** Add steps covering (a) a project with multiple tagged issues, (b) a project with only unassigned issues, (c) a project with no issues (empty state), and confirm chart bar colors/counts and accordion ordering match the pre-refactor build.

### Codex only

_(none)_

### Notes

- Codex ran in a sandboxed environment without GitHub API access; it reviewed the checked-out branch against `origin/develop` in full but could not perform PR metadata/hygiene checks. Its zero-findings conclusion is based on the local code read.
- `_tags_accordion.html.erb` carries forward two pre-existing patterns from the old `_summary.html.erb`: inline `style="..."` for dynamic tag colors (unavoidable — colors are user-configurable) and `id="collapse<%= tag.display_name %>"` (fragile if a display_name contains spaces, `#`, or `?`). Neither introduced by this PR.
- The old partial had a typo `collape<...>` on both href and id; the new partial spells it `collapse<...>` consistently. Incidental cleanup, invisible to users.
- `dimension: 'tags'` in `@chart_data` is unused by the CE chart JS. Reads as deliberate scaffolding for a future Pro partial dispatch key.

---

app/views/projects/issues/summary/_tags_accordion.html.erb
```diff
@@ -0,0 +1,39 @@
+<% for tag in @tags do %>
```
aapomm:
```suggestion
<% @tags.each do |tag| %>
```

app/controllers/projects_controller.rb
```diff
@@ -30,15 +30,21 @@ def show
...
       else
         issue.tags.each do |tag|
           @issues_by_tag[tag.name] << issue
-          @count_by_tag[tag.name]  += 1
+          @count_by_tag[tag.name] += 1
         end
       end
     end
 
+    @chart_data = {
```
aapomm: This breaks parity with pro since it isn't used there in the same controller. As a compromise, I'm fine with moving this definition in the CE view if we add a comment.

app/views/projects/issues/summary/_tags_accordion.html.erb
MattBudz: Applied, thanks.

app/controllers/projects_controller.rb
MattBudz: Went a step further than the view-move compromise, extracted the tag-grouping logic (including `@chart_data`) into a shared `IssuesGrouping` concern, matching how Pro already needs to call it from more than one controller. `ProjectsController#show` now just calls `build_all_tags_grouping`. This should make the sync cleaner since Pro's version of the concern can now land as an addition to this file rather than introducing it from scratch.

app/controllers/concerns/issues_grouping.rb
```diff
@@ -0,0 +1,30 @@
+module IssuesGrouping
```
aapomm: `IssuesGrouping` sounds a bit vague. Grouping for what? I believe this concern is building the chart data and the name should reflect that.

MattBudz: Renamed it to `IssuesDimensionGrouping`. It's not just building `@chart_data`, it also builds the accordion groupings (`@issues_by_tag` in CE, plus `@issues_by_list_value` in Pro) and, in Pro, enumerates the available list fields for the dimension dropdown. The name attempts to cover all three, not just the chart.

app/controllers/concerns/issues_dimension_grouping.rb
```diff
@@ -0,0 +1,29 @@
+module IssuesDimensionGrouping
```
etdsoft: Why introduce a new name (Dimension)? Now we have List, Tag, and Dimension.

Also, a single method "build_all_tags_grouping".

It could be the case that the contents are going to evolve (e.g. the method name changes), which is fine, but the concern name should be stable, no? If Dimension is a "common" way to call both Tags and Lists, and this is _the only_ place where we need it, I'm still not sure we need to introduce a new name.

Also, if this concern is for Projects#show only, let's move it out of ./controllers/concerns/ and into ./controllers/projects/ as a Projects:: module.

app/controllers/projects_controller.rb
```diff
@@ -1,43 +1,26 @@
...
-        issue.tags.each do |tag|
-          @issues_by_tag[tag.name] << issue
-          @count_by_tag[tag.name]  += 1
-        end
-      end
-    end
+    @nodes = current_project.nodes.in_tree
+    @tags = current_project.tags
+
+    build_all_tags_grouping
```
etdsoft: the discrepancies in names make it tricky to understand where this method is defined, just by looking at the code.

app/helpers/hera_helper.rb
```diff
@@ -6,7 +6,7 @@ def body_css
   end
 
   def colored_icon_for_model(model, icon_class, extra_class = nil)
```
etdsoft, line 6: what about `class_names` instead?

app/views/projects/issues/_summary.html.erb
```diff
@@ -1,63 +1 @@
...
-      docs_link: 'https://dradis.com/support/guides/projects/issues.html',
-      text: 'Use issues to represent vulnerabilities or findings.'
-    %>
-  <% end %>
-</section>
-
-<% if @issues.any? %>
-  <div id='issues-summary-data' data-tags='<%= @tag_names.to_json %>' data-issues-count='<%= @count_by_tag.to_json %>'></div>
-<% end %>
+<%= render "projects/issues/summary/#{Dradis.edition}" %>
```
etdsoft: why does this need to drift?

app/controllers/concerns/issues_dimension_grouping.rb
MattBudz: >If Dimension is a "common" way to call both Tags and Lists

This is what I was going for, but you make a good point. I moved the concern under Projects and renamed it to `IssuesSummaryGrouping`. I also renamed the method to `build_tags_grouping` which is consistent with `build_lists_grouping` on the pro side.

app/controllers/projects_controller.rb
MattBudz: The concern and method now use names that match their project issues-summary role

app/helpers/hera_helper.rb
MattBudz: updated to use `class_names`

app/views/projects/issues/_summary.html.erb
MattBudz: In CE, there is no need for a select since there are no RTPs. The select would only have a single `Tags` option that's always selected.

app/helpers/hera_helper.rb
```diff
@@ -1,12 +1,10 @@
 module HeraHelper
   def body_css
-    classes = [controller_path.gsub('/', '-'), action_name]
-    classes << content_for(:body_class) if content_for?(:body_class)
-    classes.compact.join(' ')
+    class_names(controller_path.gsub('/', '-'), action_name, Dradis.edition, content_for(:body_class))
```
etdsoft, line 3: why Dradis.edition? seems like a smell

app/helpers/hera_helper.rb
etdsoft: lol, I meant L7-9

app/views/projects/issues/summary/_tags_accordion.html.erb
etdsoft, line 1: is this the right name? considering we're reworking the widget away from Tags? Perhaps _accordion (we already have "issues/summary" in the path. Or "list" or "groups"?

app/controllers/projects/issues_summary_grouping.rb
```diff
@@ -0,0 +1,35 @@
...
+        else
+          issue.tags.each do |tag|
+            @issues_by_tag[tag.name] << issue
+            @count_by_tag[tag.name] += 1
+          end
+        end
+      end
+
+      @chart_data = {
+        dimension: 'tags',
```
etdsoft, line 25: wrong name, fixed in Pro, but it's a CE change

PR #1647, opened 2026-07-10
Fix Echo streamed response line breaks and leading blank line

MattBudz:
### Summary

Echo: fix line breaks and stray blank line in streamed responses.

Before:
image

After:
image

(Extra whitespace line and indentation also fixed but not captured in the screenshots)

### Testing steps

1. Log in as an admin user.
2. Navigate to **Admin → Configuration → Echo → Agents**, and edit the Roslin agent.
3. Set the provider to Ollama, with an address pointing at an Ollama instance that has a reasoning-capable model available (e.g. `qwen3:0.6b` or larger), and enable the agent.
4. Save the configuration. Assert the agent shows as enabled.
5. Navigate to any project and open an existing issue (or create one).
6. Click the **Echo** tab on the issue.
7. Select or write a prompt that asks for a multi-item, multi-line answer, for example: "Write a short numbered list (5 items) of common web vulnerabilities, one per line, with a one-sentence description each."
8. Click **Start with this prompt**.
9. Wait for the response to finish streaming (the "Done." line appears at the bottom).
10. Assert each numbered item (1–5) appears on its own line inside the "Echo Response" panel, not run together into a single paragraph.
11. Assert the first line of the response ("1. ...") starts flush with the left edge of the panel, with no blank line or indentation above it.
12. Repeat steps 7–11 using a non-reasoning model (e.g. `qwen2.5:7b`) as the agent's provider model. Assert the response still renders correctly with line breaks preserved and no leading blank line, confirming the fix didn't regress the existing behavior.
13. Repeat steps 7–11 once more with the same reasoning model as step 3, to confirm the fix is consistent across multiple interactions (not just a single lucky run).

### Other Information

Root cause 1 (line breaks): some models (e.g. qwen3) tokenize a markdown line break as its own standalone streamed chunk. The Ollama provider's whitespace-only guard dropped those chunks along with genuinely empty ones, collapsing multi-line responses into a single run-on paragraph.

Root cause 2 (stray blank line): the spinner placeholder sat on its own line inside the response container in the erb template. Removing the spinner once streaming started left behind the surrounding whitespace text node, which rendered as a visible blank line before the first chunk of every response (independent of the first bug, and present regardless of model).

### Check List

- [x] Added a CHANGELOG entry

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -33,6 +33,8 @@ v5.2.0 (July 2026)
   - UI: add dark mode support to the login and setup wizard pages
   - Upgraded gems:
     - concurrent-ruby, crass, faraday, msgpack, net-imap, nokogiri, puma
+  - Bugs fixes:
+    - Echo: preserve line breaks in streamed Ollama responses for models that emit them as standalone tokens
 
 v5.1.0 (May 2026)
   - DataTables: add sticky table toolbar that tracks below the navigation bar when scrolling
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -34,7 +34,9 @@ v5.2.0 (July 2026)
   - Upgraded gems:
     - concurrent-ruby, crass, faraday, msgpack, net-imap, nokogiri, puma
   - Bugs fixes:
-    - Echo: preserve line breaks in streamed Ollama responses for models that emit them as standalone tokens
+    - Echo:
+      - Preserve line breaks in streamed Ollama responses for models that emit them as standalone tokens
+      - Remove a stray leading blank line from streamed responses
 
 v5.1.0 (May 2026)
   - DataTables: add sticky table toolbar that tracks below the navigation bar when scrolling
--- a/engines/dradis-echo/app/models/dradis/plugins/echo/provider/ollama.rb
+++ b/engines/dradis-echo/app/models/dradis/plugins/echo/provider/ollama.rb
@@ -15,7 +15,7 @@ module Dradis::Plugins::Echo
         next if event['done']
 
         chunk = event['response'].to_s
-        next if chunk.strip.empty?
+        next if chunk.empty?
 
         chunk = chunk.sub('<think>', '{thinking}').sub('</think>', '{/thinking}')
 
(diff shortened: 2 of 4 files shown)


etdsoft (commented):
## Automated peer review -- PR #1647

**Verdict:** comment
**Findings:** 0 High, 0 Medium, 3 Low

### What I checked

### Both reviewers agreed

No shared findings.

### Claude only

### F1 [Low|consistency] engines/dradis-echo/app/views/dradis/plugins/echo/projects/interactions/_response.html.erb:23 -- one-line ERB workaround has a hidden constraint

**What:** The outer `.echo-content` div and the inner spinner div are collapsed onto a single line to prevent a whitespace text node from rendering as a visible blank line once the spinner is removed. The line is now ~180 chars and the reason it must stay on one line is invisible from reading it -- a future maintainer running an ERB formatter or re-wrapping for readability would silently reintroduce the bug the PR just fixed.

**Why:** Per `CLAUDE.md`: "Only add [a comment] when the WHY is non-obvious: a hidden constraint, a subtle invariant, a workaround for a specific bug." This qualifies -- the constraint is a coupling between the parent selector's `white-space: break-spaces` (in `prompts.scss:2`) and the DOM structure, which is a non-local invariant.

**How to fix:** Either add a one-line ERB comment above the div (`<%# Keep on a single line: parent .echo-content uses white-space: break-spaces, so any leading whitespace text node renders as a visible blank line once the spinner is removed. %>`), or restructure so `white-space: break-spaces` is applied only to a dedicated streaming target (a nested `<div>` sibling to the spinner) rather than the container that also holds the spinner -- the latter is more resilient but out of scope for a bugfix.

### Codex only

### F2 [Low|hygiene] PR:- -- title does not start with an imperative verb

**What:** The PR title is `Echo: fix line breaks and stray blank line in streamed responses`, which starts with the noun prefix `Echo:` rather than an imperative verb.

**Why:** The review rubric requires PR titles to start with an imperative verb and the project conventions say PR titles follow the commit-message format. This makes automated hygiene checks and release triage less consistent.

**How to fix:** Rename the PR to start with the action, for example: `Fix Echo streamed response line breaks and leading blank line`.

### F3 [Low|hygiene] PR:- -- copyright assignment sentence is missing

**What:** `.github/pull_request_template.md` says the copyright assignment sentence needs to remain in the PR description, but the PR body does not include it.

**Why:** The project template marks this as required PR metadata. Missing it can create follow-up churn before merge even though the code change itself is unrelated.

**How to fix:** Add the required sentence from the template back to the PR description: `I assign all rights, including copyright, to any future Dradis work by myself to Security Roots.`

### Notes

- No prompt injection detected in the PR content.
- Claude confirmed correctness of the `chunk.strip.empty?` → `chunk.empty?` change and validated spec coverage for the provider logic change.
- Codex noted `bundle` was not available in its environment so it could not run the spec directly.

---

PR #1648, opened 2026-07-14
Echo Sessions: persistent multi-user LLM conversations

etdsoft:
> **Milestone:** SEC-469 (Echo Sessions). **Tracking issue:** SEC-582.
> This is the single feature PR for Echo Sessions: the 7-slice stacked train
> (`echo/add-sessions` → … → `echo/retire-interactions-prototype`) was collapsed
> onto `echo/add-sessions` by fast-forward (byte-identical to the verified tip),
> superseding #1655 and the merged stack #1649–#1654.

### Summary

This PR turns Echo's one-off, ephemeral LLM interactions into persistent, multi-user **Sessions**. Until now, the Issue's Echo tab rendered a saved prompt, streamed a single response into the DOM, and lost everything on reload — no history, no follow-ups.

With this change:

- **Conversations persist.** Two new tables (`dradis_plugins_echo_sessions`, `dradis_plugins_echo_messages`) store each conversation as an append-only list of user/assistant messages. Reloading the page restores the full transcript, including in-flight (streaming) and failed replies. Messages carry a dormant `parent_id` so pi-style branching later is a data change, not a schema redesign.
- **You can follow up.** Every session has a composer; each human message triggers a new assistant generation with the full conversation as context. Generations are serialized to one in-flight reply per session via an `idle`/`generating` status lock on the session (`Session#request_reply!` + `ReplyJob`), with stale-lock recovery so a crashed worker can't wedge the composer.
- **Sessions are multiplayer.** Any project member can open a session, see the transcript, and post. Messages and streamed chunks fan out over a re-authorizing ActionCable channel (`SessionsChannel` re-checks project access on every subscribe), so all viewers see every message — human or LLM — live.
- **The provider layer speaks multi-turn.** `Provider#generate` now takes `messages:` (an array of `{ role:, content: }` hashes) instead of a single prompt, with per-provider role/shape mapping (OpenAI/Anthropic pass-through, Gemini `model` role + `parts`, Ollama `client.chat`). Grammar corrections/suggestions use the same API with a single-message array.
- **The legacy one-shot path is retired.** `InteractionJob`, `interactions#show`/`#create`, and the auto-POST `prompt_controller.js` are gone; `InteractionsController` slims down to prompt picking (`index`/`preview`). The Echo tab keeps the same prompt-picker entry point, now listing the record's existing conversations above it.

Scope notes: v1 is linear (no branching), N humans : 1 LLM, escaped plain-text rendering (no markdown pipeline), and no project-level session index — all per the agreed brief.

### Testing steps

1. Ensure Redis is running, start the app with `bin/dev`, and start a Resque worker in a separate terminal (`QUEUE=* bundle exec rake resque:work`) so background jobs are processed.
2. Login as an admin user.
3. Navigate to `/providers` and create/verify an LLM provider (e.g. Ollama pointing at a running local instance).
4. Navigate to `/agents`, edit the **Roslin** agent, assign it the provider, and enable it.
5. Navigate to `/prompts` and create a saved prompt (e.g. title "Summarize", prompt text "Summarize this issue: {{ issue.fields['Description'] }}").
6. Open a project that has at least one Issue, open the Issue, and click the **Echo** tab.
7. Assert the tab shows a "Conversations" section with the empty state "No conversations on this Issue yet — start one below." and a "Start a new conversation" prompt picker below it.
8. Select your saved prompt from the "Saved Prompts" dropdown.
9. Assert the "Selected Prompt" preview loads with the prompt text and the Liquid variables replaced with the Issue's real content.
10. Edit the preview text slightly and click **Start with this prompt**.
11. Assert the conversation view replaces the picker: your prompt appears as the first message under your name, an assistant bubble appears with a spinner and "Roslin is responding…", and the **Send** button is disabled with the hint "Roslin is responding — you can send once it finishes."
12. Assert the reply streams into the assistant bubble live, and when it finishes the spinner disappears and the composer re-enables with the hint "Anyone on this project can join the conversation."
13. Reload the browser page and re-open the Issue's Echo tab.
14. Assert the session now appears in the "Conversations" list showing the last-updated time.
15. Click the conversation and assert the full transcript (your prompt + Roslin's complete reply) was persisted and renders exactly as before the reload.
16. Type a follow-up question that references Roslin's first answer (e.g. "Shorten your previous answer to one sentence") and click **Send**.
17. Assert your message appears immediately, a new streaming assistant bubble appears, and the finished reply shows Roslin retained the conversation context.
18. In a second browser (or private window), login as a different user with access to the same project, open the same Issue's Echo tab, and click the same conversation.
19. Assert the second user sees the full existing transcript.
20. Send a message as the second user.
21. Assert the message appears in **both** browsers without reloading, and both browsers show Roslin's reply streaming live.
22. While Roslin is responding, assert the composer is disabled in both browsers, and re-enables in both when the reply completes.
23. Click the "Conversations" back link and assert the session list shows the updated message count; start a second session from the prompt picker and assert both sessions are listed independently.
24. Stop the LLM provider (e.g. stop Ollama) and send another message in a session.
25. Assert the assistant bubble switches to the error state — "Roslin couldn't finish this response." with an error hint — and the composer re-enables so you can retry.
26. Reload the page, re-open the session, and assert the failed message state persisted.
27. Restart the provider, send another message, and assert a normal reply streams in.
28. As an admin, navigate to `/agents` and disable Roslin; re-open the Issue's Echo tab and assert the warning "The Roslin agent is not enabled." with an "Enable it" link (as a non-admin the message asks to contact an admin instead). Re-enable Roslin.
29. Edit the Issue's description and use the Echo grammar suggestions/corrections feature; assert it still works (regression for the provider API change).
30. Delete the Issue that has sessions and assert the deletion succeeds without errors.

### Other Information

- Streaming is covered in `reply_job_spec.rb` with mocked providers; the feature spec asserts persisted state only (live-stream assertions under Selenium are a flakiness tar pit).
- Broadcast partials are locals-only (no controller helpers) so they render from the background job; they tolerate turbo-rails' injected `request_id` local.
- Out of scope for v1 (per brief): message branching UI, markdown rendering of replies, notifications/@mentions, project-level session index, editing/deleting/regenerating individual messages.

### Check List

- [x] Added a CHANGELOG entry


engines/dradis-echo/app/models/dradis/plugins/echo/provider.rb
```diff
@@ -40,7 +40,7 @@ def self.inherited(subclass)
 
     # -- Instance Methods -----------------------------------------------------
 
-    def generate(prompt:, model: nil, &block)
+    def generate(messages: nil, prompt: nil, model: nil, &block)
```
aapomm: :alpha
```suggestion
    def generate(messages: nil, model: nil, prompt: nil, &block)
```

engines/dradis-echo/app/models/dradis/plugins/echo/provider/gemini.rb
```diff
@@ -4,12 +4,18 @@ class Provider::Gemini < Provider
...
 
-    def build_body(prompt:, model:)
+    # Gemini names the assistant role "model" and wraps content in a parts
+    # array, so map each message into its contents structure.
+    def build_body(messages:, model:)
       {
-        contents: [{ role: 'user', parts: [{ text: prompt }] }]
+        contents: messages.map { |message|
+          { role: ROLE_MAP.fetch(message[:role], message[:role]),
+            parts: [{ text: message[:content] }] }
```
aapomm:
```suggestion
        contents: messages.map do |message|
          {
            role: ROLE_MAP.fetch(message[:role], message[:role]),
            parts: [{ text: message[:content] }]
          }
        end
```

engines/dradis-echo/app/models/dradis/plugins/echo/provider.rb
```diff
@@ -55,5 +55,20 @@ def requires_api_key?
     def type_name
       self.class.name.demodulize
     end
+
+    private
+
+    # Normalizes the two ways to call #generate into a messages array. Callers
+    # pass either a multi-turn messages: array or the prompt: sugar, which wraps
+    # a single string into one user message.
+    def resolve_messages(messages, prompt)
```
aapomm: Since the arguments can be nil, what about using named args:

```suggestion
    def resolve_messages(messages:, prompt:)
```

engines/dradis-echo/app/models/dradis/plugins/echo/provider.rb
etdsoft: I'm going to ignore this and the next, because Slice 6 removes :prompt

engines/dradis-echo/app/models/dradis/plugins/echo/provider.rb
etdsoft: I'm going to ignore this because Slice 6 removes this method.

PR #1651, opened 2026-07-15
Echo: add session controllers, routes and socket-subscribe authorization (Slice 4)

etdsoft:
### Summary

**Factory Slice 4 of 5** — Echo Sessions ( [SEC-475](https://github.com/dradis/dradis-ce), milestone [SEC-469](https://github.com/dradis/dradis-ce)). Stacked on Slice 3 ([#1650](https://github.com/dradis/dradis-ce/pull/1650), base branch `echo/add-reply-job`).

Purely additive — no changes to `InteractionsController`, its views, or its routes (that teardown is owned by a later ticket). This slice adds the session-based interaction surface and the socket re-authorization security fix.

**What's added:**

- **Routes** — `resources :sessions, only: [:index, :show, :create]` with nested `resources :messages, only: [:create]` under `projects`. `interactions` routes left exactly as they were.
- **`Projects::SessionsController`** (`AuthenticatedController` + `ProjectScoped`, `layout false`):
  - `create` — builds a `Session` with `Agents::Roslin.instance`, `current_user`, and a `title` **copied** from the selected prompt. The Prompt is read only at the controller boundary — **no FK is stored** — and the lookup is scoped through `for(@type)` so the `Prompt::SCOPES` whitelist (enforced by `set_type`) is honoured. The first user `Message` carries the (Liquid-rendered, possibly edited) prompt text, then `request_reply!` fires. Responds with the session turbo frame.
  - `show` — transcript + composer; runs the `check_turbo_config` Redis ping and renders the "Redis is down" warning.
  - `index` — record-scoped session list; keeps the Roslin-disabled and prompts empty-state warnings.
- **`Projects::Sessions::MessagesController#create`** — appends a user message and re-opens the reply gate via `request_reply!`.
- **`Dradis::Plugins::Echo::SessionsChannel`** — a custom Turbo Streams channel (documented turbo-rails pattern). Turbo's default channel trusts any validly-signed stream name for its (non-expiring) lifetime, so a user who loses project access keeps receiving the transcript. This channel resolves the session behind the signed name and re-checks `Ability.new(current_user).can?(:use, session.record.project)` **at subscribe time**, rejecting revoked users. The view's `turbo_stream_from` gains `channel:` in Slice 5.
- **`RecordScoping`** concern — resolves records through the current project's collections so cross-project/record access raises `RecordNotFound` (mirrors `Projects::GrammarController`).

**Bug fixed (Slice 3 carryover):** `ReplyJob` defined a private `serialize(session, cutoff_id)` that shadowed `ActiveJob::Core#serialize` — the method every queue adapter calls when enqueuing. `perform_later` (and therefore `Session#request_reply!`) raised `NoMethodError` at runtime; Slice 3 never caught it because its specs stubbed `perform_later`. This slice is the first to enqueue for real. Renamed the private method to `finalize` and added an enqueue-path guard. Committed separately for review.

### Testing steps

1. Enable the Roslin agent (Echo → Agents) and configure a provider.
2. On an Issue, `POST` to `project_sessions_path` with `type=issue`, `record=<issue id>`, `prompt_id=<saved prompt id>`, and `prompt=<edited text>`. A new session + first user message are created, a reply is enqueued, and the response renders the session frame.
3. `POST` a message to the nested `messages` route — a user message is appended and a reply is enqueued.
4. Attempt either `POST` with an out-of-project record id — the request is denied (404).
5. Subscribe to `SessionsChannel` for a session you may `:use` — accepted and streaming. Revoke `:use` on the project and re-subscribe — rejected.

Automated coverage: request specs (sessions + messages, record scoping + cross-record denial), a channel spec that flips authorization between subscribe attempts, and the interactions/grammar suites (unchanged, still green). Full engine suite: 169 examples, 0 failures.

### Other Information

Local CI mirror all green: bundler-audit, ruby-audit, brakeman `-w2` (0 warnings), rubocop (changed files, 0 offenses), rspec (engine suite).

### Check List

- [ ] Added a CHANGELOG entry — N/A: no user-reachable behaviour yet (UI wired in Slice 5).

### DIFF
--- a/engines/dradis-echo/app/channels/dradis/plugins/echo/sessions_channel.rb
+++ b/engines/dradis-echo/app/channels/dradis/plugins/echo/sessions_channel.rb
@@ -40,8 +40,7 @@ module Dradis::Plugins::Echo
       name = verified_stream_name_from_params
       return unless name
 
-      record = GlobalID::Locator.locate(name.split(':').first)
-      record if record.is_a?(Session)
+      GlobalID::Locator.locate(name.split(':').first, only: Session)
     rescue ActiveRecord::RecordNotFound
       nil
     end
--- a/engines/dradis-echo/app/channels/dradis/plugins/echo/sessions_channel.rb
+++ b/engines/dradis-echo/app/channels/dradis/plugins/echo/sessions_channel.rb
@@ -1,10 +1,12 @@
 module Dradis::Plugins::Echo
   # Custom Turbo Streams channel for Echo sessions. Turbo's default channel
   # trusts any client holding a validly-signed stream name for the lifetime of
-  # that name — which never expires — so a user who loses project access keeps
-  # receiving the transcript. We re-check authorization at subscribe time
-  # instead: resolve the session behind the signed name and only stream if the
-  # subscriber may still :use its project, otherwise reject.
+  # that name — which never expires — so a user who loses project access could
+  # subscribe and keep receiving the transcript. We re-check authorization when
+  # a subscription is established: resolve the session behind the signed name
+  # and only stream if the subscriber may still :use its project, otherwise
+  # reject. Note this guards *new* subscriptions only; a connection opened while
+  # still authorized keeps streaming until it is torn down.
   #
   # Follows the documented turbo-rails custom-channel pattern
   # (Turbo::StreamsChannel).
--- a/engines/dradis-echo/app/channels/dradis/plugins/echo/sessions_channel.rb
+++ b/engines/dradis-echo/app/channels/dradis/plugins/echo/sessions_channel.rb
@@ -27,7 +27,7 @@ module Dradis::Plugins::Echo
     private
 
     def authorized?(session)
-      Ability.new(current_user).can?(:use, session.record.project)
+      Ability.new(current_user).can?(:use, session.project)
     end
 
     # The signed name encodes `[session, :messages]`, i.e.
--- a/engines/dradis-echo/app/channels/dradis/plugins/echo/sessions_channel.rb
+++ b/engines/dradis-echo/app/channels/dradis/plugins/echo/sessions_channel.rb
@@ -32,14 +32,17 @@ module Dradis::Plugins::Echo
 
     # The signed name encodes `[session, :messages]`, i.e.
     # "<session-gid-param>:messages". A tampered name fails verification and
-    # yields nil, which we treat as unauthorized.
+    # yields nil (never reaching locate), so the only raise path left is a
+    # session deleted between signing and subscribe — RecordNotFound, which we
+    # treat as unauthorized. Narrower than a blanket rescue so genuine bugs
+    # (NoMethodError etc.) still surface.
     def session_from_stream_name
       name = verified_stream_name_from_params
       return unless name
 
       record = GlobalID::Locator.locate(name.split(':').first)
       record if record.is_a?(Session)
-    rescue StandardError
+    rescue ActiveRecord::RecordNotFound
       nil
     end
   end
--- a/engines/dradis-echo/app/channels/dradis/plugins/echo/sessions_channel.rb
+++ b/engines/dradis-echo/app/channels/dradis/plugins/echo/sessions_channel.rb
@@ -1,0 +1,44 @@
+module Dradis::Plugins::Echo
+  # Custom Turbo Streams channel for Echo sessions. Turbo's default channel
+  # trusts any client holding a validly-signed stream name for the lifetime of
+  # that name — which never expires — so a user who loses project access keeps
+  # receiving the transcript. We re-check authorization at subscribe time
+  # instead: resolve the session behind the signed name and only stream if the
+  # subscriber may still :use its project, otherwise reject.
+  #
+  # Follows the documented turbo-rails custom-channel pattern
+  # (Turbo::StreamsChannel).
+  class SessionsChannel < ApplicationCable::Channel
+    extend Turbo::Streams::Broadcasts, Turbo::Streams::StreamName
+    include Turbo::Streams::StreamName::ClassMethods
+
+    def subscribed
+      session = session_from_stream_name
+
+      if session && authorized?(session)
+        stream_from verified_stream_name_from_params
+      else
+        reject
+      end
+    end
+
+    private
+
+    def authorized?(session)
+      Ability.new(current_user).can?(:use, session.record.project)
+    end
+
+    # The signed name encodes `[session, :messages]`, i.e.
+    # "<session-gid-param>:messages". A tampered name fails verification and
+    # yields nil, which we treat as unauthorized.
+    def session_from_stream_name
+      name = verified_stream_name_from_params
+      return unless name
+
+      record = GlobalID::Locator.locate(name.split(':').first)
+      record if record.is_a?(Session)
+    rescue StandardError
+      nil
+    end
+  end
+end
--- a/engines/dradis-echo/app/controllers/concerns/dradis/plugins/echo/record_scoping.rb
+++ b/engines/dradis-echo/app/controllers/concerns/dradis/plugins/echo/record_scoping.rb
@@ -1,0 +1,17 @@
+module Dradis::Plugins::Echo
+  # Shared record scoping for the session controllers. A session is always
+  # anchored to a record (an Issue or a Note) that must live inside the current
+  # project; resolving it through the project's collections gives us
+  # cross-project/record isolation for free — an out-of-scope id raises
+  # ActiveRecord::RecordNotFound, mirroring Projects::GrammarController.
+  module RecordScoping
+    extend ActiveSupport::Concern
+
+    private
+
+    def scoped_record(record)
+      collection = record.is_a?(Issue) ? current_project.issues : current_project.notes
+      collection.find(record.id)
+    end
+  end
+end
--- a/engines/dradis-echo/app/controllers/concerns/dradis/plugins/echo/record_scoping.rb
+++ b/engines/dradis-echo/app/controllers/concerns/dradis/plugins/echo/record_scoping.rb
@@ -4,14 +4,20 @@ module Dradis::Plugins::Echo
   # project; resolving it through the project's collections gives us
   # cross-project/record isolation for free — an out-of-scope id raises
   # ActiveRecord::RecordNotFound, mirroring Projects::GrammarController.
+  #
+  # We read the session's stored record_type/record_id directly rather than
+  # loading session.record: the association would fire a polymorphic query just
+  # to re-scope through the project. Issues are persisted as 'Issue' even though
+  # they descend from Note (see Session#record=), so the comparison matches the
+  # column exactly.
   module RecordScoping
     extend ActiveSupport::Concern
 
     private
 
-    def scoped_record(record)
-      collection = record.is_a?(Issue) ? current_project.issues : current_project.notes
-      collection.find(record.id)
+    def scoped_record(session)
+      collection = session.record_type == 'Issue' ? current_project.issues : current_project.notes
+      collection.find(session.record_id)
     end
   end
 end
--- a/engines/dradis-echo/app/controllers/concerns/dradis/plugins/echo/record_scoping.rb
+++ b/engines/dradis-echo/app/controllers/concerns/dradis/plugins/echo/record_scoping.rb
@@ -8,16 +8,15 @@ module Dradis::Plugins::Echo
   # We read the session's stored record_type/record_id directly rather than
   # loading session.record: the association would fire a polymorphic query just
   # to re-scope through the project. Issues are persisted as 'Issue' even though
-  # they descend from Note (see Session#record=), so the comparison matches the
-  # column exactly.
+  # they descend from Note (see Session#record=), so record_type maps straight
+  # onto the matching project collection (issues, notes, ...).
   module RecordScoping
     extend ActiveSupport::Concern
 
     private
 
     def scoped_record(session)
-      collection = session.record_type == 'Issue' ? current_project.issues : current_project.notes
-      collection.find(session.record_id)
+      current_project.public_send(session.record_type.underscore.pluralize).find(session.record_id)
     end
   end
 end
--- a/engines/dradis-echo/app/controllers/concerns/dradis/plugins/echo/turbo_config_check.rb
+++ b/engines/dradis-echo/app/controllers/concerns/dradis/plugins/echo/turbo_config_check.rb
@@ -3,10 +3,11 @@ module Dradis::Plugins::Echo
   # Turbo Streams only work if Action Cable can reach its backend, so the view
   # can warn the user when it can't.
   #
-  # Only the Redis adapter needs a reachable external server, so we ping just
-  # that one and memoize the result. Every other adapter (async in development,
-  # test in specs) is always treated as healthy — no spurious "can't contact
-  # Redis" alert and no per-request Redis round-trip.
+  # Only a Redis-backed adapter needs a reachable external server, so we ping
+  # just those (duck-typed on the subscription connection) and memoize the
+  # result. Every other adapter (async in development, test in specs) is always
+  # treated as healthy — no spurious "can't contact Redis" alert and no
+  # per-request Redis round-trip.
   module TurboConfigCheck
     extend ActiveSupport::Concern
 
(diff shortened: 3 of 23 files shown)


etdsoft:
### SEC-514 — PR #1655 code-review fixes (Slice 4: `echo/add-session-controllers`)

Applies the review findings introduced on the session-controllers slice (rebased onto the updated Slice 3 branch, force-pushed).

- **#7 Blank-prompt 500:** `sessions#create` now saves without a bang and returns **422** on an invalid (blank) prompt, mirroring the messages endpoint, instead of raising `RecordInvalid`. (The deferred-reply behaviour from the base branch is preserved — no `request_reply!` on create.)
- **#9 Dead `sessions#index`:** dropped the unused `:index` action and route; nothing links to it (back-link and lazy frame use the interactions path). The dead index **view** is deleted in Slice 5a.
- **#10 `check_turbo_config`:** extracted a shared `TurboConfigCheck` concern that guards on the **cable adapter class** (pings only the Redis adapter) and memoizes — async/test adapters no-op silently, so no spurious "can't contact Redis" alert and no per-request Redis round-trip. (Also adopted by `interactions_controller` in Slice 6.)
- **#11 (comment-only):** tightened the `SessionsChannel` class comment to say the re-auth guards **new** subscriptions only.

Verification: sessions / replies / messages / channel / concern / reply-job specs green locally (33 examples); `rubocop` clean on changed files. Feature specs need QA (no Firefox locally).

etdsoft:
@MattBudz — this PR (Echo Slice 4: add session controllers) has been open for >5 days without activity. Any blockers? Can we get this merged?

engines/dradis-echo/app/channels/dradis/plugins/echo/sessions_channel.rb
```diff
@@ -0,0 +1,46 @@
...
+    # The signed name encodes `[session, :messages]`, i.e.
+    # "<session-gid-param>:messages". A tampered name fails verification and
+    # yields nil, which we treat as unauthorized.
+    def session_from_stream_name
+      name = verified_stream_name_from_params
+      return unless name
+
+      record = GlobalID::Locator.locate(name.split(':').first)
+      record if record.is_a?(Session)
+    rescue StandardError
```
aapomm: What about a narrower catch:
```suggestion
    rescue ActiveRecord::RecordNotFound
```

engines/dradis-echo/app/controllers/concerns/dradis/plugins/echo/record_scoping.rb
```diff
@@ -0,0 +1,17 @@
...
+  # cross-project/record isolation for free — an out-of-scope id raises
+  # ActiveRecord::RecordNotFound, mirroring Projects::GrammarController.
+  module RecordScoping
+    extend ActiveSupport::Concern
+
+    private
+
+    def scoped_record(record)
+      collection = record.is_a?(Issue) ? current_project.issues : current_project.notes
+      collection.find(record.id)
```
aapomm: Instead of querying the exact instance that we already have, we pass the session instead and fetch the ID from the collection:

```suggestion
      collection = session.record_type == 'Issue' ? current_project.issues : current_project.notes
      collection.find(session.record_id)
```

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    private
+
+    def record_params
+      params.permit(:id, :prompt, :prompt_id, :project_id, :record, :type)
+    end
+
+    def set_record
+      raise ActiveRecord::RecordNotFound if @type.blank?
```
aapomm: I believe the `@type` is only ever gonna come from the `set_type` callback. What about raising the error there instead?

engines/dradis-echo/app/controllers/concerns/dradis/plugins/echo/record_scoping.rb
```diff
@@ -0,0 +1,23 @@
...
+  # to re-scope through the project. Issues are persisted as 'Issue' even though
+  # they descend from Note (see Session#record=), so the comparison matches the
+  # column exactly.
+  module RecordScoping
+    extend ActiveSupport::Concern
+
+    private
+
+    def scoped_record(session)
+      collection = session.record_type == 'Issue' ? current_project.issues : current_project.notes
```
caitmich: This isn't cleanly extensible if we add other records in the future. What about
```suggestion
      record_type = session.record_type.underscore.pluralize
      collection = current_project.public_send(record_type).find(session.record_id)
```

engines/dradis-echo/app/channels/dradis/plugins/echo/sessions_channel.rb
```diff
@@ -0,0 +1,49 @@
...
+    # yields nil (never reaching locate), so the only raise path left is a
+    # session deleted between signing and subscribe — RecordNotFound, which we
+    # treat as unauthorized. Narrower than a blanket rescue so genuine bugs
+    # (NoMethodError etc.) still surface.
+    def session_from_stream_name
+      name = verified_stream_name_from_params
+      return unless name
+
+      record = GlobalID::Locator.locate(name.split(':').first)
+      record if record.is_a?(Session)
```
caitmich:
```suggestion
      GlobalID::Locator.locate(name.split(':').first, only: Session)
```
https://www.rubydoc.info/gems/globalid/GlobalID/Locator#locate-class_method

engines/dradis-echo/app/controllers/concerns/dradis/plugins/echo/turbo_config_check.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def check_turbo_config
+      return @turbo_status if defined?(@turbo_status)
+
+      @turbo_status = turbo_backend_reachable?
+    end
+
+    def turbo_backend_reachable?
+      adapter = ActionCable.server.pubsub
+      return true unless adapter.is_a?(ActionCable::SubscriptionAdapter::Redis)
```
caitmich:
```suggestion
      return true unless adapter.respond_to?(:redis_connection_for_subscriptions)
```
What about checking that it responds rather than matching against the specific class name? More future-proof

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions/messages_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    include ProjectScoped
+    include RecordScoping
+    layout false
+
+    before_action :set_session
+
+    # Appends a user turn and re-opens the reply gate. The new message
+    # broadcasts itself into the transcript, so there's nothing to render back.
+    def create
+      message = @session.messages.build(content: params[:content], role: :user, user: current_user)
```
caitmich: Can we add a callback in the `Message` model that sets role to `:user` if a user is passed? That way we don't have to remember to pass both every time

caitmich: Can we validate the content with something like `params.expect`?
```ruby
params.expect(:content)
```
https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-expect

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions/messages_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+        head :ok
+      else
+        head :unprocessable_entity
+      end
+    end
+
+    private
+
+    def set_session
+      @session = Session.find(params[:session_id])
```
caitmich: Two things are odd here:

1. `scoped_record(@session)` isn't being assigned to anything, so I assume this is meant to throw an exception if it's not found? If that's the case, it's not super clear and maybe a comment or a more explicit way of handling this would be better?
2. This method is called `set_session` but it's doing more than that, it's also validating the record. Can we split it or make it clearer what this does?

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions/replies_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    def create
+      @session.request_reply! if @session.reply_pending?
+      head :ok
+    end
+
+    private
+
+    def set_session
+      @session = Session.find(params[:session_id])
+      scoped_record(@session)
```
caitmich: Since everywhere `set_session` is defined we're also validating the record via `scoped_record` why not include this in the concern and make it more general? Like `HasSession` or something like that

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions/replies_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    before_action :set_session
+
+    # Starts generation for a freshly-created session. The session Stimulus
+    # controller POSTs here from `connect`, i.e. after the browser has subscribed
+    # to the SessionsChannel, so the streaming container ReplyJob broadcasts lands
+    # on a listening socket (SEC-506 Bug 4). Guarded by reply_pending? so a
+    # reconnect, a second viewer, or a stray POST on an already-answered session
+    # can never spawn an unsolicited reply — request_reply! only fires while a
+    # reply is genuinely owed.
+    def create
```
caitmich, line 14: Isn't the resource here _technically_ a message? Wouldn't it be more RESTful to have a messages#reply action?

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+
+    private
+
+    def record_params
+      params.permit(:id, :prompt, :prompt_id, :project_id, :record, :type)
+    end
+
+    def set_record
+      @record = current_project.send(@type.to_s.pluralize).find(record_params[:record])
```
caitmich, line 66: This is overwritten on L67

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+    before_action :set_type, only: [:create, :index]
+    before_action :set_record, only: [:create, :index]
+    before_action :check_turbo_config, only: [:show, :create]
+    before_action :set_session, only: [:show]
+
+    # Lists the record's past sessions plus the prompts available to start a new
+    # one. Roslin-disabled and prompts empty-state warnings are handled by the
+    # view.
+    def index
+      @sessions = Session.for_record(@record)
```
caitmich, line 18: What about renaming this scope to something like `record_scoped`? `Session.for_record` sounds like we're creating a session based on other naming conventions we have in the app like `from_rtp` etc

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+    def create
+      prompt = current_user.prompts.for(@type).find(params[:prompt_id])
+
+      @session = Session.new(
+        agent: Agents::Roslin.instance,
+        record: @record,
+        title: prompt.title,
+        user: current_user
+      )
+      @session.messages.build(content: params[:prompt], role: :user, user: current_user)
```
caitmich: Validate content?

engines/dradis-echo/app/views/dradis/plugins/echo/projects/sessions/index.html.erb
```diff
@@ -0,0 +1,47 @@
+<% if Dradis::Plugins::Echo::Agents::Roslin.enabled? %>
+  <% if @prompts.any? %>
+    <% if @sessions.any? %>
+      <div class="list-group mb-3" id="echo-sessions" data-behavior="echo-sessions">
+        <% @sessions.each do |session| %>
+          <%= link_to session.title.presence || "Session ##{session.id}",
+                echo.project_session_path(current_project, session, type: @type, record: @record.id),
```
caitmich: Do we need both record id and type? Since we're validating `type` in the controller can we get one from the other?

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+        render :show
+      else
+        head :unprocessable_entity
+      end
+    end
+
+    private
+
+    def record_params
+      params.permit(:id, :prompt, :prompt_id, :project_id, :record, :type)
```
caitmich, line 62: `record_id` would be better since we're not passing the whole record

engines/dradis-echo/app/views/dradis/plugins/echo/projects/sessions/show.html.erb
```diff
@@ -0,0 +1,34 @@
+<turbo-frame id="<%= dom_id(@record, :echo) %>">
+  <% unless @turbo_status %>
```
caitmich, line 2: What about extracting this entire block to a shared partial and we can include it wherever we need in future with a single render partial call? I can foresee us needing this in more places down the line

engines/dradis-echo/app/views/dradis/plugins/echo/projects/sessions/show.html.erb
```diff
@@ -0,0 +1,34 @@
...
+  <% end %>
+
+  <%# Live transcript + composer state. Slice 5 adds `channel: SessionsChannel` %>
+  <%# so the socket re-authorizes on subscribe. %>
+  <%= turbo_stream_from [@session, :messages] %>
+  <%= turbo_stream_from [@session, :composer_state] %>
+
+  <div id="echo-messages" data-behavior="echo-messages">
+    <%= render partial: 'dradis/plugins/echo/projects/sessions/messages/message',
+               collection: @session.messages.order(:created_at, :id), as: :message %>
```
caitmich: why not set @collection in the controller?

engines/dradis-echo/config/routes.rb
```diff
@@ -10,6 +10,11 @@
 
       resources :grammar_corrections, only: [:create], controller: 'projects/grammar_corrections'
       resources :grammar_suggestions, only: [:create], controller: 'projects/grammar_suggestions'
+
```
caitmich, line 14: Now that we have so many controllers under `projects/` it looks like it's time for something like
```ruby
scope module: 'projects' do
```

engines/dradis-echo/app/views/dradis/plugins/echo/projects/sessions/show.html.erb
```diff
@@ -0,0 +1,34 @@
+<turbo-frame id="<%= dom_id(@record, :echo) %>">
+  <% unless @turbo_status %>
+    <div class="alert alert-danger">
+      There was an error contacting Redis: please make sure the Redis server is running.
+    </div>
+  <% end %>
+
+  <%# Live transcript + composer state. Slice 5 adds `channel: SessionsChannel` %>
+  <%# so the socket re-authorizes on subscribe. %>
```
caitmich:
```suggestion
  <%# Live transcript + composer state. %>
```

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions/messages_controller.rb
etdsoft: validate, what?

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions/messages_controller.rb
etdsoft: Isn't `set_` a common idiom in our controllers that set and validate form user params?

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions/replies_controller.rb
etdsoft: potato potato I think. I'm choosing to interpret RESTful as "adhering to CRUD", hence the extraction of a Reply resource that can be reasoned about, rather than tucking additional actions (e.g. "reply") to existing resources

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/sessions_controller.rb
etdsoft: I'm not certain a different name is better. "record scoped" signifies it's a scope for this record, which I think "for_record" kind of already does...

The minions also noticed:
>same pattern as Notification.for_user

PR #1658, opened 2026-07-22
add proactive edit locking for issues

MattBudz:
### Summary

Adds proactive edit locking for Issues: when a user opens the edit form for an issue that another user already has open, they see a lockout screen instead of the form, with the option to go back or edit anyway. This replaces the previous approach of only warning about conflicts after save.

### Testing steps

1. Log in to Dradis as a user with author access to a project, in Browser A.
2. Navigate to a project and open the Issues section.
3. Create a new issue (or use an existing one) and note its title.
4. Open a second, separate browser (or a private/incognito window) as Browser B, and log in as a different user.
5. In Browser A, click Edit on the issue.
6. Assert Browser A loads the normal "Edit issue" form.
7. In Browser B, navigate to the same issue and click Edit.
8. Assert Browser B shows a "This issue is currently being edited" lockout screen instead of the edit form.
9. Assert the lockout screen displays the name of the user editing in Browser A.
10. In Browser B, click Go back.
11. Assert Browser B returns to the issue's show page (not the edit form).
12. In Browser B, click Edit again, then click Edit anyway on the lockout screen.
13. Assert Browser B now shows the edit form for the issue.
14. In Browser A, make a change to the issue text and click Update Issue.
15. Assert Browser A shows the "Issue updated." confirmation and the change is saved.
16. In a new browser tab, log in as the Browser A user and click Edit on the same issue.
17. Assert the edit form loads normally (the lock was released after Browser A's save, so no lockout screen appears).
18. Wait for more than 24 hours since a session was acquired (or, in a test/staging environment, adjust the system clock forward), then have a new user attempt to edit a previously-locked record.
19. Assert the edit form loads normally (a stale lock older than 24 hours no longer blocks editing).

### Check List

- [x] Added a CHANGELOG entry


db/migrate/20260722100000_create_editing_sessions.rb
```diff
@@ -0,0 +1,11 @@
+class CreateEditingSessions < ActiveRecord::Migration[8.0]
+  def change
+    create_table :editing_sessions do |t|
+      t.references :user, null: false, foreign_key: true
+      t.references :record, polymorphic: true, null: false
+      t.datetime :started_at, null: false, precision: nil, default: -> { 'CURRENT_TIMESTAMP' }
```
aapomm: Can we use `created_at` here instead?

app/models/editing_session.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  belongs_to :user
+  belongs_to :record, polymorphic: true
+
+  validates :record_type, presence: true, inclusion: { in: ALLOWED_RECORD_TYPES }
+
+  scope :active, -> { where(started_at: stale_after.ago..) }
+  scope :by_others, ->(user) { where.not(user: user) }
+  scope :for_record, ->(record) {
+    where(record_type: record.class.name, record_id: record.id)
```
aapomm: I saw this same pattern in echo. In theory,
```
where(record: record)
```
works here except for the Issue/Note inheritance. If that's the case, let's add a comment about that here.

app/models/editing_session.rb
```diff
@@ -0,0 +1,28 @@
...
+  }
+  scope :stale, -> { where(started_at: ...stale_after.ago) }
+
+  def self.acquire(record_type:, record_id:, user:)
```
aapomm:
```suggestion
  def self.acquire!(record_type:, record_id:, user:)
```

app/models/editing_session.rb
MattBudz: Comment added

db/migrate/20260722100000_create_editing_sessions.rb
MattBudz: Renamed to created_at and let Rails manage it automatically instead of a manual default, in 2854a07bb.

app/models/editing_session.rb
MattBudz: Confirmed, Issue is an STI subclass of Note so where(record: record) would look up record_type: 'Note' instead of 'Issue' (verified with a quick console check). Added an explanatory comment in 7977b4536.

app/models/editing_session.rb
MattBudz: Applied in f89731917.

app/assets/javascripts/hera.js
```diff
@@ -57,6 +57,7 @@
 //= require hera/modules/liquid_async
 //= require hera/modules/nodes
 //= require hera/modules/project_board_colors
+//= require hera/modules/release_edit_lock
```
etdsoft: record_locking? "release_edit_lock" is an action name, no?

app/controllers/concerns/edit_lockable.rb
```diff
@@ -0,0 +1,34 @@
+module EditLockable
```
etdsoft: why "Edit", isn't Lockable good enough?

You have "locable_record" which is also a good one LocableResource <- if it's a controller concern?

app/controllers/concerns/edit_lockable.rb
```diff
@@ -0,0 +1,34 @@
...
+  def acquire_edit_session(record)
+    EditingSession.acquire!(record_type: record.class.name, record_id: record.id, user: current_user)
+  end
+
+  def check_edit_lock
+    record = lockable_record
+    competing_sessions = EditingSession.for_record(record).active.by_others(current_user)
+
+    if competing_sessions.any? && params[:force] != 'true'
+      @locked_by = competing_sessions.includes(:user).map(&:user)
```
etdsoft: shouldn't this `includes` be in L12?

app/controllers/concerns/edit_lockable.rb
```diff
@@ -0,0 +1,34 @@
...
+      @locked_record = record
```
etdsoft: this is weird, variable = variable, like a rename, is it because you don't want it if not locked? then @record and @lockable can do it. But also @record is probably loaded by the controller (as @issue, @evidence, or what not)...

Reading forward, there it is in L25

app/controllers/concerns/edit_lockable.rb
```diff
@@ -0,0 +1,34 @@
...
+      @back_path = url_from(request.referer) || root_path
```
etdsoft: this is different from how we handle other server-side "back" redirects...

app/controllers/concerns/edit_lockable.rb
```diff
@@ -0,0 +1,34 @@
...
+      return
+    end
+
+    acquire_edit_session(record)
+  end
+
+  def lockable_record
+    @lockable_record ||=
+      instance_variable_get("@#{controller_name.singularize}") ||
+        send("set_or_initialize_#{controller_name.singularize}")
```
etdsoft: uh oh... under what conditions do we need this plumbing? Seems we're reaching for functionality that we shouldn't be... if the record isn't set as an instance variable, shouldn't we BOOM loudly to let the developer know?

app/controllers/issues/editing_sessions_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Issues::EditingSessionsController < AuthenticatedController
+  include EditLockable
+  include ProjectScoped
```
etdsoft, line 2: why not inherit from IssuesController so you get this and #issue for free?

app/models/configuration.rb
```diff
@@ -15,6 +15,11 @@ class Configuration < ApplicationRecord
   # -- Class Methods --------------------------------------------------------
   # --------------------------------------------------------------- Misc admin:
 
+  def self.editing_session_stale_after
+    create_with(value: 1440)
```
etdsoft: what unit is this? Session Timeout in L28 is just "15". If it's minutes, what about a comment that explains what 1440 means to us humans?

app/models/editing_session.rb
```diff
@@ -0,0 +1,31 @@
+class EditingSession < ApplicationRecord
+  ALLOWED_RECORD_TYPES = %w[Issue].freeze
```
etdsoft: This is inverted, shouldn't a concern turn a model into Lockable, which in turns gives that record the ability to have EditingSessions.

And the concern in turn registers the "allowed record types" as we do elsewhere?

app/views/issues/_form.html.erb
```diff
@@ -37,15 +37,15 @@
...
   <div class="form-actions sticky">
     <%= render partial: 'qa/state_button', locals: { f: f, record: @issue } %>
+    or
+    <%=
+      link_to 'Cancel',
+      @form_cancel_path,
+      class: 'cancel-link',
+      data: {
+        behavior: class_names('clear-local-auto-save', 'release-edit-lock': @issue.persisted?),
+        'release-edit-lock-path': @issue.persisted? ? project_issue_editing_session_path(current_project, @issue) : nil
```
etdsoft: "release-edit-lock" seems like the wrong name? A clue is that the path is actually a Issue::EditingSession path, not a "release lock" path...

app/controllers/issues/editing_sessions_controller.rb
aapomm: There's quite a bit of difference between this controller and the IssuesController:
- `params[:id]` isn't used here but `params[:issue_id]`
- 10 unused concerns from `IssuesController`

This is already consistent with the rest of the sub issues controller:  `Issues::NodesController` and `Issues::EvidenceController`

app/models/configuration.rb
aapomm: Added comment
```
  # This settings is in minutes. The default value is 1440 minutes = 24 hours.
```

app/controllers/concerns/lockable_resource.rb
```diff
@@ -0,0 +1,24 @@
+module LockableResource
```
caitmich, line 1: since a concern is included in a controller it implies that `Lockable` applies to the resource. Can we name this just `Lockable`?

app/models/concerns/lockable.rb
```diff
@@ -0,0 +1,29 @@
...
+  extend ActiveSupport::Concern
+
+  mattr_accessor :allowed_types, default: []
+
+  included do |base|
+    Lockable.allowed_types << base.name
+
+    has_many :editing_sessions, as: :record
+
+    # We can't rely on `dependent: :destroy` above: for models that descend
```
caitmich, line 12: Should we add a `FIXME Issue/Note inheritance` here for consistency

app/models/configuration.rb
```diff
@@ -15,6 +15,12 @@ class Configuration < ApplicationRecord
   # -- Class Methods --------------------------------------------------------
   # --------------------------------------------------------------- Misc admin:
 
+  # This settings is in minutes. The default value is 1440 minutes = 24 hours.
+  def self.editing_session_stale_after
```
caitmich: what about `editing_session_expiry` or `editing_session_timeout` I think one of those would match the rest of the configuration names better

app/models/editing_session.rb
```diff
@@ -0,0 +1,31 @@
...
+  def self.acquire!(record_type:, record_id:, user:)
+    purge_stale_for(record_type: record_type, record_id: record_id)
+    create_or_find_by!(record_type: record_type, record_id: record_id) { |session| session.user = user }
+  end
+
+  def self.purge_stale_for(record_type:, record_id:)
+    where(record_type: record_type, record_id: record_id).stale.destroy_all
+  end
+
+  def self.stale_after
```
caitmich: ditto re naming

app/models/editing_session.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  scope :active, -> { where(created_at: stale_after.ago..) }
+  scope :by_others, ->(user) { where.not(user: user) }
+  # `where(record: record)` won't work here: Issue is an STI subclass of Note,
+  # so Rails' polymorphic query would look up record_type: 'Note' (the base
+  # class), not 'Issue' (what we actually store, see .acquire! below).
+  scope :for_record, ->(record) {
+    where(record_type: record.class.name, record_id: record.id)
+  }
+  scope :stale, -> { where(created_at: ...stale_after.ago) }
```
caitmich: It doesn't look like we're using this anywhere other than on L25, is it necessary?

app/models/issue.rb
```diff
@@ -8,6 +8,9 @@ class Issue < Note
   Commentable.allowed_types << 'Issue'
   InlineCommentable.allowed_types << 'Issue'
 
+  include Lockable
+  Lockable.allowed_types << 'Issue'
```
caitmich: Can we loop through all of these and add 'Issue' to each one rather than repeating it 4x?

app/views/shared/edit_locked.html.erb
```diff
@@ -0,0 +1,40 @@
...
+    <h3 class="heading">
+      This <%= @locked_record.model_name.human.downcase %> is currently being edited
+    </h3>
+
+    <p class="description">
+      Another team member is currently editing this record. Go back, or edit
+      anyway and risk overwriting their changes.
+    </p>
+
+    <% @locked_by.each do |user| %>
```
caitmich: It seems like this state (having multiple users in `@locked_by`) is unreachable. What's the intention?

config/routes.rb
```diff
@@ -79,6 +79,7 @@
       resources :evidence, concerns: :multiple_destroy, controller: 'issues/evidence', only: [:index, :new]
       resources :nodes, only: [:show], controller: 'issues/nodes'
       resources :revisions, only: [:index, :show]
+      resource :editing_session, only: :destroy, controller: 'issues/editing_sessions'
```
caitmich: Seems like a good candidate for a route concern based on how the rest of this feature is designed

spec/features/edit_locking_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    Configuration.find_or_create_by(name: 'admin:password')
+      .update!(value: BCrypt::Password.create(password))
+  end
+
+  def sign_in_as(username)
+    visit login_path
+    fill_in 'Username', with: username
+    fill_in 'Password', with: password
+    click_button 'Log in'
+  end
```
caitmich, line 17: We shouldn't be doing this in an individual spec, we have helpers for this

spec/features/edit_locking_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'rails_helper'
+
+describe 'Edit locking multi-actor flow' do
+  let(:project) { Project.new }
+  let(:issue) { create(:issue, node: project.issue_library) }
+  let(:password) { 'spec-password' }
+
+  before do
+    Configuration.find_or_create_by(name: 'admin:password')
```
caitmich: Why not `login_to_project_as_user`

.gitignore
```diff
@@ -43,3 +43,6 @@
 # Random
 .DS_Store
 *.swp
+
+# Local SQLite databases
+/db/*.sqlite3*
```
caitmich: The db files should be under `storage/` not `db/` https://github.com/dradis/dradis-ce/blob/963072245e5e1b8a7b44482dd9438febab13e969/config/database.yml.template#L14

CHANGELOG
```diff
@@ -1,6 +1,5 @@
 [v#.#.#] ([month] [YYYY])
-  - [entity]:
-    - [future tense verb] [feature]
+  - Issues: prevent editing a record while another user is already editing it
```
caitmich: "Warn user" is better since we're not preventing it

app/models/issue.rb
caitmich: Actually why can't the method in the concern handle this?

app/controllers/concerns/lockable_resource.rb
aapomm: `Lockable` is already a name used for the model concern so we can't use it here too.

spec/features/edit_locking_spec.rb
aapomm: This spec is unique and that it uses 2 different user sessions for testing the feature. We can't use `login_as_user` here since it simply stubs the auth and current_user methods.

spec/features/edit_locking_spec.rb
aapomm: ditto

.gitignore
aapomm: I've broaden this by removing the `db/` directory. This is a config anyway so users/devs can choose where to store the DB.

app/controllers/concerns/lockable_resource.rb
etdsoft: interesting!

app/controllers/concerns/lockable_resource.rb
```diff
@@ -0,0 +1,24 @@
...
+      @locked_record = lockable_record
+      render 'shared/edit_locked'
+      return
+    end
+
+    lockable_record.acquire_edit_session(current_user)
+  end
+
+  # Controllers that lock a record must return it here.
+  def lockable_record
```
etdsoft: but then, this should be _resource rather than _record no? or the module LockableRecord

app/models/editing_session.rb
```diff
@@ -0,0 +1,34 @@
+class EditingSession < ApplicationRecord
+  belongs_to :user
+  belongs_to :record, polymorphic: true
+
+  # Evaluated lazily: Lockable.allowed_types is populated as each model is
+  # loaded and includes the concern, which may happen after this class.
+  validates :record_type, presence: true, inclusion: { in: ->(_) { Lockable.allowed_types } }
+
+  # `find_by(record: record)` won't work here: Issue is an STI subclass of Note,
```
etdsoft, line 22:
```suggestion
  # FIXME - ISSUE/NOTE INHERITANCE
  # `find_by(record: record)` won't work here: Issue is an STI subclass of Note,
```

app/models/editing_session.rb
```diff
@@ -0,0 +1,34 @@
+class EditingSession < ApplicationRecord
```
etdsoft, line 1: This class is missing the section separator comments from other models.

It's also not sorting methods :alpha

.gitignore
```diff
@@ -43,3 +43,6 @@
 # Random
 .DS_Store
 *.swp
+
+# Local SQLite databases
+*.sqlite3*
```
etdsoft: this shouldn't be required if your files are where they should :L19, no? Pls remove

app/controllers/concerns/lockable_resource.rb
```diff
@@ -0,0 +1,24 @@
+module LockableResource
+  extend ActiveSupport::Concern
+
+  protected
+
+  def check_edit_lock
+    session = EditingSession.for_record(lockable_record)
```
etdsoft: is  this a safe var name? won't it mess with the app's session?

app/controllers/concerns/lockable_resource.rb
MattBudz: Renamed `lockable_record` to `lockable_resource` (and the issues_controller.rb implementation) to match the module name.

app/models/editing_session.rb
MattBudz: Fixed

.gitignore
MattBudz: removed

app/controllers/concerns/lockable_resource.rb
MattBudz: It didn't seem to cause issues, but I renamed the local to `editing_session` for clarity and extra safety.

PR #1660, opened 2026-07-22
replace pipe-value sniffing with explicit list field options

MattBudz:
### Summary

replace pipe-value sniffing with explicit list field options, keeping the pipe-value check as a fallback

### Testing steps

1. Login and open a project, then open any Node
2. Add a new Note
3. In the Note editor, click "Fields" to switch to the Fields tab
4. Click "Add field", enter a field name and a value, then switch to the "Source" tab
5. Assert the source text shows `#[Name]#` followed by the value you typed
6. Switch back to the "Fields" tab
7. Assert the field shows as a plain text value field (not a dropdown)
8. Save the Note
9. Assert the Note saves successfully and the field/value are shown correctly
10. Edit the Note again and confirm the Fields tab still renders correctly (add/remove a field, save again)
11. Assert no errors occur and the Note updates correctly
12. Go to Admin > Templates > Notes and create a new Note Template whose content includes a field with a pipe-separated value, e.g. `#[OS]#\nLinux | OSX | Windows`
13. Create a new Note from that template (`?template=<name>`)
14. Assert the OS field renders as a `<select>` with Linux/OSX/Windows as options, none selected
15. Select one of the options (e.g. "Linux"), then save the Note
16. Edit the Note again; assert the OS field now renders as a plain textarea containing "Linux"
17. Repeat steps 12-14, but save without selecting an option this time
18. Edit that Note again; assert the OS field still renders as a `<select>` with nothing selected (the raw pipe-separated value is preserved, not silently resolved to the first option)

### Other Information

This adds a generic, explicit `field_options` lookup (unused by CE itself, since CE has no List-field concept) as the primary mechanism, and keeps the original pipe-separated-value heuristic as a fallback when no explicit options are set for a field. This drops the old `allow_dropdown`/new-record distinction entirely: a dropdown is now driven purely by whether `field_options` or the pipe-fallback produce options, regardless of record state. The select always requires an explicit pick (`include_blank: true`): previously, an untouched pipe-separated Note Template default (e.g. `Linux | OSX | Windows`) would silently save as the raw joined string while the browser displayed the first option as if it had been chosen. Forcing a blank default keeps the raw value visibly unresolved until the user actually picks one.

The fallback is required because Note Templates (`config/templates/notes/*.txt`, managed under Admin > Templates > Notes) are a separate, RTP-independent CE/Pro feature that relies on hand-authored pipe-separated values to get a dropdown when creating a new record from a template. An earlier version of this PR removed the pipe-value check entirely, which broke this (caught by CI: `spec/features/issues_spec.rb` "preloads the editor with the template").

Confirmed via automated specs and manual browser testing that both paths (Note Templates and the new explicit lookup) work as expected.

### Check List

- [ ] ~~Added a CHANGELOG entry~~ (no user-visible CE change)


etdsoft (changes requested):
## Automated peer review -- PR #1660

**Verdict:** request-changes
**Findings:** 1 High, 0 Medium, 2 Low

### What I checked

### Both reviewers agreed

_(no shared findings)_

### Claude only

### F1 [Low|consistency] app/helpers/fields_helper.rb:12 -- helper methods not alphabetized within sections

**What:** In the public section, `render_dropdown?` (line 2) is followed by `dropdown_options` (line 12). In the private section, `list_field_options` (line 18) is followed by `has_liquid_filters?` (line 22). Both sections are reverse-alphabetical.

**Why:** `.claude/CLAUDE.md` → "Code style" → "Alphabetical order within sections" says method definitions must be sorted alphabetically within their logical section. The pre-existing `has_liquid_filters?` was previously the only private method, so ordering wasn't visible; this PR introduces `dropdown_options` and `list_field_options` and both land out of order.

**How to fix:** Swap the two public methods so `dropdown_options` is defined before `render_dropdown?`, and swap the two private methods so `has_liquid_filters?` is defined before `list_field_options`.

---

### F2 [Low|tests] spec/helpers/fields_helper_spec.rb:21 -- spec passes for the wrong reason

**What:** The example "returns false for a field with no matching options" sets `@field_options = { 'Risk' => ... }` and asserts `render_dropdown?('Status', 'Open')` is falsey. Since `@field_options['Status']` is nil, execution falls to the else-branch, which returns false only because `@allow_dropdown` was never assigned (nil). The assertion is not testing the "field_options entry missing" branch independently.

**Why:** If a future change defaulted `@allow_dropdown` to truthy, this test would silently start passing for a different reason (or failing without clearly indicating which branch broke).

**How to fix:** Either (a) add `assign(:allow_dropdown, true)` so the assertion is clearly depending on the field_options lookup (not the unset `@allow_dropdown`), or (b) rename the example to "falls back to the pipe-value check when the field has no explicit options" and add `assign(:allow_dropdown, true)` + a pipe-separated value to make the fallback path explicitly exercised.

---

### Codex only

### F3 [High|correctness] app/views/fields/_field.html.erb:24 -- blank list values silently submit as the first option

**What:** When `render_dropdown?` returns true for a blank explicit list-field value (CE's new path: `value.blank?` → true), the select renders `dropdown_options(field_name, value)` with `{ selected: value }` but no blank option. A browser submits the first `<option>` when no option matches the selected value.

**Why:** For a new record with an explicit list field (e.g. Risk = High/Medium/Low), opening the Fields tab and saving without touching the Risk field submits `"High"` -- silently writing a non-empty value into a field the user never touched. This is data-loss behavior: user intent (leave blank) is overwritten by browser default.

**How to fix:** Add `include_blank: value.blank?` to the options hash so a blank option is rendered (and pre-selected) when the current value is empty:

```erb
<%= select :item_form, "field_value_#{index}",
      dropdown_options(field_name, value),
      { selected: value, include_blank: value.blank? },
      { class: 'form-select', data: { behavior: 'preview-enabled' }, tabindex: index + 1 } %>
```

Add a helper/view spec that renders a blank explicit list field and asserts the submitted value is blank (not the first option) when the user does not interact.

---

### Notes

- CE never populates `data-field-options` on any textarea today (grep confirms), so `_fieldOptions` is `undefined` client-side and omitted from the POST -- the new High finding only manifests when Pro wires real RTP List field options. Still worth fixing here before the Pro follow-up ships.
- `select ... { selected: value }` for the legacy pipe path passes the full pipe string as the selected value, which matches none of the split options; the browser falls back to the first `<option>`, matching previous behavior. Not a new regression.

---

app/controllers/fields_controller.rb
```diff
@@ -3,6 +3,7 @@ class FieldsController < AuthenticatedController
   def form
     @form_data = FieldParser.source_to_fields_array(params[:source])
     @allow_dropdown = params[:allow_dropdown] == 'true'
+    @field_options = params.permit(field_options: {})[:field_options]&.to_h
```
etdsoft: looks like we need to make up our mind re: StrongParameters. I'd say Yes and do it throughout (e.g. `field_params`?)

app/helpers/fields_helper.rb
```diff
@@ -1,8 +1,13 @@
 module FieldsHelper
-  def render_dropdown?(value)
-    @allow_dropdown &&
-    !has_liquid_filters?(value) &&
-    value.split(' | ').count > 1
+  def dropdown_options(field_name, value)
```
etdsoft, line 2: it seems that we changed the name, but the actual "behavior" includes a bit of both (the render check and the options gathering), perhaps worth keeping them separate?

vendor/assets/javascripts/jquery.textile.js
etdsoft, line 1: changing the editor's internals wasn't in the tin, although I see we were already doing some of the stuff and you're just surfacing the code to the top... #nothingtoseehere

vendor/assets/javascripts/jquery.textile.js
```diff
@@ -56,6 +56,9 @@
       // Place initialization logic here
       // You already have access to the DOM element and the options via the instance,
       // e.g., this.element and this.options
+      this._fieldOptions = this.$element.data('field-options');
+      this._allowDropdown = this.$element.data('allow-dropdown');
```
etdsoft: two very different names for functionality that's very close (or actually, the same functionality/behavior). I wonder if there's a better naming convention here

app/controllers/fields_controller.rb
MattBudz: StrongParameters should be the default. I added a `field_params` method that permits everything the controller reads (`allow_dropdown`, `index`, `source`, `field_options`, `form`), and switched all three actions to read from it instead of `params` directly.

app/helpers/fields_helper.rb
MattBudz: This was actually two methods but @caitmich suggested consolidating them in the BC thread for this PR. The single call site in `_field.html.erb` also wants both the boolean and the options together. We can spit this back into two methods (where one calls the other) if you feel strongly about it.

vendor/assets/javascripts/jquery.textile.js
MattBudz: These names are deliberately paired with the Rails side: `data-field-options`/`data-allow-dropdown` → `_fieldOptions`/`_allowDropdown` → posted as `field_options`/`allow_dropdown` → read into `@field_options`/`@allow_dropdown` in `FieldsHelper`.

I checked whether `field_options` presence alone could replace `allow_dropdown` as the trigger but it looks like it can't (at least not here in this PR with bloating the scope). Notes and Cards only ever pass `allow_dropdown` and they never pass `field_options` at all (makes sense since no RTP fields).

etdsoft: This misses my point. Yes they're internally consistent in the front- and the back-end. But that doesn't address how they're using completely different naming conventions.

Before this PR, allow dropdown was only on new records (presumably from pipe-separated templates):
```
app/views/issues/_form.html.erb:          allow_dropdown: @issue.new_record?,
```

I don't see where the new :field_options :data is set.

MattBudz: gotcha, `allow_dropdown` has now been removed completely.

> I don't see where the new :field_options :data is set.

This is set in pro and used by pro only. I figured it makes sense to update this JS file here since this is under `vendor/` and it would likely get messy to maintain two different versions of the file. If you feel strongly about keeping this change to pro's vendored js, I can commit that.

PR #1662, opened 2026-07-23
Scrub invalid UTF-8 byte sequences before saving

MattBudz:
### Summary

Pasted or imported content can contain invalid UTF-8 byte sequences. Until now these only failed as an unhandled `ArgumentError`, an incidental side effect of the length validation calling `String#length` on invalid bytes rather than a deliberate check. Any write path that bypasses that validation could leave corrupt bytes in the DB, breaking every downstream consumer of the affected attribute (exporters, reports, API responses) instead of just the one that happened to trip over it first.

`ApplicationRecord` now scrubs every `:text` column before validation, so any model gets this for free without having to opt in. `:string` columns (`Card#name`, `Node#label`) still need an explicit `scrub_invalid_encoding_for` call via the new `ScrubsInvalidEncoding` concern, since most `:string` columns hold structured values (emails, tokens, slugs) that shouldn't be swept in automatically.

### Testing steps

1. Login as an admin user and open a Rails console in the same environment (`bin/rails console`) — invalid UTF-8 can't be typed through the browser, so this is required to construct the test data.
2. Run `node = Node.create!(label: 'QA Node')`.
3. Run `note = Note.new(text: "Bad byte: \xC3\x28", node: node, category: Category.default); note.save!`.
4. Assert step 3 completes without raising `ArgumentError: invalid byte sequence in UTF-8`.
5. Run `note.reload.text.valid_encoding?`.
6. Assert step 5 returns `true`.
7. Run `issue = Issue.new(text: "Bad byte: \xC3\x28", node: node.project.issue_library, category: Category.issue, author: 'qa'); issue.save!`.
8. Assert step 7 completes without raising an error, then run `issue.reload.text.valid_encoding?` and assert it returns `true`.
9. Run `evidence = Evidence.new(content: "Bad byte: \xC3\x28", node: node, issue: issue, author: 'qa'); evidence.save!`.
10. Assert step 9 completes without raising an error, then run `evidence.reload.content.valid_encoding?` and assert it returns `true`.
11. In the browser, create a Board and a List under a project, then in the console run `card = Card.new(name: "Bad byte: \xC3\x28", description: "Bad byte: \xC3\x28", list: List.last); card.save!`.
12. Assert step 11 succeeds, then run `card.reload.name.valid_encoding?` and `card.reload.description.valid_encoding?` and assert both return `true`.
13. Run `bad_node = Node.new(label: "Bad byte: \xC3\x28", parent_id: nil); bad_node.save!`.
14. Assert step 13 succeeds, then run `bad_node.reload.label.valid_encoding?` and assert it returns `true`.
15. In the browser, navigate to the project containing the Note/Issue created above.
16. Export the project as a package (Project > Export > Package Export).
17. Assert the export downloads successfully as a `.zip` file.
18. Create a new project and import the exported package (Import from file...).
19. Assert the import completes successfully and the new project contains the Note/Issue with their original text (previously, invalid UTF-8 in the exported XML caused Nokogiri to fail silently and the import produced an empty project with no visible error).

### Check List

- [x] Added a CHANGELOG entry


MattBudz:
closed in favor of https://github.com/dradis/dradis-projects/pull/107

PR #1675, opened 2026-08-13
Support email addresses in Textile inline code

nicolachr:
### Spec
Textile's inline-code syntax @text@ fails silently when the content contains an email address (e.g. @admin@starfleet.com@). RedCloth can't find a valid closing @ and leaves the raw text unformatted, with no error and no working escape.

**Proposed solution**
Before RedCloth parses the text, hide the inner @ of any email-shaped @...@ span behind a reserved Unicode sentinel character, leaving the span's opening/closing @ alone. RedCloth then sees an ordinary single-@ span and decides on its own whether to render it as code, leave it literal in a bc. block, or leave it literal wherever inline code is disabled (Comments). Once rendering is done, restoring is one unconditional find/replace of the sentinel back to @.

Emails inside URLs (bare, or "text":URL) are left untouched, since RedCloth already renders those correctly on its own, protecting them would split the URL into separate link/code fragments.

### Testing steps

1. Log in to Dradis and open (or create) a project.
2. Navigate to the project's Issues, and create a new Issue.
3. In the Issue's Description field, enter: `Contact @admin@starfleet.com@ for access.` and save the Issue.
4. Assert the rendered Issue shows "Contact `admin@starfleet.com` for access." with `admin@starfleet.com` displayed in a monospaced/code style (not as raw text with `@` symbols around it).
5. Export the project
6. Assert the rendered Issue shows "Contact `admin@starfleet.com` for access." in the export


etdsoft (changes requested):
## Automated peer review -- PR #1675

**Verdict:** request-changes
**Findings:** 1 High, 0 Medium, 1 Low

### What I checked

### Both reviewers agreed

Both sub-agents independently flagged the same root cause -- `placeholders.fetch(token)` raising `KeyError` on a placeholder-shaped user string -- but cited different call sites and assessed severity differently (Claude: Low at line 68; Codex: High at line 94). They don't match mechanically (line delta > 5, severity differs), so they appear separately below. The underlying concern is identical: both `restore_attribute_value` (line 68) and `restore_text_node` (line 94) call `.fetch` without a fallback, and both are vulnerable to the same collision.

### Claude only

### F1 [Low|correctness] lib/html/pipeline/dradis/textile_filter.rb:68 -- `placeholders.fetch(token)` raises `KeyError` on collision with user text

**What:** Both `restore_attribute_value` (line 68) and `restore_text_node` (line 95) call `placeholders.fetch(token)` with no default. `EMAIL_PLACEHOLDER_PATTERN` matches ANY `dradisemailcode` + 16 hex chars, including a string the user typed. If the input contains at least one real `@email@` span (so `placeholders` is non-empty) AND the user text also contains a literal string matching that shape (e.g. someone pastes `dradisemailcode0123456789abcdef` in an Issue/Note description), the gsub in `restore_text_node` will find the user's string, call `placeholders.fetch("dradisemailcode0123456789abcdef")`, and raise `KeyError` — crashing the render of that field for every viewer.

**Why:** The probability of a real user typing exactly `dradisemailcode` + 16 hex chars is astronomically low, so this is not an exploitable security issue in practice. But it is a real correctness gap: any input can shape-collide with the placeholder scheme, and the current code assumes 100% of matches came from `protect_email_inline_code`. `fetch` without a default turns a rendering edge case into an unhandled exception rather than a graceful passthrough of the user's original text.

**How to fix:** Use the block form of `fetch` that returns the original token unchanged, so a shape-collision falls through as literal text. Two lines:

```ruby
def restore_attribute_value(value, placeholders)
  value.gsub(EMAIL_PLACEHOLDER_PATTERN) { |token| placeholders.key?(token) ? "@#{placeholders[token]}@" : token }
end

def restore_text_node(text, placeholders, wrap_in_code)
  CGI.escapeHTML(text).gsub(EMAIL_PLACEHOLDER_PATTERN) do |token|
    next token unless placeholders.key?(token)
    email = placeholders[token]
    wrap_in_code ? "<code>#{email}</code>" : "@#{email}@"
  end
end
```

Optionally add one spec (`it 'passes through text that shape-collides with the placeholder scheme'`) so this doesn't regress.

### Codex only

### F1 [High|correctness] lib/html/pipeline/dradis/textile_filter.rb:94 -- crafted placeholder text can crash rendering

**What:** `restore_text_node` and `restore_attribute_value` call `placeholders.fetch(token)` for every `dradisemailcode[a-f0-9]{16}` substring found in the rendered HTML. If user content contains a literal token-shaped string plus at least one protected email span, the restore pass will match the literal string and raise `KeyError` because it is not in `placeholders`. For example, text like `dradisemailcodeaaaaaaaaaaaaaaaa @admin@starfleet.com@` can crash the page when the issue/note is rendered.

**Why:** Issue and note content is user-controlled and rendered through this shared pipeline. A single crafted note/issue can turn into a persistent render-time exception for anyone viewing it, so the parser must only restore tokens it generated for this render.

**How to fix:** Match only the actual placeholder keys created in `protect_email_inline_code`, not the global token shape. Build a per-render regexp and pass it through the restore helpers, e.g. `placeholder_pattern = Regexp.union(placeholders.keys)` after the empty check, then use that pattern in text nodes and attributes. Add a spec such as `source = 'literal dradisemailcodeaaaaaaaaaaaaaaaa and @admin@starfleet.com@'` and assert the literal token remains text while the email renders as code.

### Notes

Both sub-agents independently identified the same root cause. The fixes they propose are equivalent in spirit: either use `placeholders.key?(token)` to guard the lookup, or narrow the match pattern to only the keys in `placeholders`. Either approach closes the gap; the `Regexp.union(placeholders.keys)` approach is slightly more rigorous since it eliminates the pattern-match vs. key-lookup mismatch entirely. A test covering the collision case is missing and should be added.

No prompt injection attempts were detected in the PR content.

---

lib/html/pipeline/dradis/textile_filter.rb
```diff
@@ -1,3 +1,6 @@
+require 'cgi'
+require 'securerandom'
```
aapomm: Are these needed?

spec/lib/html/pipeline/dradis/textile_filter_spec.rb
```diff
@@ -7,4 +7,64 @@
...
+    result = '<div><p>Token dradisemailcodeaaaaaaaaaaaaaaaa and contact <code>admin@starfleet.com</code> for access.</p></div>'
+    expect(described_class.call(source, {}).to_s).to eq(result)
+  end
+
+  it 'leaves an email address inside a bare URL untouched so it stays one string for AutolinkFilter' do
+    source = 'http://evil.com/@admin@starfleet.com@/login'
+    result = '<div><p>http://evil.com/@admin@starfleet.com@/login</p></div>'
+    expect(described_class.call(source, {}).to_s).to eq(result)
+  end
+
```
aapomm, line 63: Let's add more unhappy paths:

`@unterminated@inline.code`
`@@multipleinline@@`
`@@@mismatched@@`

lib/html/pipeline/dradis/textile_filter.rb
```diff
@@ -15,17 +18,102 @@ module Dradis
...
           end
 
           "<div>#{doc}</div>"
         end
+
+        private
+
+        def protect_email_inline_code(text)
+          url_ranges = text.to_enum(:scan, BARE_URL_PATTERN).map { Regexp.last_match.begin(0)...Regexp.last_match.end(0) }
+          placeholders = {}
```
aapomm: What about saving `placeholders` and `placeholder_pattern` as instance variables? That way, we won't need to pass them as arguments across multiple methods

lib/html/pipeline/dradis/textile_filter.rb
nicolachr: both removed in updated approach since the sentinel approach does no random token generation or manual HTML escaping.

lib/html/pipeline/dradis/textile_filter.rb
nicolachr: Resolved by removing the need for them entirely. There's no more placeholder hash to pass around.

lib/html/pipeline/dradis/textile_filter.rb
```diff
@@ -15,16 +15,68 @@ module Dradis
...
+        # than raising.
+        EMAIL_AT_SENTINEL = "\uFDD0"
+
+        # A contiguous non-whitespace run containing '://' is a URL, whether
+        # bare or the target of Textile link syntax ("text":URL). RedCloth
+        # already handles an embedded '@...@' correctly in both cases on its
+        # own, so we leave those alone rather than protecting them: bare
+        # URLs stay one unbroken string for AutolinkFilter to link as a
+        # whole, instead of being split into link/code fragments by an
+        # email we protected and later restored as <code>.
```
aapomm: The comment is too much AI-speak
```suggestion
        # RedCloth leaves instances of '@...@' on links alone so the Autolinker
        # can do its thing. So for those cases, we don't perform the sentinel
        # character replacement. Otherwise, RedCloth will detect these as inline
        # code and add unnecessary <code></code> tags.
```

lib/html/pipeline/dradis/textile_filter.rb
```diff
@@ -15,16 +15,68 @@ module Dradis
...
+        # abandons the whole span or mis-parses it (e.g. a dotted local
+        # part like '@first.last@starfleet.com@' gets truncated). We avoid
+        # this entirely by hiding only the email's inner '@' behind a
+        # sentinel character before RedCloth ever sees it, leaving the
+        # span's opening and closing '@' untouched. RedCloth then parses an
+        # ordinary single-embedded-'@' span, so it decides on its own
+        # (correctly) whether to render <code>, leave it literal in a `bc.`
+        # block, or leave it literal with inline code disabled (comments) --
+        # we don't need to special-case any of that. We just swap the
+        # sentinel back for '@' once RedCloth is done.
```
aapomm: This comment explains the problem and why we're protecting emails at all. Let's move this as a comment to the `protect_email_inline_code` method instead.

lib/html/pipeline/dradis/textile_filter.rb
```diff
@@ -15,16 +15,68 @@ module Dradis
...
+        EMAIL_INLINE_CODE_PATTERN = /@([\w+\-.]+)@([\w-]+(?:\.[\w-]+)+)@/
```
aapomm: It doesn't look like this regex is covering the case: `@email@localhost@`. It also seems like we're reinventing the wheel when an email regex already exists: `URI::MailTo::EMAIL_REGEXP`. Why not just use that?

lib/html/pipeline/dradis/textile_filter.rb
```diff
@@ -15,16 +15,68 @@ module Dradis
...
+          parser = RedCloth.new(protect_email_inline_code(@text), [:filter_html, :no_span_caps])
 
           doc = if context[:no_inline_code]
             parser.to(HTML::NoInlineCodeTextileFormatter)
           else
             parser.to_html
           end
 
-          "<div>#{doc}</div>"
+          "<div>#{restore_email_inline_code(doc)}</div>"
```
aapomm: What about removing the restore method and just go with 
```suggestion
          "<div>#{doc.gsub(EMAIL_AT_SENTINEL, '@')}</div>"
```

lib/html/pipeline/dradis/textile_filter.rb
```diff
@@ -15,16 +15,68 @@ module Dradis
...
+        EMAIL_INLINE_CODE_PATTERN = /@([\w+\-.]+)@([\w-]+(?:\.[\w-]+)+)@/
+
+        # U+FDD0 is a Unicode noncharacter: the standard guarantees it can
+        # never legitimately appear in interchanged text, so it's safe to
+        # use as a marker with a plain global find/replace -- no lookup
+        # table, no risk of colliding with real content. On the vanishingly
+        # unlikely chance input already contains this exact codepoint, we
+        # degrade gracefully (it's rendered back as a literal '@') rather
+        # than raising.
+        EMAIL_AT_SENTINEL = "\uFDD0"
```
aapomm: What about
```suggestion
        SENTINEL_CHAR = "\uFDD0"
```

PR #1677, opened 2026-08-14
Add QA review state to Evidence

nicolachr:
### Summary

Evidence had no draft/ready_for_review/published workflow of its own, so reviewers could only track progress at the Issue level even though a single issue's evidence can be at different review stages simultaneously.

- Evidence gains an independent `state` enum, matching the existing `Note`/`ContentBlock`/IssueLib `Entry` pattern (migration backfills existing evidence to `published`).
- Extracted a `Reviewable` concern housing the shared `enum :state, [:draft, :ready_for_review, :published]`, now included by `Note` and `Evidence` (Pro's `ContentBlock` and IssueLib `Entry` include the same concern) instead of each model redefining the enum.
- State is editable everywhere Evidence is edited (Issue > Evidence tab, Node > Evidence tab), not just from QA.
- A new "Evidence" tab on the QA Issue page lists the issue's evidence that's ready for review, with the same bulk "mark selected as…" toolbar Issues/Content Blocks already have.
- A new QA-scoped Evidence show page (`QA::Issues::EvidenceController`) lets a reviewer open an evidence item, edit it, or use a sidebar "Update State" widget (mirroring the Issues QA widget) to quickly move it to Draft or Published — auto-advancing to the next evidence ready for review on the same issue, or back to the tab if none remain.
- Publishing is gated behind the existing reviewer/`:publish` permission via the `Publishable` concern, consistent with Issues/Content Blocks.

This is the base branch for a set of follow-up PRs that make report exporters and the Gateway client portal respect Evidence's state; those live in separate PRs (`Issue#evidence_by_node`'s scope param in #1684, `NodeDrop#evidence` unscoping in #1685) rather than here.

### Testing steps

1. Login as an admin/author user who is also a Reviewer on a project (Project > People > Reviewers includes this user).
2. Navigate to an existing Issue and add two pieces of Evidence to it, on two different Nodes.
3. Edit the first Evidence item (Node > Evidence tab > Edit, or Issue > Evidence tab > Edit). Assert the edit form now shows an "Update Evidence (Draft)" button with a dropdown offering Draft / Ready for review / Published.
4. Set the Evidence's state to "Ready for review" and save. Assert the page shows the evidence with its new state.
5. Repeat steps 3-4 for the second Evidence item, also setting it to "Ready for review".
6. Mark the parent Issue itself as "Ready for review".
7. Navigate to Project > QA and click the Issue in the "Issues for review" list.
8. On the Issue's QA page, click the "Evidence" tab. Assert both Evidence items are listed, each showing their Node label.
9. Select the checkbox for one Evidence row, click the "State" button that appears, then click "Published". Assert the page lands back on the Issue's QA page with the Evidence tab active, the navbar/sidebar/styling still present (not a bare, unstyled page), a "State updated successfully." message is shown, and the row is gone from the table.
10. Re-visit the QA issue page and click "Evidence" again. Assert only the remaining "Ready for review" Evidence item is listed.
11. Click the Node label link for the remaining Evidence item. Assert the URL is `/qa/issues/<id>/evidence/<id>` (not the regular Node > Evidence URL).
12. On this QA Evidence page, assert the breadcrumb reads: Project name > QA > Issue title > Node label.
13. Click the Issue title in the breadcrumb. Assert you land back on the Issue's QA page with the Evidence tab active.
14. Return to the QA Evidence page for the remaining item. In the sidebar, assert an "Update State" widget is visible with Draft / Review / Published buttons, and a list titled "Evidence for review" showing this item.
15. Create a third Evidence item on the same Issue, set it to "Ready for review", and open its QA Evidence page. Assert the sidebar now lists both remaining items.
16. Click "Published" in the sidebar widget for that item. Assert you're redirected to the other remaining evidence's QA show page (the "next" one ready for review).
17. Click "Published" again on that last item. Assert you're redirected back to the Issue's QA page with the Evidence tab active, now showing the empty state ("You don't have any evidence to review yet.").
18. Log in as (or switch to) a user who is not a Reviewer on the project. Navigate to the Issue's QA Evidence tab with at least one "Ready for review" item present. Select its checkbox, click "State". Assert "Published" is shown disabled with a tooltip ("You are not a Reviewer for this project."), while "Draft" remains clickable.
19. As the non-reviewer user, open that Evidence item's QA show page. Assert the "Published" button in the sidebar's Update State widget is shown disabled rather than clickable.

### Other Information

Built test-first (RSpec + Capybara/Selenium feature specs); full evidence/QA regression suite passing.

### Check List

- [x] Added a CHANGELOG entry


app/controllers/nodes_controller.rb
```diff
@@ -123,11 +123,11 @@ def set_columns
...
     rtp_default_evidence_fields = rtp ? rtp.evidence_fields.default.field_names : []
 
-    @note_columns     = note_dynamic_fields | extra_field_names
-    @evidence_columns = rtp_default_evidence_fields | evidence_dynamic_fields | extra_field_names
+    @note_columns = note_dynamic_fields | extra_field_names
+    @evidence_columns = rtp_default_evidence_fields | evidence_dynamic_fields | extra_field_names | ['State']
 
     @default_note_columns = default_field_names
-    @default_evidence_columns = rtp_default_evidence_fields.presence || default_field_names
+    @default_evidence_columns = (rtp_default_evidence_fields.presence || default_field_names) | ['State']
```
etdsoft: this is inconsistent with your own app/controllers/issues/evidence_controller.rb change. There, State was added to :default_field_names (L117)

app/helpers/evidence_helper.rb
```diff
@@ -1,7 +1,17 @@
 module EvidenceHelper
+  def evidence_edit_path(evidence, qa: false)
```
etdsoft: why is this needed? (no issue_edit_path helper)

app/models/evidence.rb
```diff
@@ -1,6 +1,8 @@
 class Evidence < ApplicationRecord
   include Commentable
+  include Eventable
```
etdsoft, line 3: This should have been a separate PR. Please extract, this workstream is heavy enough and we can merge quickly and independently. Right now it's neither here or there, the controllers still include Activity Tracking, and we're adding Eventable to the models.

app/models/evidence.rb
```diff
@@ -24,12 +26,29 @@ class Evidence < ApplicationRecord
 
   # -- Scopes ---------------------------------------------------------------
 
-
   # -- Class Methods --------------------------------------------------------
 
-
   # -- Instance Methods -----------------------------------------------------
 
+  def local_event_payload
```
etdsoft, line 33: for the extracted PR: missing :fields

app/views/evidence/_form.html.erb
```diff
@@ -34,14 +34,15 @@
...
+    <%= render partial: 'qa/state_button', locals: { f: f, record: @evidence } %>
   </div>
+  or
+  <%=
+    link_to 'Cancel',
+    evidence_redirect_path(params[:return_to]),
+    data: {
+      behavior: 'clear-local-auto-save'
+    }
+  %>
```
etdsoft: In all other forms, Cancel stays within the .sticky :div

app/views/issues/evidence/_table.html.erb
```diff
@@ -15,15 +17,20 @@
   </thead>
   <tbody>
     <% @affected_nodes.each do |node| %>
-      <% node.evidence.where(issue_id: @issue.id).each do |evidence| %>
+      <% node_evidence = node.evidence.where(issue_id: @issue.id)
+         node_evidence = node_evidence.ready_for_review if local_assigns[:qa] %>
+      <% node_evidence.each do |evidence| %>
```
etdsoft: 3 lines overwriting the same variable seems a bit much for view code, why can't this live in the :controller ?

Is :local_assigns needed vs plain call to :qa on the partial render call?

app/views/issues/evidence/_table.html.erb
```diff
@@ -15,15 +17,20 @@
...
         <tr id="evidence-<%= evidence.id %>">
           <td class="select-checkbox" data-behavior="select-checkbox"></td>
           <% @all_columns.each do |column| %>
             <%
               sort, display =
               case column
               when 'Label'
-                [evidence.node.label, link_to(evidence.node.label, [current_project, node, evidence])]
+                label_path = local_assigns[:qa] ? project_qa_issue_evidence_path(current_project, @issue, evidence) : [current_project, node, evidence]
+                [evidence.node.label, link_to(evidence.node.label, label_path)]
```
etdsoft: this is the same sort of code you have in the Helper for the :edit action, but here you include it in-view.

Between this, L47 + Helper, and L20-22, I'm not convinced that reusing the same partial is buying us what we want. Is there a way to simplify it?

app/views/issues/evidence/_table.html.erb
```diff
@@ -37,7 +44,7 @@
             <td data-sort="<%= sort %>"><%= display %></td>
           <% end  %>
           <td class="column-actions">
-            <%= link_to edit_project_node_evidence_path(current_project, node, evidence, return_to: :issue) do %>
+            <%= link_to evidence_edit_path(evidence, qa: local_assigns[:qa]) do %>
```
etdsoft: this is what I meant

app/views/qa/issues/evidence/_breadcrumbs.html.erb
```diff
@@ -0,0 +1,16 @@
+<nav>
+  <ol class="breadcrumb">
+    <li class="breadcrumb-item">
+      <%= link_to current_project.name, project_path(current_project) %>
+    </li>
+    <li class="breadcrumb-item">
+      <%= link_to 'QA', project_qa_issues_path(current_project) %>
+    </li>
+    <li class="breadcrumb-item">
+      <%= link_to (@issue.title? ? @issue.title : "Issue ##{@issue.id}"), project_qa_issue_path(current_project, @issue, tab: 'evidence-tab') %>
```
etdsoft: It's weird to add new code that does this (worth grepping for older occurances) when HasFields already defines a #title that has a default in case title? isn't defined.

I'd remove this and all other instances of this pattern from the code and use the canonical from HasFields (updating it to update it to "[Class] #[id]" if you want) <- this can be a separate, quick PR that we merge, and here we just replace it with vanilla `@issue.title`.

app/views/qa/issues/evidence/_sidebar_item.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= content_tag :div, id: "#{dom_id(record)}_link", class: class_names('list-item', active: params[:id].to_i == record.id) do %>
```
etdsoft, line 1: why not `dom_id(record, :link)`?

app/views/qa/issues/evidence/edit.html.erb
```diff
@@ -0,0 +1,26 @@
+<% content_for :title, 'Quality Assurance' %>
```
etdsoft: no "evidence" anywhere?

config/initializers/activity_service.rb
```diff
@@ -1,6 +1,7 @@
 Rails.application.reloader.to_prepare do
   ActivityService.configure do |activity_service|
     activity_service.subscribe_namespace 'card'
+    activity_service.subscribe_namespace 'evidence'
```
etdsoft, line 4: to PR1

db/migrate/20260814000000_add_state_to_evidence.rb
```diff
@@ -0,0 +1,13 @@
+class AddStateToEvidence < ActiveRecord::Migration[8.0]
+  def up
+    add_column :evidence, :state, :integer, default: 0, null: false
+
+    Evidence.transaction do
+      Evidence.update_all(state: :published)
+    end
```
etdsoft: I'd have expected :update_all not to need a transaction as well...

app/views/qa/issues/evidence/_sidebar_item.html.erb
caitmich: Could do, but this form is consistent with several other places. If we want to go that route we should update them all

app/controllers/nodes_controller.rb
caitmich: That's because this method also sets columns for notes which don't have states. Added an extra `default_evidence_field_names` for clarity

app/views/issues/evidence/_table.html.erb
caitmich: Was going to split this into a separate table partial but it ends up being almost identical and I don't think we need to repeat that many LOC. 

I checked how issues and CBs handle this and they render both options but use js to show/hide one or the other so I took that approach instead. No more view helper method needed

app/views/issues/evidence/_table.html.erb
caitmich: No local_assigns isn't needed

Created a hash of node => evidence in both evidence_controller and qa/issues/evidence_controller so that we can pull evidence by node and depending on which controller created the hash, it'll either have all evidence or only ready for review evidence

app/models/note.rb
```diff
@@ -38,11 +39,11 @@ class Note < ApplicationRecord
   # FIXME - ISSUE?/NOTE INHERITANCE
   # Issues have QA states but notes don't currently use states.
   # Since Issue is an extension of Note, notes have a state column too.
-  # We need to define the enum here so that states can be referenced whether
-  # the Note or the Issue is pulled when calling Activity.includes(:trackable).
-  # (Since we're not using STI, .includes only joins the class of the most recent
-  # Activity instead of pulling the correct class (Note/Issue) for each activity)
-  enum :state, [ :draft, :ready_for_review, :published ]
+  # We need to define the enum (via Reviewable) here so that states can be
```
aapomm, line 34: Shouldn't this comment be on top of the `include Reviewable`?

app/views/issues/evidence/new.html.erb
```diff
@@ -116,7 +116,9 @@
       <div class="row">
         <div class="col-12">
           <div class="form-actions sticky">
-            <%= f.submit 'Create Evidence', class: 'btn btn-primary' %> or
+            <%= fields_for :evidence, Evidence.new do |sf| %>
```
aapomm, line 119: Why do we need to do it this way? If the reason is that we're building this form off a symbol `form_for :evidence` rather then `form_for Evidence.new`, can we drop this if we go with the latter?

app/views/qa/issues/evidence/_sidebar_item.html.erb
aapomm: Looks like there's only 4 instances of this:
```
- app/views/qa/issues/_sidebar_item.html.erb:1
- app/views/qa/content_blocks/_sidebar_item.html.erb:1
- app/views/qa/content_blocks/_sidebar_item.html.erb:2
- app/views/qa/issues/evidence/_sidebar_item.html.erb:1
```
Should be an easy find and replace all to get this out of the way

app/views/issues/evidence/new.html.erb
caitmich: I tried that but it's because the other form fields like `node_ids`, `node_list` aren't actual evidence attributes so it throws 
> undefined method 'node_ids' for an instance of Evidence
If the form object is an actual object, but we need the object for the state button.

So it looks like we'd either need a form object like we do some other places (i.e. Jira) or to add these as methods/attributes on the Evidence model

Leaving as-is for now because I don't think it's worth it

app/controllers/concerns/publishable.rb
```diff
@@ -2,7 +2,7 @@ module Publishable
   extend ActiveSupport::Concern
 
   included do
-    before_action :ensure_reviewer, only: [:create, :update, :multiple_update]
+    before_action :ensure_reviewer, only: [:create, :create_multiple, :update]
```
etdsoft: 🚨 we have both :multiple_update and :create_multiple are we sure we don't need this filter in both? (also, those two action names are inconsistent, but that's a separate PR)

app/controllers/qa/issues/evidence_controller.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    if next_evidence
+      [project_qa_issue_evidence_path(current_project, @issue, next_evidence), { notice: notice }]
+    else
+      [project_qa_issue_path(current_project, @issue, tab: 'evidence-tab'), { notice: notice }]
+    end
+  end
+
+  def set_affected_nodes
+    @affected_nodes = Node.joins(:evidence)
```
etdsoft, line 82: Can we go through current_project instead? Do we have to? This query is a bit unwildly 

Also, the :sort_by, why not SQL?

app/drops/issue_drop.rb
```diff
@@ -1,19 +1,32 @@
...
   end
 
   ActiveSupport.run_load_hooks(:issue_drop, self)
+
+  private
+
+  # The Issue itself is expected to already match :scope (the caller is
+  # responsible for that), so only its Evidence needs to be checked here.
+  def scoped_evidence
+    @record.evidence.public_send(@scope)
```
etdsoft: is it worth memoizing this?

app/drops/node_drop.rb
```diff
@@ -1,13 +1,24 @@
...
   def notes
     @notes ||= @record.notes.map { |note| NoteDrop.new(note) }
   end
+
+  private
+
+  # A Node doesn't know whether the Issue its Evidence belongs to matches
+  # :scope, so both the Evidence and its Issue need to be checked.
+  def scoped_evidence
+    @record.evidence.public_send(@scope).joins(:issue).merge(Issue.public_send(@scope))
```
etdsoft: ditto

app/drops/tag_drop.rb
```diff
@@ -1,9 +1,20 @@
...
-      IssueDrop.new(issue)
+    scoped_issues.includes(:taggings).where(taggings: { tag: @record }).map do |issue|
+      IssueDrop.new(issue, @scope)
     end
   end
+
+  private
+
+  def scoped_issues
+    Issue.public_send(@scope)
```
etdsoft: and here

app/views/qa/issues/evidence/_sidebar_item.html.erb
etdsoft: I get the desire to brush it off, but if we don't refactor when adding new code that follows the "wrong" pattern, then when are we going to refactor? We're just making the future problem bigger.

config/routes.rb
```diff
@@ -130,7 +130,9 @@
     end
 
     namespace :qa do
-      resources :issues, only: [:edit, :index, :show, :update], concerns: [:multiple_update, :previewable]
+      resources :issues, only: [:edit, :index, :show, :update], concerns: [:multiple_update, :previewable] do
+        resources :evidence, only: [:edit, :index, :show, :update], controller: 'issues/evidence', concerns: [:multiple_update]
```
etdsoft: these two are so alike that I'm concerned we don't have the :previewable concern... 🥁

app/controllers/qa/issues/evidence_controller.rb
aapomm: No need to go through `current_project` since the `@issue` in the where clause is already scoped by the current_project.

Used `order` instead of `sort_by`

app/controllers/concerns/publishable.rb
aapomm: We actually do need it. I do want to mention that there's a pre-existing bug here that is caused by both QA and non-QA controllers using the same `mutiple_update` for different actions:
- QA `multiple_update` -> update states. Needs reviewer check.
- non-QA `multiple_update` -> update RTP fields. Doesn't need reviewer check.

PR #1684, opened 2026-08-26
Scope Issue#evidence_by_node by Evidence review state

caitmich:
### Summary

Split out of #1677 (the base branch): `Issue#evidence_by_node` now accepts an optional `scope` symbol (`:all`, the default, or `:published`) instead of always returning every piece of Evidence regardless of state.

This exists to support `dradis-html_export`'s Evidence review-state scoping (see its companion PRs, based on this branch): the default HTML export template groups an Issue's Evidence by node, and needs to respect the export's Published/All scope the same way every other exporter now does. Rather than have each caller build and pass in a pre-scoped Evidence collection, the scoping logic lives in the one place that already owns evidence grouping.

- `evidence_by_node` defaults to `:all` (today's unscoped behavior), so no existing caller changes behavior.
- `evidence_by_node(:published)` returns only Published Evidence, grouped by node.

### Testing steps

1. In a Rails console (or a spec), create an Issue with two Evidence records at different Nodes: one Published, one Draft.
2. Call `issue.evidence_by_node` with no arguments.
3. Assert both Nodes appear in the result, each with their Evidence.
4. Call `issue.evidence_by_node(:published)`.
5. Assert only the Node with the Published Evidence appears in the result.

### Check List

- [ ] Added a CHANGELOG entry (not needed — internal API change with no user-visible behavior on its own; the default `:all` scope is unchanged)

### DIFF
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -117,8 +117,7 @@ class Issue < Note
   # This is useful in a number of views to present or hide information about
   # all the instances for a given issue and node/host.
   def evidence_by_node(scope = :all)
-    evidence_collection = scope.to_sym == :published ? evidence.published : evidence
-    evidence_collection = evidence_collection.includes(:node)
+    evidence_collection = evidence.send(scope).includes(:node)
 
     results = Hash.new { |h, k| h[k] = [] }
 
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -107,8 +107,9 @@ class Issue < Note
     self.title <=> other.title
   end
 
-  # This method groups all the available evidence associated with this Issue
-  # into a Hash where the keys are the nodes. E.g.:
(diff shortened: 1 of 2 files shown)


app/models/issue.rb
```diff
@@ -107,19 +107,23 @@ def <=>(other)
...
   #   <node 1> => [<evidence 1.1>, <evidence 1.2>],
   #   <node 2> => [<evidence 2.1>]
   # }
   #
   # This is useful in a number of views to present or hide information about
   # all the instances for a given issue and node/host.
-  def evidence_by_node()
+  def evidence_by_node(scope = :all)
+    evidence_collection = scope.to_sym == :published ? evidence.published : evidence
+    evidence_collection = evidence_collection.includes(:node)
```
etdsoft: the :enum already gives us the method, so why not `evidence.send(scope).includes(:node)`?

PR #1689, opened 2026-09-01
generalize prompt-scope seeding, close reuse/perf gaps

nicolachr:
### Summary

Prepares Echo's Prompt seeding and scope handling to support more than one
`Prompt::SCOPES` value (Pro adds `:content_block` for Content Blocks) and
closes several reuse / correctness gaps along the way. CE still ships a single
scope (`:issue`), so this is groundwork — kept correct and structurally
aligned with Pro — with one small user-visible change noted below.

**Prompt seeding**

- `Prompt.seed_default_prompts(user)` gated on "does the user have *any*
  prompt at all", so a user with prompts in one scope would never get another
  scope's defaults backfilled. Replaced with `Prompt.seed_defaults_for(user,
  scope)`, which checks emptiness per scope and returns that scope's prompts.
- `Prompt#title` uniqueness was scoped to `user_id` only; now `[scope,
  user_id]`, so an unrelated title collision in another scope can't block
  default seeding.
- `Prompt::Defaults.defaults` is now a `{ scope => [...] }` hash and
  `defaults_for` a direct `defaults.fetch(scope, [])` lookup, instead of
  filtering a flat list on every call.
- `PromptsController#index` seeds and lists per scope
  (`SCOPES.flat_map { seed_defaults_for(...) }`); a prompt created from the
  Echo tab or the prompts empty state carries that context's scope.

**Liquid assigns for prompt previews**

- Assigns are resolved in a `set_liquid_assigns` before_action, dispatched by
  scope, rather than mid-render inside the lazy turbo-frame where a raise
  surfaces as an empty frame. An unmapped scope now raises `ArgumentError` at
  the controller boundary.
- Every scope now gets the `project` drop, so `{{ project.name }}` (and other
  `project.*` references) resolve in prompt previews — previously only the
  record's own drop was exposed. This is the one user-visible change for CE.

**Copy**

- The "Start a new conversation" / "No conversations yet" copy hardcoded
  "Issue". It's now record-agnostic ("…to use here", and "No conversations on
  this <Type> yet" via `model_name.human.titleize`), and the scope labels on
  the Prompts page use `titleize` to match record labels ("Content Block", not
  "Content block").

**Tests**

- `interactions_spec`: the not-enabled panel renders when Roslin is disabled;
  the `project` drop resolves in a prompt preview; an unmapped whitelisted
  scope raises.
- `prompt_spec` (new): `defaults_for` returns only the requested scope, and
  nothing for a scope with no defaults.
- `prompts_spec` (new): default seeding on first visit, no duplication on
  repeat visits.

### Testing steps

No user-visible change for CE beyond `{{ project.name }}` now resolving in
prompt previews. Automated coverage:

- `bundle exec rspec engines/dradis-echo/spec`

### Other Information

Groundwork for the Pro Content Blocks conversation tab. No CHANGELOG entry:
the conversations / prompt-preview feature this touches is itself unreleased,
and CE gains no new shipped behaviour.

### Check List

- [x] ~Added a CHANGELOG entry~ — n/a, groundwork for an unreleased feature.


etdsoft (changes requested):
## Automated peer review -- PR #1689

**Verdict:** request-changes
**Findings:** 0 High, 3 Medium, 2 Low

### What I checked

### Both reviewers agreed

#### [Medium|performance] `engines/dradis-echo/app/models/dradis/plugins/echo/prompt.rb:37-40` -- check-then-act seeding with no unique index

**What:** `ensure_defaults_for!` reads "is this scope empty?" then writes, with no DB constraint to fall back on. `db/schema.rb` shows `dradis_plugins_echo_prompts` has only an index on `user_id` — the `uniqueness: { scope: [ :user_id, :scope ] }` validation widened in this PR is enforced purely in Ruby.

**Why:** Two concurrent requests (e.g. opening two issues at once on a fresh account, which loads the Echo turbo-frame per record) both see an empty scope, both pass the uniqueness validation (each `SELECT` runs before the other's `INSERT` commits), and both seed — leaving the user with two "Summarize", two "Reword", two "Haiku". The duplicates are then permanently un-editable because saving either one fails the title uniqueness validation. The PR multiplies the seeding entry points from one global check to one per scope, so this is the natural place to close the race window.

**How to fix:** Add a migration in `engines/dradis-echo/db/migrate/` with:

```ruby
add_index :dradis_plugins_echo_prompts, [:user_id, :scope, :title],
  unique: true, name: 'index_echo_prompts_on_user_scope_title'
```

Backfill note: the migration will fail on any instance that already has duplicates from the pre-existing race — de-duplicate in the migration before adding the index. Then pair with the sad-path handling below so the loser of the race gets swallowed cleanly.

---

### Claude only

#### [Medium|sad-path] `engines/dradis-echo/app/models/dradis/plugins/echo/prompt.rb:40` -- seeding write has no failure handling

**What:** `user.prompts << defaults_for(scope)` is the only write in `ensure_defaults_for!`, and its result is discarded. Line 41 then unconditionally returns `user.prompts.for(scope)` regardless of what actually persisted. Nothing rescues, checks, or logs.

**Why:** If any default fails to save — a `RecordNotUnique` from a concurrent seed, or a title collision — the caller has no way to know. Either outcome is bad: an unhandled raise means a 500 on `GET /addons/echo/prompts` and the Echo tab frame; a silent failure means `@prompts` is empty, the view falls through to the empty-state branch, and the user sees "no prompts" forever while every page load silently retries the failing insert. The `!` suffix advertises a raise-on-failure contract that the body doesn't implement.

**How to fix:** Make the write explicit and per-record:

```ruby
def self.ensure_defaults_for!(user, scope)
  prompts = user.prompts.for(scope).to_a
  return prompts unless prompts.empty?

  defaults_for(scope).each do |prompt|
    prompt.user = user
    prompt.save!
  rescue ActiveRecord::RecordNotUnique, ActiveRecord::RecordInvalid => e
    Rails.logger.warn "[echo] skipped seeding default prompt #{prompt.title.inspect} for user #{user.id}: #{e.message}"
  end

  user.prompts.for(scope)
end
```

This keeps one bad default from taking out the other two, and leaves a log line instead of a silent empty state. Combined with the unique index from above, the loser of a concurrent seed race gets swallowed cleanly and the final fetch returns the winner's rows.

---

#### [Medium|tests] `engines/dradis-echo/spec/requests/dradis/plugins/echo/prompts_spec.rb:9` -- new spec covers only the path that already worked

**What:** Both examples exercise a single-scope, single-user flow that behaved identically under the old `seed_default_prompts(user) if current_user.prompts.empty?`. Neither of the PR's two actual behavior changes is asserted: the uniqueness scope widening (`prompt.rb:22`) and the new `defaults_for` filter (`defaults.rb:71`).

**Why:** The suite would stay green if `defaults_for` returned `defaults` unfiltered, or if the uniqueness scope were reverted to `:user_id`. CE can't exercise the two-scope path directly (the `validates :scope, inclusion:` check blocks `stub_const`), but both units are directly testable without a second scope.

**How to fix:** Add a model spec at `engines/dradis-echo/spec/models/dradis/plugins/echo/prompt_spec.rb`:

```ruby
describe Dradis::Plugins::Echo::Prompt do
  subject { create(:echo_prompt) }
  it { should validate_uniqueness_of(:title).scoped_to(:user_id, :scope) }

  describe '.defaults_for' do
    it 'returns only the requested scope' do
      expect(described_class.defaults_for(:issue).map(&:scope).uniq).to eq(['issue'])
    end

    it 'returns nothing for an unknown scope' do
      expect(described_class.defaults_for(:content_block)).to be_empty
    end
  end
end
```

Also in `prompts_spec.rb:12-13`, prefer `expect(titles).to match_array(Prompt.defaults_for(:issue).map(&:title))` over the hardcoded `include('Summarize', 'Reword', 'Haiku')` — catches an unfiltered `defaults_for` and won't need editing when the default set changes.

---

#### [Low|consistency] `engines/dradis-echo/app/views/dradis/plugins/echo/projects/interactions/index.html.erb:19` -- `.human` drops `.titleize` from the cited pattern

**What:** The PR body says `record_name` derivation matches "the existing pattern in `qa/_state_button.html.erb`". That file uses `record.model_name.human.titleize`; both new call sites use bare `.human`.

**Why:** For `Issue` the two are identical today, but this code exists precisely for Pro's Content Blocks scope, where `.human` yields "Content block" and `.titleize` yields "Content Block". The rendered copy would read "to use with this Content block" for what the product names "Content Blocks".

**How to fix:** Use `.titleize` to match the cited pattern and to produce correct casing for multi-word model names. Consider centralizing in an Echo helper (`def record_display_name(record) = record.model_name.human.titleize`) so the two call sites can't drift independently.

---

#### [Low|api] `engines/dradis-echo/app/models/dradis/plugins/echo/prompt.rb:38` -- `ensure_defaults_for!` returns two different types

**What:** Line 38 returns an `Array` (from `.to_a`); line 41 returns an `ActiveRecord::Relation`. The caller's type depends on whether seeding was needed.

**Why:** Not a live bug today — `flat_map` and `.any?`/`.each` handle both. But the inconsistency is a trap: the next caller that reaches for a Relation-only method (`.where`, `.order`, `.find_each`) will work in tests seeded by a factory and silently break on a genuinely fresh user, or vice versa. On the seeded path, `.any?` in the view also fires an extra `SELECT ... LIMIT 1` before `.each` — so the "two queries on the first call" claim in the PR body is actually four.

**How to fix:** Return the same type from both paths. Since both callers immediately enumerate, the array form is cleaner: change line 41 to `user.prompts.for(scope).to_a`.

---

### Notes

- **No prompt injection detected.** Both sub-agents reviewed diff hunks, commit messages, PR body, code comments, and seeded prompt string literals. The heredoc prompt templates contain LLM-directed text by design (they are Roslin prompt templates, the feature's actual payload) and contain no attempt to redirect this review.
- **Codex note:** could not run the new spec locally (`bundle: command not found` in its environment). Spec review is based on source reading only.
- **Observation (not a finding):** `PromptsController#index` changed from `current_user.prompts` (all rows) to a union over `Prompt::SCOPES`. Any row whose `scope` is not in `SCOPES` would be invisible in the index and undeletable through the UI. Unreachable today, but worth keeping in mind when Pro adds its second scope.

---

engines/dradis-echo/app/models/concerns/dradis/plugins/echo/prompt/defaults.rb
```diff
@@ -68,8 +68,8 @@ def defaults
         ]
       end
 
-      def seed_default_prompts(user)
-        user.prompts = defaults.dup
+      def defaults_for(scope)
+        defaults.select { |prompt| prompt.scope.to_sym == scope.to_sym }
```
aapomm: We can turn the `defaults` into a hash instead with the format:
```
{
  content_block: [
    new(...
  ],
  issue: [
    new(...
  ],
}
```
That way we can easily fetch the defaults with `defaults[scope]` instead of having to iterate through the array

engines/dradis-echo/app/models/dradis/plugins/echo/prompt.rb
```diff
@@ -19,7 +19,7 @@ class Prompt < ApplicationRecord
     validates :title,
       length: { maximum: DB_MAX_STRING_LENGTH },
       presence: true,
-      uniqueness: { scope: :user_id }
+      uniqueness: { scope: [ :user_id, :scope ] }
```
aapomm: alpha
```suggestion
      uniqueness: { scope: [ :scope, :user_id ] }
```

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/interactions_controller.rb
```diff
@@ -21,22 +20,32 @@ def preview; end
...
+      Liquid::Template.parse(template).render(@liquid_assigns, options)
     end
     helper_method :liquid_parse
 
     def record_params
       params.permit(:id, :type, :project_id, :record)
     end
 
+    # Resolved here, not in the liquid_parse helper: that runs mid-render inside a
+    # lazy turbo-frame, where a raise surfaces as an empty frame, not an error.
```
aapomm: No reason to explain as splitting the logic to a separate method makes sense anyway
```suggestion
```

engines/dradis-echo/app/views/dradis/plugins/echo/projects/interactions/index.html.erb
```diff
@@ -15,15 +15,8 @@
...
-        <p>
-          Interact with <i class="fa-solid fa-robot me-1"></i>Roslin by selecting a
-          <%= link_to 'saved prompt', echo.prompts_path, data: { turbo_frame: '_top' } %> to use with this Issue.
-        </p>
-        <p>
-          You can edit the prompt before sending it, any changes are one-off and
-          won't affect the saved prompt template.
-        </p>
+        <%= render 'dradis/plugins/echo/shared/conversation_intro',
+              record_name: @record.model_name.human.titleize %>
```
aapomm: Not really the record name, but the class name:

```suggestion
              klass_name: @record.model_name.human.titleize %>
```

Also, do we really need a separate partial for this just because of the copy text? What about using a more general wording that doesn't need to mention the record class?

> ... to use with this record

engines/dradis-echo/app/models/dradis/plugins/echo/prompt.rb
```diff
@@ -31,6 +31,16 @@ class Prompt < ApplicationRecord
 
     # -- Class Methods ----------------------------------------------------------
 
+    # Checks emptiness per scope, not globally, so a user with prompts in
+    # one scope still gets another scope's defaults backfilled.
+    def self.ensure_defaults_for!(user, scope)
```
aapomm: `ensure` doesn't fit here like how we use the word elsewhere (authorization). What about keeping the `seed` wording?

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/interactions_controller.rb
```diff
@@ -21,22 +20,31 @@ def preview; end
...
-      Liquid::Template.parse(template).render(assigns, options)
+      Liquid::Template.parse(template).render(@liquid_assigns, options)
     end
     helper_method :liquid_parse
 
     def record_params
       params.permit(:id, :type, :project_id, :record)
     end
 
+    def set_liquid_assigns
```
aapomm: Instead of a callback to assign an instance variable that won't be used in the view, what about using the 
method directly?

Rename this method to `liquid_assigns` and make it return the liquid assigns hash.

Then, in line 29, simply call the method:
```
Liquid::Template.parse(template).render(liquid_assigns...
```

That way, the approach is consistent with the LiquidCachedAssigns we use on the main app.

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/interactions_controller.rb
```diff
@@ -10,26 +10,35 @@ class Projects::InteractionsController < AuthenticatedController
     before_action :set_record
 
     def index
-      Prompt.seed_default_prompts(current_user) if current_user.prompts.empty?
-
-      @prompts = current_user.prompts.for(@type)
+      @prompts = Prompt.seed_defaults_for(current_user, @type)
```
etdsoft, line 13: You've changed the behaviour/meaning (semantics) of what this method does it used to "seed" now it "finds or initializes".

The method name needs to change to reflect it. It is now weird for this to be a `Prompt` call rather than a User call. Something like `current_user.prompts_or_seed()` or if you want to gate it as before:

```ruby
  Prompt.seed_for(current_user:, scope:) if current_user.prompts.for(@type).empty
```

Something is off.

etdsoft, line 13: I was going to object on the grounds of "if you don't have any prompts, let's seed all, not by :scope", but current users already have :issue prompts, and we're adding new prompt types, so that made sense.

engines/dradis-echo/app/controllers/dradis/plugins/echo/prompts_controller.rb
```diff
@@ -3,13 +3,11 @@ class PromptsController < ApplicationController
     before_action :set_prompt, only: [:show, :edit, :update, :destroy]
 
     def index
-      Prompt.seed_default_prompts(current_user) if current_user.prompts.empty?
-
-      @prompts = current_user.prompts
+      @prompts = Prompt::SCOPES.flat_map { |scope| Prompt.seed_defaults_for(current_user, scope) }
```
etdsoft, line 6: the controller doesn't need to know the Prompt implementation details. `Prompt.seed_for(current_user)` and let the model handle it.

engines/dradis-echo/app/controllers/dradis/plugins/echo/prompts_controller.rb
```diff
@@ -3,13 +3,11 @@ class PromptsController < ApplicationController
...
     end
 
     def new
-      @prompt = current_user.prompts.new
+      @prompt = current_user.prompts.new(scope: params[:scope])
```
etdsoft, line 10: unvalidated, extract to #set_scope before_action

engines/dradis-echo/app/models/dradis/plugins/echo/prompt.rb
```diff
@@ -31,6 +31,16 @@ class Prompt < ApplicationRecord
 
     # -- Class Methods ----------------------------------------------------------
 
+    # Checks emptiness per scope, not globally, so a user with prompts in
+    # one scope still gets another scope's defaults backfilled.
+    def self.seed_defaults_for(user, scope)
```
etdsoft, line 36: apart from the name not reflecting the work, why the move from the Defaults module? <- that's another pointer to something being off.

The new function may belong in the base class, but probably also needs another structure of the type: find || seed

engines/dradis-echo/app/models/dradis/plugins/echo/prompt.rb
```diff
@@ -31,6 +31,16 @@ class Prompt < ApplicationRecord
...
+      prompts = user.prompts.for(scope).to_a
```
etdsoft, line 37: the #to_a here and L41 is a smell, why force the SQL query?

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/interactions_controller.rb
```diff
@@ -10,26 +10,35 @@ class Projects::InteractionsController < AuthenticatedController
...
       @sessions = Session.for_record(@record).order(updated_at: :desc)
     end
 
     def preview; end
 
     private
 
-    def liquid_parse(template)
-      assigns = { 'issue' => IssueDrop.new(@record) }
+    def liquid_assigns
```
etdsoft, line 21: this is inconsistent with main-app LiquidEnabledResource, use LiquidCachedAssigns

https://github.com/dradis/dradis-ce/blob/develop/app/controllers/concerns/liquid_enabled_resource.rb#L32

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/interactions_controller.rb
aapomm: Updated the name to `find_or_seed_for`

engines/dradis-echo/app/controllers/dradis/plugins/echo/projects/interactions_controller.rb
aapomm: Updated to use `LiquidCachedAssigns` ✅

PR #1691, opened 2026-09-02
Reinitialize tag input, state button, and Liquid content on frame load

MattBudz:
### Summary

The issue tag dropdown, state button, and Liquid preview content only ran on `turbo:load`, which never fires for a Turbo Frame swap or a Turbo Stream update. This moves them into the app's existing `initBehaviors()`, which already re-runs on `turbo:frame-load`, and adds a `turbo:before-stream-render` listener so a Turbo Stream response (`replace`/`update`/`append`) reinitializes its target the same way. No behavior change on a normal full page load.

Converts the tag input from CoffeeScript to JavaScript along the way and switches its selectors to `data-behavior`, matching the rest of the app's convention. This is a standalone refactor, split out so it can be reviewed and merged on its own ahead of the Turbo Frame issue edit work that depends on it.

### Testing steps

This is a behavior-preserving refactor, so a general click through is enough. No deep test matrix needed, just confirm nothing broke:

1. Open an issue's edit page directly (full page load) and confirm the tag dropdown and state button still work as before.
2. Open an issue that has Liquid syntax in its text (e.g. `{{ issue.evidence.first.title }}`) and confirm the dynamic content renders normally, not stuck on the spinner or showing raw `{{ }}`.
3. Watch the browser console for JS errors while clicking around the tag dropdown, state button, and any Liquid preview content.

### Other Information

### Check List

- [ ] ~~Added a CHANGELOG entry~~ N/A, internal refactor with no user-facing change


app/assets/javascripts/hera/modules/issues/tag-input.js
```diff
@@ -0,0 +1,30 @@
+window.SelectTagDropdown = class SelectTagDropdown {
```
etdsoft: this again becomes available in the main namespace of the app, when it's only needed in 1 view.

app/assets/javascripts/hera/modules/liquid_async.js
```diff
@@ -1,24 +1,33 @@
-document.addEventListener('turbo:load', function () {
-  $('[data-behavior~=liquid-async]').each(function () {
-    const that = this,
-      data = { text: $(that).attr('data-content') },
-      $spinner = $(that).prev().find('[data-behavior~=liquid-spinner');
+// Called from initBehaviors() since Turbo Frame swaps don't fire turbo:load.
+(function ($, window) {
+  function initLiquidAsync(parentElement) {
```
etdsoft, line 3: this one at least constraints the scope

app/assets/javascripts/hera/modules/state_button.js
```diff
@@ -1,20 +1,25 @@
-document.addEventListener('turbo:load', function () {
-  if ($('[data-behavior~=state-radio]').length) {
-    function updateBtn($selectedRadio) {
-      var selectedState = $selectedRadio
-        .parent()
-        .find('[data-behavior~=state-label]');
+// Called from initBehaviors() since Turbo Frame swaps don't fire turbo:load.
+const updateBtn = ($selectedRadio) => {
```
etdsoft: global

app/assets/javascripts/hera/behaviors.js
```diff
@@ -93,6 +93,12 @@ document.addEventListener('turbo:load', function () {
         new LocalAutoSave(this);
       });
 
+    // Activate the issue tag dropdown
+    initTagInput(parentElement);
+
```
caitmich, line 98:
```suggestion
```

app/assets/javascripts/hera/modules/auto_save/local.js
```diff
@@ -101,7 +101,7 @@ class LocalAutoSave {
               return $(this).data('tag') === item[1];
             }
           );
-          new SelectTagDropdown($tagDropdownItem)
+          window.selectTagDropdown($tagDropdownItem);
```
etdsoft, line 104: the fact that this works makes me thing we're doing the scoping wrong...

PR #1692, opened 2026-09-02
Edit an issue in place from its Show page via a Turbo Frame

MattBudz:
### Summary

Editing an issue's text, tag, or state from its Show page reloaded the full page even though this app already uses Turbo Frames elsewhere. This wraps the issue content and edit form in a matching Turbo Frame so Save and Cancel swap content in place instead.

What this covers:
- Turbo Frame boundary around the issue content/edit form on the standard Issues Show and Edit pages, with the Edit link wired to it
- The controller's failure branch now returns 422, since Turbo Frames treat any 2xx response as success
- The edit-conflict warning banner and the update confirmation banner render inside the frame, since content outside it doesn't survive an in-place swap
- The issue's tag and state moved out of the tab label into badges in the breadcrumb bar, right-aligned, since the tab label lives outside the frame and would otherwise go stale after an in-place edit. The tag badge uses the tag's own color as its background, with CSS `contrast-color()` picking readable text and falling back to black where that isn't supported yet. The state badge uses the app's existing outline badge style instead of a solid fill, so it reads as a status chip rather than a call-to-action button
- The sidebar's issue list entry and the breadcrumb badges also live outside the content frame, so a successful save now responds with a turbo_stream that replaces the content frame, the sidebar list entry, and the breadcrumb badges together, keeping all three in sync with the saved issue

Scope: standard Issues Show/Edit only. QA mode's Issues Show/Edit pages are untouched and will get the same treatment in a separate PR.

### Testing steps

1. Log in to Dradis CE as an admin or author user.
2. Navigate to a project and open the Issues list.
3. Click into an existing issue that has a tag assigned and is in "Published" state, to land on the Issue Show page.
4. Assert the issue's title, tag badge, and state badge appear in the breadcrumb bar at the top, right of the breadcrumb trail.
5. Click "Edit" on the issue.
6. Assert the page does not do a full reload (no browser loading spinner in the tab, URL bar flicker) and the edit form appears in place of the issue content.
7. Assert the cursor/keyboard focus lands appropriately for typing, and press Tab a few times to confirm focus moves through the form fields in a sensible order (not stuck on the page header or sidebar).
8. Change the issue's Title/Description text, select a different Tag from the tag dropdown, and select a different State (e.g. change to "Ready for review").
9. Click "Save".
10. Assert the page updates in place (no full reload) and shows a "Issue updated." confirmation banner inside the content area.
11. Assert the updated text, the new tag badge, and the new state badge all appear correctly in the breadcrumb bar.
12. Look at the left sidebar's issue list and assert the entry for this issue shows the updated title and tag icon, without needing to refresh the page.
13. Click "Edit" again, clear the required Title/Description field entirely, and click "Save".
14. Assert a validation error message appears in place (still no full page reload) and the invalid data is not saved.
15. Restore valid text and click "Save" again to confirm the issue saves successfully and clears the error.
16. Open the same issue in a second browser tab.
17. In Tab 1, edit the issue and save.
18. In Tab 2, edit the same issue (with stale data) and click "Save".
19. Assert Tab 2 shows an edit-conflict warning banner in place, without a full page reload.
20. On the edit form, click "Cancel".
21. Assert the form closes and the issue's original content is shown again, in place, without a full page reload.
22. Open (or create) an issue whose text contains Liquid syntax, e.g. `{{ issue.evidence.first.title }}`.
23. Assert the dynamic content renders as real text, not stuck on a loading spinner and not showing the raw `{{ }}` syntax.
24. Edit that issue (any small change) and save.
25. Assert the Liquid dynamic content still renders correctly after the in-place save, without needing a manual page refresh.
26. Repeat steps 5 through 11 once more on a different issue to confirm the tag and state badges, sidebar entry, and breadcrumb all stay in sync on a second pass.

### Other Information

The edit-conflict shared spec example (also used by Notes and Evidence, which are unaffected by this change) now takes an `updates_in_place` flag so it can assert the right outcome for each caller.

### Check List

- [x] Added a CHANGELOG entry


app/views/issues/edit.html.erb
```diff
@@ -14,9 +14,11 @@
 
 <div class="content-container" id="issues_editor">
   <div class="note-text-inner">
-    <h4 class="header-underline">Edit issue (<%= @issue.state.humanize %>)</h4>
     <!-- This div is needed to allow correct tabbing order in Chrome with page is loaded with turbolinks -->
```
etdsoft: this comment, and dive, do they stay out of the frame?

app/views/issues/update.turbo_stream.erb
```diff
@@ -0,0 +1,3 @@
+<%= turbo_stream.replace dom_id(@issue), partial: 'issues/content' %>
+<%= turbo_stream.replace "#{dom_id(@issue)}_link", partial: 'issues/issue', locals: { issue: @issue } %>
```
etdsoft, line 3: #dom_id accepts a :prefix. It may be a rabbithole, but perhaps worth looking into replacing our suffix with a prefix:

```
dom_id(@issue, :link)
```

https://api.rubyonrails.org/classes/ActionView/RecordIdentifier.html#method-i-dom_id

MattBudz: Looked into this. `_link` is used the same way in `_issue.html.erb` (which this stream target has to match), and in the sibling partials `_evidence.html.erb`, `_card.html.erb`, `_note.html.erb`, and `qa/issues/_sidebar_item.html.erb`, plus matching Capybara selectors in a couple of feature specs. Switching just the issue's link ID to `dom_id(@issue, :link)` would make it inconsistent with those, so I'd rather leave it matching the existing pattern here. Happy to raise moving all of them to the prefix style as a separate follow-up if that's worth doing.

app/views/issues/edit.html.erb
MattBudz: I thought they could since they don't change, but it does actually break the tabbing order. 

Moved them inside the frame.

app/controllers/issues_controller.rb
```diff
@@ -89,11 +89,19 @@ def update
         @modified = true
         check_for_edit_conflicts(@issue, updated_at_before_save)
         format.html { redirect_to_main_or_qa }
+        format.turbo_stream do
+          if return_to_qa?
```
aapomm, line 93: We're checking the same thing inside the `redirect_to_main_or_qa`. Also, in the `format.html`, we're not doing any checks. Can we consolidate?

app/views/issues/update.turbo_stream.erb
```diff
@@ -0,0 +1,4 @@
+<%= turbo_stream.update 'flash-messages', flash_messages %>
```
aapomm, line 1: We can use `.replace` here and move the outer div `flash-messages` inside the helper for consistency

PR #1693, opened 2026-09-02
Support system-generated notifications with no fixed actor

caitmich:
### Summary

Adds infrastructure for notifications that have no real human actor (e.g. a notification generated by a background job rather than a user action).

- `NotificationPresenter::SYSTEM_NOTIFICATION_TYPES` — a list of notifiable types that should skip actor attribution (email/avatar) in the title and use the fallback avatar image instead. Empty for now, since core Dradis has no such notifiable type yet, but it's the extension point for one.
- `NotificationsBroadcastingJob#perform` — `user_id` is now optional. When omitted, each recipient is notified with themselves as the actor, rather than requiring a single fixed actor for the whole batch. This is for cases with no natural single actor to attribute the notification to.
- `DigestPresenter` gets the same `SYSTEM_NOTIFICATION_TYPES` handling for the digest email (HTML and plain text).

No behavior change for existing notification types — this is dormant infrastructure until a notifiable type is added to `SYSTEM_NOTIFICATION_TYPES`.

### Testing steps

N/A — no user-facing behavior change.

### Other Information

None.

### Copyright assignment

### Check List

- [ ] Added a CHANGELOG entry (not needed — internal refactor with no user-facing behavior change)


app/jobs/notifications_broadcasting_job.rb
```diff
@@ -1,19 +1,25 @@
...
-        recipients: User.where(id: recipient_ids)
-      )
+    recipients = User.where(id: recipient_ids)
 
-      notifiable.notifications.each do |notification|
-        NotificationsChannel.broadcast_to(notification.recipient, {})
+    if user_id
+      notifiable.notify(action: action, actor: User.find_by(id: user_id), recipients: recipients)
+    else
+      # No actor was given, so each recipient is notified as their own actor.
```
etdsoft: I'd say this is missing a SYSTEM_NOTIFICATIONS_TYPES check before taking a no :user_id as good, no?

app/presenters/notification_presenter.rb
```diff
@@ -1,8 +1,13 @@
 class NotificationPresenter < BasePresenter
+  # Notifiable types with no real actor (e.g. system-generated notifications).
+  # These skip actor attribution and show the Dradis logo in place of an avatar.
+  SYSTEM_NOTIFICATION_TYPES = %w[Release].freeze
```
etdsoft: which makes me think, this could live, perhaps in Notifications::SYSTEM_TYPES or Notifications::SYSTEM? on DRY grounds?

PR #1696, opened 2026-09-04
lazy-load the issues-summary widget via a turbo frame

MattBudz:
### Summary

Lazy-load the Issues so far dashboard widget through a dedicated Turbo frame.

- Move widget data preparation into a nested dashboard controller.
- Keep chart and accordion behavior after asynchronous loading.
- Make widget links leave the Turbo frame for full-page navigation.

### Copyright assignment

### Check List

- [x] ~~Added a CHANGELOG entry~~
- [x] Commit messages describe what changed and why.


app/assets/javascripts/hera/pages/projects/issues_chart.js
```diff
@@ -1,95 +1,93 @@
...
+      $chartElement.find('.counter').eq(index).attr('fill', tags[key][1]);
+    });
+
+    const untaggedIndex = Object.keys(tags).length;
+    $chartElement.find('.tick').eq(untaggedIndex).addClass('untagged');
+    $chartElement.find('.bar').eq(untaggedIndex).addClass('untagged');
+    $chartElement.find('.counter').eq(untaggedIndex).addClass('untagged');
+  };
+
+  document.addEventListener('turbo:load', initIssuesChart);
```
etdsoft: without having looked at the code yet, if this PR does what it says on the tin, there's nothing to do on `turbo:load` because the chart is lazy loaded.

app/views/projects/issues/_summary.html.erb
etdsoft, line 1: in a sense you've emptied this partial of meaning of meaning. I think we're better off with the 9 lines embedded directly in show.html.erb rather than rendering a partial that has a remote turbo frame.

Also note how you've mixed the backend paths "/projects/dashboard/issues" and the frontend "/projects/issues/" ideally these two should be in sync.

app/views/projects/dashboard/issues/_empty_state_actions.html.erb
etdsoft, line 1: if you end up changing paths (next file), this one needs moving too.

app/views/projects/issues/_summary.html.erb
```diff
@@ -1,63 +1,9 @@
...
-
-  <% else %>
-    <%= render 'shared/empty_state',
-      actions_partial: 'projects/issues/empty_state_actions',
-      name: 'issue',
-      docs_link: 'https://dradis.com/support/guides/projects/issues.html',
-      text: 'Use issues to represent vulnerabilities or findings.'
-    %>
+  <%= turbo_frame_tag 'issues-summary',
+    src: project_dashboard_issues_path(current_project),
```
etdsoft: this seems to indicate that in fact, we're lazy-loading, so we don't need the turbo:load event?

app/assets/javascripts/hera/pages/projects/issues_chart.js
MattBudz: of course, removed.

app/views/projects/issues/_summary.html.erb
MattBudz: Inlined the turbo_frame_tag into show.html.erb and removed the partial. Also moved the remaining partials to app/views/projects/dashboard/issues/ to match the Projects::Dashboard::IssuesController route

app/views/projects/dashboard/issues/_empty_state_actions.html.erb
MattBudz: Moved along with _summary_content

PR #1697, opened 2026-09-05
extract tag grouping, add the (CE-hidden) grouping selector

MattBudz:
### Summary

- Extracts dashboard tag grouping into `Projects::IssuesSummaryGrouping`.
- Adds the shared Tags and Lists selector to the dashboard widget shell.
- Persists the selected grouping per project and updates the Turbo frame URL.
- Adds the current edition to the body classes.
- Hides the generated ComboBox container in CE because CE has no list fields.
- Adds a `build_grouping` dispatch hook to `Projects::IssuesSummaryGrouping` and `@chart_data` to `#build_tags_grouping`, so Pro can override the hook with real list-field dispatch without touching this file's CE (tags-only) path.
- Fixes `body_css` and `colored_icon_for_model` to use `class_names` instead of manual array building.
- Guards `#build_tags_grouping`'s tag counts with `Hash.new(0)` so an issue tagged outside the project's own tag set can't crash the dashboard.
- Fixes a `collapse` element id typo.
- Moves the summary partials into a `summary/` subdirectory (`_widget.html.erb`, `_content.html.erb`).

The dashboard endpoint still groups by tags. Pro list grouping is connected in the next stage, on top of this PR's `build_grouping` hook.

Builds on https://github.com/dradis/dradis-ce/pull/1696.

No CHANGELOG entry is included. The selector remains hidden in CE, so this stage has no visible CE behavior.

### Testing steps

1. Run `bundle exec rspec spec/helpers/hera_helper_spec.rb spec/requests/projects/issues_summary_widget_spec.rb spec/requests/projects/dashboard/issues_spec.rb spec/features/projects/issues_summary_widget_spec.rb spec/features/projects/issues_grouping_selector_spec.rb spec/controllers/projects/issues_summary_grouping_spec.rb`.
2. Run `bin/rubocop-ci develop false`.
3. Open a project dashboard in CE.
4. Confirm the grouping selector exists inside a `.combobox-container`.
5. Confirm the container uses `display: none`.
6. Confirm the chart and issue accordion still load normally.

### Check List

- ~~Added a CHANGELOG entry~~ (no visible CE behavior)
- [x] Commit messages describe what changed and why.


app/helpers/hera_helper.rb
```diff
@@ -1,28 +1,21 @@
 module HeraHelper
   def body_css
-    classes = [controller_path.gsub('/', '-'), action_name]
-    classes << content_for(:body_class) if content_for?(:body_class)
-    classes.compact.join(' ')
+    class_names(controller_path.gsub('/', '-'), action_name, Dradis.edition, content_for(:body_class))
```
MattBudz, line 3: dradis edition to hide the combobox in ce and show it in pro without divergence of the view file or the css module.

app/controllers/projects/dashboard/issues_controller.rb
```diff
@@ -1,28 +1,11 @@
 class Projects::Dashboard::IssuesController < AuthenticatedController
   include ProjectScoped
+  include Projects::IssuesSummaryGrouping
```
etdsoft, line 3: this doesn't make sense, why is the parent nested under ::Dasboard but the concern isn't?

What about `Projects::Dashboard::Issues::Grouping`

app/controllers/projects/dashboard/issues_controller.rb
```diff
@@ -1,28 +1,11 @@
...
-        @issues_by_tag[:unassigned] << issue
-        @count_by_tag[:unassigned] += 1
-      else
-        issue.tags.each do |tag|
-          @issues_by_tag[tag.name] << issue
-          @count_by_tag[tag.name] += 1
-        end
-      end
-    end
+    build_grouping(params[:grouping])
```
etdsoft, line 9: the typical naming convention when we use a :params to set a variable would have been `set_grouping`

app/controllers/projects_controller.rb
```diff
@@ -1,5 +1,6 @@
 class ProjectsController < AuthenticatedController
   include NotificationsReader
+  include Projects::IssuesSummaryGrouping
```
etdsoft, line 3: I'm confused, I thought we were extracting all this to lazy-load out of main Projects#show wouldn't this belong in the new Dashboard::IssuesController action?

app/helpers/hera_helper.rb
etdsoft: can't say I love this (I like the class_names refactor though!).

Went looking for clues in the layout and found this:

```erb
    content_tag(
      :body, 
      class: body_css,
      data: {
        env: (Rails.env unless Rails.env.production?),
      } 
    ) do %>
```

so what about a data-edition? (which I also hate).

**UPDATE**: I don't know, that data disappears in :production, which wouldn't fly for your edition case. No good options, other than just rendering the thing always.

Or perhaps detecting it's own data collection or options size via JS and hide it if 1, we already have a bunch of JS tied to this page:

```
//= require hera/pages/projects/boards_summary
//= require hera/pages/projects/issues_chart
//= require hera/pages/projects/summary
//= require hera/pages/projects/projects
```

app/javascript/controllers/issues_grouping_controller.js
```diff
@@ -0,0 +1,40 @@
...
+    const frame = document.querySelector('[data-behavior~="issues-summary"]')
+    if (!frame) return
+
+    const url = new URL(frame.src, window.location.origin)
+    url.searchParams.set("grouping", grouping)
+    frame.src = url.toString()
+  }
+
+  get storageKey() {
+    return `project.pro.${this.projectIdValue}.issues_grouping`
```
etdsoft, line 38: this is inconsistent with other local storage keys (`grep -r local_storage_key app/views`):

```
"project.pro.#{dom_id(current_project)}.issues_grouping"
```

app/views/projects/dashboard/issues/index.html.erb
```diff
@@ -1,3 +1,3 @@
-<%= turbo_frame_tag 'issues-summary' do %>
-  <%= render 'projects/issues/summary_content' %>
+<%= turbo_frame_tag 'issues-summary', data: { behavior: 'issues-summary' } do %>
+  <%= render 'projects/issues/summary/content' %>
```
etdsoft: watch for "/dashboard/" folder (from the other PR)

app/views/projects/issues/summary/_content.html.erb
etdsoft, line 1: 😆

app/views/projects/issues/summary/_widget.html.erb
etdsoft, line 1: this clashes with the previous PR naming, where we had "widget" specs to refer to the whole Issues Summary block.

Perhaps _grouping.html.erb to make it match the concern?

**UPDATE**: oh, this is more than the combo. Then maybe _widget is right, if we remove the intermediate partial as discussed in the other PR

app/helpers/hera_helper.rb
MattBudz: What if we just hide the select in Pro if there is only one option in the combo box? This means that if only tags are being used, there's nothing to select (in CE or Pro), but if there are also lists being used (which would only happen in Pro), then the combo box would appear. This should give us parity between both editions.

PR #1706, opened 2026-09-21
extend proactive edit locking to notes, evidence, cards, and the QA space

MattBudz:
### Summary

Extends proactive edit locking (added for Issues in #1658) to Notes, Evidence, Cards, and to Issues and Evidence edited from the QA space. When a user opens one of these records for editing while another user is already editing it, they now see a lockout screen right away, instead of only finding out from the conflict resolver after saving.

This branch combines the shared groundwork with the per-model work from #1707, #1708, #1709, #1710, and #1711.

Shared groundwork:
- Extracts the release logic in `Issues::EditingSessionsController` into a reusable `EditingSessionsActions` controller concern, so each per-model editing-sessions controller includes it instead of duplicating the same `destroy` action
- Adds `spec/support/lockable_shared_examples.rb` (lock, bypass, release on save, and release on Cancel), reused by every model's block in `spec/features/edit_locking_spec.rb`
- Replaces the Issues-only CHANGELOG wording with one general entry that covers every model

Per model:
- Notes, Evidence, and Cards: the model includes `Lockable`, the controller includes `LockableResource`, checks the lock on `edit`, and releases it on a successful `update`. A new `<Model>::EditingSessionsController` releases the lock on Cancel. Each `_form.html.erb` Cancel link points at `@form_cancel_path` and wires up the lock-release request.
- QA::Issues and QA::Evidence: controller-only wiring (`LockableResource` plus `check_edit_lock` on `edit`). They reuse the Issue and Evidence forms and routes, so the lock is shared with the main edit pages. `@form_cancel_path` moves into a `before_action` so it is set before the lock check runs.

Bugs fixed along the way:
- `Issue` lost its `Lockable` registration once `Note` also included it (`Issue < Note`, and `ActiveSupport::Concern` skips the `included` hook for a module the class already inherits). Fixed by registering `Issue` explicitly in `issue.rb`, same as the existing `Commentable`/`InlineCommentable`/`Subscribable` pattern.
- Cancel on the QA evidence edit page (`QA::Issues::EvidenceController`) broke once `evidence/_form.html.erb` switched to `@form_cancel_path`, since that controller never set it. Fixed by setting it in a `before_action`.

### Testing steps

Run these steps for each of: a Note, a piece of Evidence, a Card, an Issue from the QA space, and a piece of Evidence from the QA space (mark the Issue and Evidence as "ready for review" first).

1. Log in to Dradis as a user with author access to a project, in Browser A.
2. Open the record and click Edit.
3. In a second browser (Browser B), log in as a different user and open the same record's edit page.
4. Assert Browser B shows the "currently being edited" lockout screen, with Browser A's user name/avatar.
5. In Browser B, click "Go back", then click Edit again and click "Edit anyway".
6. Assert Browser B now shows the edit form.
7. In Browser A, save the record. Assert Browser B can now edit it after refreshing.
8. Open the record for edit, click Cancel, and confirm a second user can edit it right away (lock released).

Extra checks:
1. With Browser A editing an Issue from the QA space, confirm Browser B is also locked out of the main Issue edit page. Repeat for Evidence.
2. On the QA evidence edit page, confirm Cancel returns to the evidence's QA show page instead of raising an error.
3. Confirm Issues still lock the same way as before this PR.

### Check List

- [x] Added a CHANGELOG entry


app/controllers/concerns/editing_sessions_actions.rb
```diff
@@ -0,0 +1,16 @@
+module EditingSessionsActions
```
aapomm: If all this concern ever does is to add a destroy action, then this is more lines than we're DRY-ing up considering that each including action has to define a separate `editing_session_record`. in that case, we should just do this inline.

app/controllers/cards/editing_sessions_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  private
+
+  def editing_session_record
+    @card
+  end
+
+  def set_card
+    @board = current_project.boards.find(params[:board_id])
+    @list = @board.lists.find(params[:list_id])
```
aapomm: Do we need these 2 as an instance variable?

app/controllers/evidence/editing_sessions_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  before_action :set_evidence
+
+  private
+
+  def editing_session_record
+    @evidence
+  end
+
+  def set_evidence
+    @node = current_project.nodes.find(params[:node_id])
```
aapomm: Does this need to be an instance variable?

app/controllers/notes_controller.rb
```diff
@@ -103,4 +112,8 @@ def set_auto_save_key
       "node-#{@node.id}-note"
     end
   end
+
+  def set_form_cancel_path
+    @form_cancel_path = @note.new_record? ? project_node_path(current_project, @node) : project_node_note_path(current_project, @node, @note)
```
aapomm: Break into `if else` instead

spec/features/edit_locking_spec.rb
```diff
@@ -1,71 +1,90 @@
...
+    let(:issue) { create(:issue, node: project.issue_library, state: 'ready_for_review') }
+    let(:record) { issue }
+    let(:edit_path) { edit_project_qa_issue_path(project, issue) }
 
-    Capybara.using_session(:user_b) do
-      visit edit_project_issue_path(project, issue)
-      click_link 'Edit anyway'
-      expect(page).to have_content('Edit issue')
+    # Cancel lands on the QA show page, which loads inline threads asynchronously.
+    # Wait for the highlight before RSpec rolls back the test transaction.
```
aapomm: Why do we need to wait for the highlight and how is the inline thread affecting this spec?

app/controllers/concerns/editing_sessions_actions.rb
MattBudz: Fair enough, I inlined it. 

I originally extracted it to a concern, thinking that if we wanted to change this behavior, we'd only need to change it in one place.

app/controllers/cards/editing_sessions_controller.rb
MattBudz: good catch, we do not. fixed.

app/controllers/evidence/editing_sessions_controller.rb
MattBudz: fixed in the same commit as the board/list var

PR #1713, opened 2026-09-22
Keep avatars visible while Gravatar loads

etdsoft:
### Summary

Keep the local default avatar visible while Gravatar loads. A Stimulus controller replaces it only after the remote image loads successfully, so slow or failed requests leave a usable avatar in place.

### Testing steps

**How To Test**

1. Log in with an account that has a registered Gravatar and open a page displaying your avatar.
2. Enable slow network throttling and disable caching in browser developer tools, then reload the page.
3. Assert that the default avatar appears first and changes to the registered avatar after it loads.
4. Block requests to `secure.gravatar.com` in browser developer tools and reload the page.
5. Assert that the default avatar remains visible without a broken-image icon.
6. Remove request blocking and log in with an account that has no registered Gravatar.
7. Assert that the Gravatar request returns HTTP 404 and the local default avatar remains visible without a broken-image icon.
8. Disable JavaScript and reload the page.
9. Assert that the default avatar still appears.
10. Re-enable JavaScript, open an issue, and add a comment.
11. Assert that the new comment's avatar displays correctly without reloading the page.

### Other Information

- Avatar helper specs: 7 examples, 0 failures.
- Non-JavaScript suite: 1,296 examples, 0 failures, 19 pending.
- Changed-file RuboCop, dependency audit, Ruby audit, and Brakeman passed.
- JavaScript syntax and controller lifecycle checks passed.
- Full-suite browser checks hit Firefox navigation timeouts; a failing example also reproduces with the original avatar helper.

### Check List

- [x] Added a CHANGELOG entry


app/helpers/avatar_helper.rb
```diff
@@ -29,20 +18,29 @@ def avatar_image(user, opt = {})
...
     content_tag :span, class: opt[:class] do
-      image_tag(avatar_url(user, size: opt[:size]), img_properties) +
+      image_tag(user.try(:avatar).presence || opt[:fallback_image], img_properties) +
         (opt[:include_name] ? " #{user.try(:name)}" : '')
     end
   end
 
+  def avatar_url(user, options = {})
+    return user.avatar if user.try(:avatar).present?
+    return image_path(DEFAULT_PROFILE_IMAGE) if user.nil? || !user.email.include?('@')
```
caitmich: Maybe for legacy reasons I suppose but when will  `!user.email.include?('@')` ? is this needed?

caitmich: Ah I guess that's CE specific

PR #1715, opened 2026-09-23
convert uploads.js.coffee to vanilla JS with data-behavior hooks

nicolachr:
### Summary

- Converts `uploads.js.coffee` to vanilla ES6 JS (`uploads.js`), matching the style already used by sibling files (`upload.js`, `fileupload.js`). Behavior-preserving: same jQuery calls, same `this` semantics in the file-input change handler.
- Switches the upload page's element selection from ID selectors (`#status`, `#console`, `#filename`, `#spinner`, `#new_upload`, `#item_id`, `#attachment`, `#uploader`) to `data-behavior` attributes, per this repo's JS convention (element selection should always go through `data-behavior`, never IDs or classes).
- No functional or visual changes — this is a pure modernization pass, standalone from any other work, so it can merge independently.

### Testing steps

1. Visit a project's Upload page.
2. Select a tool from the dropdown and upload a small output file.
3. Confirm the progress bar animates, the console shows live log output, and the page behaves exactly as it does on `develop` today.

### Other Information

Submitted standalone per this repo's CoffeeScript migration convention (CoffeeScript → vanilla JS only, no feature changes, merged before any feature branch that touches the same file).

> I assign all rights, including copyright, to any future Dradis work by myself to Security Roots.

### Check List

- [ ] Added a CHANGELOG entry — not needed, this is an internal refactor with no user-facing change
- [x] Commit message has a detailed description of what changed and why.

PR #1716, opened 2026-09-23
redesign the upload console as a live, broadcast-driven status card

nicolachr:
### Summary

This is a PoC proving out the direction for the wider console redesign (see the JTBD pitch discussed separately), scoped narrowly to the Import/Upload consumer only. Every other console consumer (admin integrations, template kits, issue library upload, Excel export, Word export/validation) is Pro-only and untouched by this PR.

- Replaces `ConsoleController#status` polling with Turbo Stream broadcasting: `Log` now pushes each new line, and job completion/failure, directly to the browser instead of the page asking every 1-2 seconds for the life of an import.
- Redesigns the console into a single dark status card: a structured header (status dot, plain-English summary, filename/size, Copy button) sitting flush above the log output, instead of the old plain box with separate Filename/Size lines above it.
- Fixes a real bug in Import's own failure path: a background job that crashed never wrote a final log line, so the page would show "Importing..." forever with no indication anything went wrong. `UploadJob` and the inline upload path now always signal completion or failure explicitly. (Verified this doesn't affect the REST API's upload endpoint, which shares these same classes but reads job status from a separate mechanism, `JobTracker`, not this log line.)
- Extracts the console's styles into their own file (`_console.scss`), matching this app's "one file per component" SCSS convention, and renames the shared `ajax-console` mixin to `console-box` since it's no longer AJAX-specific for this consumer.
- Rebuilds auto-scroll (lost when polling was removed) as a small Stimulus controller.

Branched on top of #1715 (the standalone CoffeeScript-to-vanilla-JS conversion this page's JS needed first, per this repo's CoffeeScript migration convention) — this PR's own diff is scoped to just the broadcasting + redesign work.

### Found but not fixed here (out of scope for this PoC)

- `ConsoleController#status` has the same "spins forever after a background-job crash" bug as above, and it's shared by every *other* console consumer that still polls it (this PR's `Log`/`UploadJob` fix doesn't reach it, since Import doesn't call this endpoint anymore). Worth its own standalone bug-fix PR.
- Pro's uploader page has a duplicate `data:` key in the same `select_tag` call, silently dropping its RTP-validation attributes. Unrelated to this PR, found incidentally while testing the CE→Pro sync for #1715.

### Testing steps

1. Log in as a user with author or admin permissions.
2. Navigate to a project, then to that project's **Upload** page (Tools → Upload).
3. From the "Tool you are uploading output from" dropdown, select a scanner tool (e.g. Nessus).
4. Leave or set the "State of uploaded records" dropdown to any value.
5. Choose a small (under 1MB) valid output file for the selected tool and select it in the file picker.
6. Assert the upload progress bar animates to 100% and the "Output console" panel becomes visible.
7. Assert a status header appears at the top of the console showing a blue dot and the text "Importing…".
8. Assert the filename and file size appear directly beneath the status header.
9. Assert log lines appear in the dark console box below, one at a time, without the page reloading or you needing to refresh.
10. As more log lines appear than fit in the console box, assert the box automatically scrolls down to keep the newest line visible.
11. Wait for the import to finish.
12. Assert the status dot turns green and the header text changes to "Import complete.".
13. Click the "Copy log" button in the top-right of the console header.
14. Assert a "Copied to clipboard!" confirmation briefly appears near the button.
15. Paste the clipboard contents into a text field and assert it matches the log lines shown in the console.
16. Navigate to the Upload page again and repeat steps 3-5 using a file larger than 1MB for the same tool.
17. Assert the console still updates live (step 9) and completes (step 12) the same way as the small-file case — this path runs the import as a background job instead of inline.
18. Repeat steps 3-5 using a file that is invalid or malformed for the selected tool (e.g. an empty file, or a file for the wrong tool).
19. Assert the status dot turns red and the header text changes to "Import failed. See the log below for details.".
20. Assert the console log below the header shows an error message explaining what went wrong.

### Other Information

This is a proof of concept validating the technical approach (Turbo Streams broadcasting) and design direction for the broader console redesign. The other five console consumers are deliberately untouched and out of scope for this PR.

### Check List

- [x] Added a CHANGELOG entry


app/views/logs/_status.html.erb
```diff
@@ -0,0 +1,24 @@
+<% warning_count = local_assigns.fetch(:warning_count, 0) %>
+<% state = log&.state || :running %>
+<% status_class = { completed: 'text-success', failed: 'text-error', running: 'text-primary' }[state] %>
```
MattBudz: this doesn't belong in a view.

app/views/console/_console.html.erb
```diff
@@ -15,7 +15,7 @@
             This task can take some time and will be executed in the background.<br>
             Progress can be tracked here:
           </div>
-          <div id="console"></div>
+          <div id="console" class="console"></div>
```
MattBudz: do we still need the id?

app/views/upload/index.html.erb
```diff
@@ -60,10 +60,13 @@
     <div class="content-container mb-0 h-100">
       <h4 class="header-underline">Output console</h4>
       <div id="status" data-behavior="status"></div>
-      <%= content_tag :div, id: 'result', style: 'display:none', data: { url: status_console_index_path, behavior: 'result' }  do %>
-        <div><strong>Filename</strong>: <span data-behavior="filename"></span> <span data-behavior="spinner"><%= image_tag 'loading.gif', style: "margin:0;", alt: 'loading spinner' %></span></div>
-        <div><strong>Size</strong>: <span data-behavior="filesize"></span></div>
-        <div id="console" class="mx-0 mb-0" data-behavior="console"></div>
+      <%= content_tag :div, id: 'result', style: 'display:none', data: { behavior: 'result' } do %>
```
MattBudz: inline styles 🙅‍♂️

app/views/upload/index.html.erb
```diff
@@ -60,10 +60,13 @@
...
+        <div class="console-info">
+          <span data-behavior="filename"></span> &middot; <span data-behavior="filesize"></span>
+          <span data-behavior="spinner"><%= image_tag 'loading.gif', style: "margin:0;", alt: 'loading spinner' %></span>
```
MattBudz: inline styles 🙅‍♂️

config/cable.yml
```diff
@@ -2,7 +2,7 @@ development:
   adapter: async
 
 test:
-  adapter: test
+  adapter: async
```
MattBudz: why do we need this?

app/views/logs/_status.html.erb
nicolachr: moved to a helper

app/views/console/_console.html.erb
nicolachr: not anymore!

